PR #14, opened 2020-03-13
Private files

dholms:
## Problem
User should be able to add/get/list public & private files as if interacting with a file system

## Solution
Create an FFS object that contains both a public & private tree of nodes. And exposes an api to:
- `ls`
- `mkdir`
- `add`
- `cat`
- `getTree`

The private file tree implements an encrypted DAG node structure as described here: https://whitepaper.fission.codes/file-system/private-directories

Note for @icidasset: I see that you use `ipfs.catReadableStream` in drive. Unfortunately WebCrypto does not allow encrypting/decrypting streams so we can't implement that yet :unamused:. Which isn't great, especially for large file I have a plan for implementing streaming tho, tracking here: https://github.com/fission-suite/ts-sdk/issues/19

Closes https://github.com/fission-suite/ts-sdk/issues/5
Closes https://github.com/fission-suite/ts-sdk/issues/6


src/ffs/ffs.ts
```diff
@@ -0,0 +1,104 @@
...
+  if(pubLink === null || privLink === null) {
+    return null
+  }
+  const pubTreeInstance = await pubTree.resolve(pubLink.cid)
+  const privTreeInstance = await privTree.resolve(privLink.cid, keyStr)
+  return new FFS(root, pubTreeInstance, privTreeInstance)
+}
+
+export default {
+  FFS,
```
icidasset: Do we need to export FFS at all? 🤷‍♂ 
Are there any use-cases besides `empty` and `resolve` that we have here?

src/ffs/file.ts
```diff
@@ -28,21 +19,19 @@ export async function catRaw(cid: CID): Promise<Uint8Array[]> {
...
   const raw = await catRaw(cid)
-  return Buffer.concat(raw).toString()
+  return Buffer.concat(raw)
 }
 
-export async function cidToDAGLink(cid: CID, name: string): Promise<DAGLink> {
-  const ipfs = await getIpfs()
-  const stat = await ipfs.object.stat(cid)
-  return new dagPB.DAGLink(name, stat.CumulativeSize, cid)
+export async function cat(cid: CID): Promise<string> {
```
icidasset: Not a request or anything, but just wondering, how difficult would it be to make this behave the same as the js-ipfs API and have everything return an `AsyncIterable`? Or is this only possible if there were streaming?

src/ffs/ffs.ts
dholms: Yeah you might want to bootstrap an FFS off of an existing public folder so do something like
```ts
const root = await pubNode.empty()
const pubTreeInstance = await pubTree.resolve(PUB_TREE_CID)
const privTreeInstance = await privTree.empty()
const ffs = new FFS(root, pubTreeInstance, privTreeInstance)
}
```
Then again, you could also do
```ts
const ffs = await empty()
ffs.pubTree = await pubTree.resolve(PUBLIC_TREE_CID)
await ffs.updateRoot()
```
I was arguable over-aggressive with exports, but was trying to give as much flexibility as possible :laughing:

src/ffs/file.ts
dholms: yeah very manageable to do for public files. Not possible with private files until we have streaming crypto. But I was gonna to keep the APIs the same until it's supported for both

src/ffs/ffs.ts
icidasset: Ah awesome, thanks for explaining 👍

src/ffs/file.ts
icidasset: Purrfect 🐈

icidasset:
Oh yeah, this might be good to know, how do I start using this? :sweat_smile: 

So based on previous conversations I guess we're starting out with a single private directory in the root. And for that we need some auth, which will give us a private key, which I then use in `ffs.resolve(cid_from_dnslink, key)`?

expede:
👀

src/ffs/ffs.ts
```diff
@@ -0,0 +1,104 @@
...
+  }
+
+  async makeDir(path: string): Promise<FFS> {
+    const { tree, tail } = this.parsePath(path)
+    await tree.makeDir(tail)
+    return this.updateRoot()
+  }
+
+  async addFile(path: string, content: FileContent): Promise<FFS> {
+    const { tree, tail } = this.parsePath(path)
```
expede: `tail`?

src/ffs/ffs.ts
```diff
@@ -0,0 +1,104 @@
...
+  }
+
+  async cid(): Promise<CID> {
+    return this.put()
+  }
+
+  async listDir(path: string): Promise<Link[] | null> {
+    const { tree, tail } = this.parsePath(path)
+    return tree.listDir(tail)
+  }
```
expede: Minor thought: have you considered cloning the POSIX interface?

```
FFS.listDir    -> FFS.ls
FFS.makeDir    -> FFS.mkdir
FFS.getFile    -> FFS.cat
FFS.updateRoot -> FFS.sync
```

src/ffs/priv/node.ts
```diff
@@ -0,0 +1,110 @@
+import { emptyDir, encryptNode, decryptNode, decrypt, encrypt, contentToBytes, bytesToContent } from './helpers'
+import file from '../file'
+import { PrivateNodeData, Link, Node } from '../types'
+import { CID, FileContent } from '../../ipfs'
+
+export class PrivateNode implements Node {
+
+  links: Link[]
```
expede: Why not as a map?

expede:
Moving from Discord here

expede:
An idea for ontology:

```ts
class FileSystem
```

Stateful imperative wrapper / main interface

```ts
class Tree.Public
```

Stateless public tree type

```ts
class Tree.Private
```

Stateless private tree type

expede:
Some POSIX-based interface thoughts:

```ts
const ffs = new Fisson.FileSystem("Qm12345")
await ffs.mount() // defaults to background prefetching. Awaits top-level only.
await ffs.ls()
// ["photos/", "documents/"]
await ffs.cd("photos/vacation")
await ffs.ls()
// ["vacation.png"]
const myPhoto = await ffs.cat("vacation.png")
const myPhotoData = await ffs.metadata("vacation.png")
await ffs.map(/* some action on each dir/file */)
```

src/ffs/pub/tree.ts
```diff
@@ -0,0 +1,78 @@
...
+    return this.put()
+  }
+
+  async listDir(path: string): Promise<Link[] | null> {
+    const node = await this.getNode(path)
+    return node?.links() || []
+  }
+
+  async makeDir(path: string): Promise<PublicTree> {
+    const toAdd = await publicNode.empty()
```
expede: My editor is warning that the `await` doesn't do anything here

src/ipfs/types.ts
```diff
@@ -12,7 +12,7 @@ export type DAGNode = {
   toDAGLink: (opt?: { name?: string }) => Promise<DAGLink>
   addLink: (link: DAGLink) => void
   rmLink: (name: string) => void
-  serialize: () => any
+  toJSON: () => object
```
expede: 👍

src/keystore/index.ts
```diff
@@ -0,0 +1,18 @@
...
+  if(ks){
+    return ks
+  }
+  ks = await keystore.init(KEYSTORE_CFG)
+  return ks
+}
+
+export default {
+  get
+}
```
expede: Honest question: why not `export default function get()`

Also why not use an arrow functions in the vast majority of places? They're preferred now because of various simpler rules, no?

src/ffs/ffs.ts
```diff
@@ -0,0 +1,104 @@
...
+import pubNode, { PublicNode } from './pub/node'
+import { Tree, Node, Link } from './types'
+import { CID, FileContent } from '../ipfs'
+import pathUtil from './path'
+
+export class FFS {
+
+  root: PublicNode
+  pubTree: PublicTree
+  privTree: PrivateTree
```
expede: More from the conversation in Discord:

If we rearrange stuff, this could be perhaps streamlined and the stateful classes split out (functional OO design). For example:

```ts
class Session {
  // Stateful
  // Manages where the user "is" in the system, mutable updates, and so on
  // Think of this as a bash session
}

class Directory {
  // Stateless
  // i.e. Currently `PubTree`

  cid: CID
  files: {String => File}
  directories: {String => Directory}

  // ...
}

class Directory.Private 
  extends Directory 
  implements Private {
    key: AES256
    // ...
}

class File {
  cid: CID
  data: Buffer // or something

  constructor(cid_: CID) {
    this.cid = cid_
  }

  // ...some methods
}

class File.Private 
  extends File
  implements Private {
    key: AES256

    // ...
}
```

src/ffs/pub/tree.ts
```diff
@@ -0,0 +1,78 @@
...
+    return this.root.put()
+  }
+
+  async cid(): Promise<CID> {
+    return this.put()
+  }
+
+  async listDir(path: string): Promise<Link[] | null> {
+    const node = await this.getNode(path)
+    return node?.links() || []
```
icidasset: Would it be possible to use https://github.com/ipfs/js-ipfs/blob/master/docs/core-api/FILES.md#ls here instead, so we can get the `mtime` (last-modified timestamp)? Or is there another way to get that? Also, should I run `ffs.pubNode.isFile` on each of these links to find out if I'm dealing with a directory or a file? (`files.ls` has a `type` property for this)

dholms: yeah totally. Sorry didn't realize these were being used! I'll modify our `Link` type to include `mtime: number` & `isFile: boolean` and add these values to public nodes using `ipfs.ls` and include the data in our uploaded encrypted nodes as well :+1:

src/ffs/ffs.ts
dholms: yup its a file path. I could use `relPath` or something similar instead

src/ffs/priv/node.ts
dholms: IPFS DAGNode's use an array and I was modeling our node after theirs. I don't think there's any reason that we need to stick to it though... I remember reading a discussion on this somewhere in IPFS's Github. I'll try and dig that up to see their rationale

src/ffs/pub/tree.ts
dholms: So this is a weird one because `PrivateTree.mkdir` is async but `PublicTree.mkdir` is not. But I think we want to keep the API the same :thinking:

src/keystore/index.ts
dholms: I kind of like the semantics of 
```ts
import keystore from '../keystore'
const ks = await keystore.get()
```
But that's not a big deal. We could always something like this (which reads fine as well):
```ts
import getKeystore from '../keystore'
const ks = await getKeystore()
```

The bigger reason is that we might have a `set` function in this module as well (similar to the ipfs module). If a user wants to use their own predefined keystore, they can set it for the sdk. I just hadn't added that function yet :sweat_smile:

dholms: As far as arrow functions go, isn't the main difference that they bind the current scope to the function? `.bind(this)`

I don't use `this` anywhere except for in methods on classes (where you wouldn't want to bind a larger scope than the class). 

I've just always done function declarations like this in files & used arrow functions for short anonymous functions in code. Happy to switch it up though if better/more reliable

src/ffs/ffs.ts
expede: Yeah, that would be clearer 👍

src/ffs/priv/node.ts
expede: Yeah, let's make this work for our use cases. `Link[]` has issue with namespace collisions, &c. It's a level of detail that we don't probably don't want to deal with, but we can always convert back and forth (FFS -> DAGNode and DAGNode -> FFS)

src/ffs/pub/tree.ts
expede: Oh I see. Yeah, okay that makes sense. Better than switching on each. I wonder if we can suppress that warning?

src/keystore/index.ts
expede: > I kind of like the semantics of

Cool sounds good 👍

expede: > Happy to switch it up though if better/more reliable

Yeah, there's way fewer edge cases (and/or is much more intuitive). `function` does behave in surprising ways sometimes that arrows don't. Implicit scoping can lead to some very subtle bugs.

An oldie but a goodie: https://stackoverflow.com/a/23045200

(I will note that `.bind(this)` is actually surprisingly slow, though I doubt it'll bite us in this case because we probably won't be calling it enough / often will pass explicitly as an arg)

But fundamentally this is your decision, given that you're more in the code.

Thoughts?

expede:
Also just to clarify: all paths in this API are full paths, yes? e.g. you need to pass like `public/foo/bar/baz.jpg` to the top-level object

dholms:
> Also just to clarify: all paths in this API are full paths, yes? e.g. you need to pass like `public/foo/bar/baz.jpg` to the top-level object

At the moment yes. I'll be incorporating your suggestion around a stateful "Session" class soon

src/ffs/ffs.ts
```diff
@@ -0,0 +1,109 @@
...
+
+  static async empty(rootKeyName: string = 'filesystem-root'): Promise<FFS> {
+    const root = await pubNode.empty()
+    const pubTreeInstance = await pubTree.empty()
+    const privTreeInstance = await privTree.empty(rootKeyName)
+    return new FFS(root, pubTreeInstance, privTreeInstance)
+  }
+
+  // upgrade public IPFS folder to FileSystem
+  static async upgradePublicCID(cid: CID, rootKeyName: string = 'filesystem-root'): Promise<FFS> {
```
icidasset: @dholms Does this do the following?

```shell
# Suppose I have the CID `Qm111` of a __folder__ with the following contents
file_a.txt
file_b.pdf
```

If I then run `upgradePublicCID("Qm111")`, would I get following tree?

```shell
# New Tree
/public/file_a.txt
/public/file_b.pdf
```

dholms: yes but it also creates an empty private folder as well. so you'd get 
```
root
  public
    file_a.txt
    file_b.pdf
  private
```

src/keystore/index.ts
dholms: Alright I'm convinced. I think I'll use arrow functions over function declarations when defining functions, but not for methods on classes. That seems like it can carry with it it's own class of bugs

PR #20, opened 2020-03-31
Refactor private

dholms:
## Problem
Abstractions aren't quite right. Node & Tree are really the same data structure + lots of repeat code between public & private

## Solution
- Merge Node & Tree
- PrivateTree extends PublicTree
- change commands to POSIX-style interface


src/ffs/ffs.ts
```diff
@@ -1,109 +1,111 @@
...
-    this.root.replaceLink(privLink)
-    return this
+    this.root = this.root
+                  .replaceLink(pubLink)
+                  .replaceLink(privLink)
+    return this.root.put()
   }
 
-  parsePath(path: string): { tree: Tree, tail: string } {
+  async runOnTree(path: string, fn: (tree: Tree, relPath: string) => Promise<Tree>): Promise<Tree> {
```
dholms: I think this is the most confusing part of the new code. It's used in `mkdir`/`add` since we need to update the respective tree root after adding the file. Any thoughts to clean this up? The only other option I can think of is to do the same if/else/throw in both mkdir & add which could end up being pretty repetative :thinking:

PR #25, opened 2020-04-13
User creation and authentication

icidasset:
Adds the following functions:
- `createAccount`
- `didJWT`
- `didKey`
- `isUsernameAvailable`
- `updateRoot`


publish.sh
```diff
@@ -2,5 +2,5 @@ yarn build
 cp ./package.json dist/
 cp ./README.md dist/
 cp ./LICENSE dist/
-cd dist
-npm publish
+# cd dist
+# npm publish
```
icidasset, line 6: Note to self, remove this changes when done.

src/common.ts
```diff
@@ -0,0 +1,33 @@
...
+export function makeBase64UrlSafe(a: string): string {
+  return a.replace(/\//g, "_").replace(/\+/g, "-").replace(/=+$/, "")
+}
+
+
+
+// CRYPTO
+
+
+export function isRSAKeystore(ks: KeyStore): boolean {
```
dholms, line 31: I'd maybe probably move this over `keystore/basic.ts`

I've been thinking of `common` as only having javascript helpers: encoding/decoding, typechecks, lodash-esque functions, etc

Also this should arguably be a function in the `keystore-idb` pkg

src/user/identity.ts
```diff
@@ -0,0 +1,73 @@
...
+import eccOperations from 'keystore-idb/ecc/operations'
+import rsaOperations from 'keystore-idb/rsa/operations'
+import utils from 'keystore-idb/utils'
+
+import { base64UrlEncode, isRSAKeystore } from '../common'
+import { getKeystore } from '../keystore'
+
+
+const ED_DID_PREFIX: ArrayBuffer = new Uint8Array([ 0xed, 0x01 ]).buffer
+const RSA_DID_PREFIX: ArrayBuffer = new Uint8Array([ 0x00, 0xf5, 0x02 ]).buffer
```
dholms, line 11: :sparkles: Magic bytes :sparkles:

src/user/identity.ts
```diff
@@ -0,0 +1,73 @@
...
+  // Encode parts in JSON & Base64Url
+  const encodedHeader = base64UrlEncode(JSON.stringify(header))
+  const encodedPayload = base64UrlEncode(JSON.stringify(payload))
+
+  // Signature
+  const operator = isRSA
+    ? rsaOperations
+    : eccOperations
+
+  const signed = await operator.signBytes(
```
dholms, line 41: This is all fine, but it's essentially what `keystore-idb` does under the hood.

Which means there's a short coming with the `keystore` pkg :sweat_smile: 

I actually ran into a similar issue as well.... I'll create an issue over on that repo to add byte operations to the keystore interface as well so you don't have to do all this stuff

dholms: https://github.com/fission-suite/keystore-idb/issues/22

dholms:
Nice this looks really good :+1: 

I think a few shortcomings in `keystore-idb` are putting more crypto code here than there really needs to be. I'll see about making those updates soon, shouldn't be too much work

src/user/identity.ts
dholms: wait actually misread this a bit on first pass. we are dealing with strings :sweat_smile: 
so
```ts

  // Signature
  const operator = isRSA
    ? rsaOperations
    : eccOperations

  const signed = await operator.signBytes(
    utils.strToArrBuf(`${encodedHeader}.${encodedPayload}`, 8),
    ks.writeKey.privateKey,
    ks.cfg.hashAlg
  )

  const encodedSignature = base64UrlEncode(
    utils.arrBufToStr(signed, 8)
  )
```

can actually just be shortened to 
```ts
  const sig = await ks.sign(`${encodedHeader}.${encodedPayload}`)
  const encodedSiganture = makeBase64UrlSafe(sig)
```

Unless I'm missing something here :thinking:

src/common.ts
icidasset: Yeah, that makes sense 👍 

I'll add it there. The only reason I really need it is to differentiate between the types of key stores. Is there another way of doing this? I was thinking it'd be something along the lines of:
```
keystore.algorithm => RSA
```
and then we'd have a list somewhere with all the algorithms, or whatever you wanna call it.
Thoughts?

src/user/identity.ts
icidasset: 🧙

src/user/identity.ts
icidasset: Yeah, we should definitely change it in keystore-idb. I needed `signBytes` originally for something different, I needed the array buffer for usage with the crypto api. But it turned out that was unnecessary, so ya.

Anyhow, long story short, you're right, I can use `ks.sign`, __except__, I need to be able to pass a custom `charSize`. Here it needs to be `8`, not `16` (what it is currently set to).

What do you think? Should I add that as an option to the `sign` method, or am I doing something wrong?

dholms: you can pass a custom `charSize` in the `keystore-idb` cfg. But it might be the case that you want your keystore to default to `16` for instance, but `8` for a single function. We could add an optional `cfg` param to every method in the keystore for overriding the cfg on method call. Thoughts?

src/common.ts
dholms: Yeah this could be tightened up as 
```ts
return ks.cfg.type === 'rsa'
```

There is an enum in `keystore-idb` for this:
```ts
export enum CryptoSystem {
  ECC = 'ecc',
  RSA = 'rsa',
}
```

So a non-stringly typed version would be
```ts
import { CryptoSystem } from 'keystore-idb/types'

function isRSAKeystore(ks: KeyStore): boolean {
  return ks.cfg.type === CryptoSystem.RSA
}
```

We also might want to add in a type assertion for TS here. not really necessary but also doesn't hurt :
```ts
isRSAKeystore(ks: Keystore): ks is RSAKeyStore
```
(still returns a boolean, but also lets the TS Compiler know what type of KS it is)

dholms: but yeah I'm actually leaning towards just putting this on the `KeystoreBase` class in `keystore-idb`:
```ts
  isRSA(): this is RSAKeyStore {
    return this.cfg.type === CryptoSystem.RSA
  }

  isECC(): this is ECCKeyStore {
    return this.cfg.type === CryptoSystem.ECC
  }
```

src/user/identity.ts
icidasset: I guess we should add it as an option.
Seeing that all the javascript stuff will need to be `utf-16`,
and other, less common, things could be `utf-8`.

dholms: Okay I went over to have a look at it, started messing around, and just figured that I'd finish it off :sweat_smile: 

Check out the PR here: we can get it merged & deployed

dholms: oops: https://github.com/fission-suite/keystore-idb/pull/24

src/user/identity.ts
```diff
@@ -0,0 +1,87 @@
...
+// 🧙
+
+
+/**
+ * JWT algorithm to be used in a JWT header.
+ */
+function jwtAlgorithm(cryptoSystem: CryptoSystem): string | null {
+  switch (cryptoSystem) {
+    case CryptoSystem.ECC: return 'Ed25519';
+    case CryptoSystem.RSA: return 'RS256';
```
icidasset, line 72: @dholms I just did this instead of adding this `isRSA` and `isECC` methods. Is that ok?

src/user/identity.ts
icidasset: Done 👍

src/user/identity.ts
dholms: Yes that's great :+1: 
_Except..._ WebCrypto doesn't support Ed25519, it uses NIST Curves (P-256, P-384, P-521). Sooo if we want to support those, we'll need to figure out magic bytes & what not for them on the server...

We may want to have `CryptoSystem.ECC` return `null` until we get that set up server-side. The ts-sdk always generates RSA keys unless the user overrides it

icidasset: Ohhh ok, woops. I assumed ECC was the same as ED 😅 Yeah, we should discuss this later. In the meantime, will remove the ecc/ed stuff.

dholms: Yeah too many acronyms, it's all so confusing :sweat_smile: 

ED is a _type of_ ECC. But multiformats are supposed to be completely self-describing so each ECC curve needs it's own :sparkles: magic bytes :sparkles:

src/ffs/ffs.ts
```diff
@@ -92,10 +95,27 @@ export class FileSystem {
...
+  }
+
+  addSyncHook(hook: SyncHook): Array<SyncHook> {
+    this.syncHooks.push(hook)
+    return this.syncHooks
+  }
+
+  removeSyncHook(hook: SyncHook): Array<SyncHook> {
+    return this.syncHooks.filter(h => h !== hook)
   }
```
icidasset, line 119: Added hooks here ☝️

PR #26, opened 2020-04-14
Header folder

dholms:
## Problem
We want to associate metadata, and various other header information with files and folders
We also want to preserve the option for a "bare file system" (no header info)

## Solution
Add a "nested" approach to folders/files
A CID for a dir/file points to a dir containing header information (`isFile`, `mtime`, `cache`, `key` etc) as well as `index` which either contains links (for a dir) or the file contents

Leave the simple way of creating/navigating the file system as well. Store `version` in header folder (using semver), if not found, default to `0.0.0` (non-nested file system)

Check the version of a given node before getting/parsing it

Also merged https://github.com/fission-suite/ts-sdk/pull/27 into this branch for storing pinlists in the header folder of  private nodes


src/ffs/normalizer/basic.ts
```diff
@@ -0,0 +1,36 @@
+import { CID, FileContent } from '../../ipfs'
+import { TreeData, PrivateTreeData, Metadata} from '../types'
+import util from './util'
```
icidasset, line 3: Do you think it makes sense to rename these `util` module to `common`?
And then rename `common/index.ts` to `common.ts`?

I think it makes more sense here.
Seeing that `normalizer/basic.ts` and `normalizer/nested.ts` have these things "in common".

src/ipfs/constants.ts
```diff
@@ -0,0 +1,5 @@
+export const DAG_NODE_DATA = Buffer.from([8, 1])
```
icidasset, line 1: Could we add some documentation here? 🙏 So we know what this is for.

src/ipfs/encoded.ts
```diff
@@ -0,0 +1,44 @@
...
+  return basic.add(toAdd)
+}
+
+export const catAndDecode = async (cid: CID, key?: string): Promise<any> => {
+  const buf = await basic.catBuf(cid)
+  const toDecode = key !== undefined ? await keystore.decrypt(buf, key) : buf
+  return cbor.decode(toDecode)
+}
+
+export const getBool = async (cid: CID, key?: string): Promise<boolean | undefined> => {
```
icidasset, line 18: Can we use a custom type here instead of `undefined`? Would be great to have something along the lines of a `Result`. Something like:
```
type DecodingError = ContentTypeDoesNotMatch
Ipfs.Encoded.getBool(ContentType): Promise<Result<DecodingError, Bool>>
```

src/ffs/normalizer/basic.ts
dholms: Yeah I actually changed the naming of these around in the branch that I just merged into this one.
There's now `basic`, `nested`, `v0_0_0` and `v_1_0_0`, no more `util`.
The renaming is basically:
- `util -> basic`
- `basic -> v0_0_0`
- `nested` split into `nested` (utility functions) & `v_1_0_0` (exposed functions that get called by the lib)

Not totally sure how I feel about naming files after the versions :thinking:, but I also think it helps make it very explicit how each version of the file system operates. And we hopefully won't have _that_ many versions. Thoughts?

dholms: > And then rename common/index.ts to common.ts

Yeah I wasn't sure if I would end up having more commons functions that I'd want to structure a little better. But agreed that this is unnecessary right now. I'll change it for now and we can always change back if it becomes an issue

src/ipfs/encoded.ts
dholms: ended up moving around/changing some of this code. but incorporated your idea around custom error type. I think error handling could be improved in general, gonna think on this...

src/ffs/normalizer/basic.ts
icidasset: > but I also think it helps make it very explicit how each version of the file system operates
Agreed, I like the versioned modules. We shouldn't have that many versions anyway, and if we do, we can refactor then.

src/ffs/normalizer/header.ts
```diff
@@ -0,0 +1,83 @@
...
+    Object.entries(withVersion).map(async ([name, val]) => {
+      if(val !== undefined){
+        const cid = await ipfs.encoded.add(val, key)
+        return { name, cid, isFile: true }
+      }
+      return null
+    })
+  )
+  linksArr.push({ name: 'index', cid: index, isFile: false })
+  const links = link.arrToMap(linksArr.filter(notNull))
```
expede, line 48: I guess these are heterogeneous lists?

src/ffs/normalizer/header.ts
```diff
@@ -0,0 +1,83 @@
...
+  return basic.putLinks(links, key)
+}
+
+export const getVersion = async(cid: CID, key?: string): Promise<FileSystemVersion> => {
+  const version = await getValue(cid, 'version', isString, key)
+  switch(version) {
+    case "1.0.0":
```
expede, line 55: We may want to actually parse this and make it work with `1.x.x`

src/ffs/normalizer/versions/v1_0_0.ts
```diff
@@ -0,0 +1,73 @@
...
+  return basic.getFile(indexCID, key)
+}
+
+export const getTreeData = async (cid: CID, key?: string): Promise<TreeData | PrivateTreeData> => {
+  const indexCID = await basic.getLinkCID(cid, 'index', key)
+  if(!indexCID) {
+    throw new Error(`Links do not exist: ${indexCID}`)
+  }
+  const links = await basic.getLinks(indexCID, key)
+  const childKey = key ? await header.getChildKey(cid, key) : undefined
```
expede, line 23: Just double checking: `undefined`, not `null`?

src/ffs/private/tree.ts
```diff
@@ -73,8 +83,24 @@ export class PrivateTree extends PublicTree {
...
-    return new PrivateTree(links, this.key)
+    return new PrivateTree(links, this.version, this.key, this.pinMap)
+  }
+
+  pinList(): CID[] {
+    return Object.entries(this.pinMap).reduce((acc, cur) => {
+      const [parent, children] = cur
+      return acc
+              .concat([parent])
+              .concat(children)
```
expede, line 102: 💅 Any reason to not use the spread operator here?

src/ffs/normalizer/header.ts
```diff
@@ -0,0 +1,83 @@
...
+}
+
+export const put = async(index: CID, header: Header, key?: string): Promise<CID> => {
+  const withVersion = {
+    ...header,
+    version: FileSystemVersion.v1_0_0
+  }
+  const linksArr = await Promise.all(
+    Object.entries(withVersion).map(async ([name, val]) => {
+      if(val !== undefined){
```
expede, line 40: Best practice: match on positives. Also becomes a guard in this case.

```ts
if (val === undefined) return null // Guard statement!

const cid = await ipfs.encoded.add(val, key) 
return {name, isFile: true}
```

src/ffs/private/tree.ts
dholms: Nope. good call :+1:

src/ffs/normalizer/header.ts
dholms: :+1:

src/ffs/normalizer/header.ts
dholms: Ah better yet, filter the list before mapping it, then the list is homogeneous & we don't have to use type guards:

```ts
    Object.entries(withVersion)
      .filter(([_name, val]) => val !== undefined)
      .map(async ([name, val]) => {
        const cid = await ipfs.encoded.add(val, key)
        return { name, cid, isFile: true }
      })
````

File: src/ffs/normalizer/versions/v1_0_0.ts (hunk 6 above)
Comment (dholms) [reply src/ffs/normalizer/versions/v1_0_0.ts:23]: Yup. A lot of these methods do essentially the same thing but with/without encryption so they take an optional `key` param
For instance:
```ts
export const getMetadata = async (cid: CID, key?: string): Promise<Metadata>
```

So here I'm getting the childKey that would be passed to `getMetadata`. Since it's an optional param, it is type `string | undefined` instead of `null`

This could be cleaner with the `Maybe` type that we've been talking about adding. Would force more intentionality :thinking:

src/ffs/normalizer/header.ts
dholms: Hmmm so like a semver obj instead of an enum?

expede: Yeah, exactly. It's a bit brittle right now, since `1.0.1` will fail the comparison, when that's totally okay! We can't assume that everyone is running the same version, which is where semver shines.

PR #27, opened 2020-04-15
Pinlist for private tree

dholms:
## Problem
Our IPFS node cannot navigate private trees in order to recursively pin them

## Solution
Store a list of all child CIDs (direct & recursive children) in a node's header folder. Update when children are updated.

`FFS.pinList` returns `[...privateTree.pinList(), ffs.rootCID]` (since the ipfs node _can_ recursively traverse the private tree)


dholms:
Gonna merge this into the header branch cause it's really the same feature set & it fixes some issues in that branch

PR #28, opened 2020-04-15
Fix get recursion for empty paths

dholms:
## Problem
`getRecurse` base cause is throwing an error when accessing an empty path instead of returning the parent tree

## Solution
Change `getRecurse` to return `null` instead, use the `NonEmptyPath` data type, and leave error handling to the caller

### DIFF
--- a/src/ffs/public/util.ts
+++ b/src/ffs/public/util.ts
@@ -40,15 +40,16 @@ export const addRecurse = async (tree: Tree, path: NonEmptyPath, child: Tree | F
   return tree.updateDirectChild(toAdd, name)
 }
 
-export const getRecurse = async (tree: Tree, path: string[]): Promise<Tree | File> => {
-  const nextTree = await tree.getDirectChild(path[0])
-  if(path.length <= 1 && nextTree !== null){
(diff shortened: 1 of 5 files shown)


src/ffs/public/util.ts
```diff
@@ -40,15 +40,16 @@ export const addRecurse = async (tree: Tree, path: NonEmptyPath, child: Tree | F
...
+  const nextTree = await tree.getDirectChild(head)
+  if(nextPath === null){
     return nextTree
-  }
-  if(nextTree === null || isFile(nextTree)){
-    throw new Error("Path does not exist")
-  }
-  return getRecurse(nextTree, path.slice(1))
+  } else if (nextTree === null || isFile(nextTree)){
+    return null
```
icidasset, line 50: Shouldn't this be `return nextTree`? 🤔

icidasset: Or does it only return a file when `nextPath === null`?

dholms: yup only when `nextPath === `null`
since if there is still some path to go down, and the "next tree" is either a file or doesn't exist, it's impossible to do so

PR #29, opened 2020-04-16
Readme

dholms:
## Problem
No docs/README

## Solution
Add a README with an overview of filesystem + in depth API docs


README.md
```diff
@@ -1,12 +1,264 @@
-![](https://github.com/fission-suite/PROJECTNAME/raw/master/assets/logo.png?sanitize=true)
+f![](https://github.com/fission-suite/PROJECTNAME/raw/master/assets/logo.png?sanitize=true)
```
icidasset, line 1: Small typo here, `f` at the beginning.

README.md
```diff
@@ -1,12 +1,264 @@
...
+**empty**
+
+Creates a file system with an empty public tree & an empty private tree at the root
+
+Params: 
+- cfg: `FileSystemConfig` _optional_
+
+Example:
+```ts
+import FileSystem from 'fission-sdk/ffs'
```
icidasset, line 71: Should we rename `ffs` to just `fs`? 🤷

README.md
```diff
@@ -1,12 +1,264 @@
...
+const wnfs = await FileSystem.foruser("boris")
+```
+
+---
+
+### Methods
+Methods for interacting with the filesystem all use **absolute** paths. We're planning on adding a [stateful session](https://github.com/fission-suite/ts-sdk/issues/24) but for now, filesystem state will need to be tracked in your application.
```
icidasset, line 111: Should we mention here that it's an absolute path without a leading slash? Or does it accept leading slashes as well?

README.md
```diff
@@ -1,12 +1,264 @@
...
+const content = "hello world"
+const updatedCID = await wnfs.add("public/some/path/to/a/file", content)
+// creates a file called "file" at "public/some/path/to/a"
+```
+
+---
+
+**cat**
+
+Retrieves some file content at a given path
```
icidasset, line 164: Might be good to mention the return type here. Does this return a blob or an array buffer?

README.md
dholms: It accepts both :+1:

README.md
dholms: Yes good call. had been meaning to do that

README.md
dholms: Well data in the private tree is cbor encoded before being uploaded, so from the private tree `cat` can return deserialized info like an object, array, string etc. From the public tree, everything comes out as a `Buffer` (or `Uint8Array` in the browse). So it returns type `FileContent (object | string | Blob | Buffer)`.

I'm really not quite sure the best way to handle this... Standardize it around always return a `Buffer` or return a deserialized object when possible (so the vague `FileContent` type)?

Regardless, I took your advice on adding return types :+1:

icidasset: Hmm, I'd say standardize on one format. Not sure if that format would be a Buffer, Blob or Async Iterable 🤷

PR #31, opened 2020-04-22
Travis

dholms:
## Problem
Travis is always failing

## Solution
Fix linting errors, remove code coverage reqs (until we have more tests)

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -9,4 +9,17 @@ module.exports = {
     'plugin:@typescript-eslint/eslint-recommended',
     'plugin:@typescript-eslint/recommended',
   ],
+  rules: {
+    '@typescript-eslint/member-delimiter-style': ['error', {
+      "multiline": {
+        "delimiter": "none",
+        "requireLast": false
+      },
+      "singleline": {
+        "delimiter": "comma",
+        "requireLast": false
+      }
+    }],
+    "@typescript-eslint/no-use-before-define": ["off"]
+  }
 };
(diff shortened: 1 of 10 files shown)


.eslintrc.js
```diff
@@ -9,4 +9,17 @@ module.exports = {
     'plugin:@typescript-eslint/eslint-recommended',
     'plugin:@typescript-eslint/recommended',
   ],
+  rules: {
```
dholms, line 12: Let me know your thoughts on these rules!
`@typescript-eslint/no-use-before-define` was causing it to error when using a function before it's defined. In Haskell, we normally put the most important function at the top of a file, and helper functions beneath it.

`'@typescript-eslint/member-delimiter-style'` was set to a semi-colon by default so for instance, types had to be laid out like:
```ts
type Fake = {
  key1: string;
  key2: string;
  key2: string;
}
```
It's mostly personal preference, but I much prefer the clean look of no delimiter. Thoughts?

icidasset: Yeah same, only semicolons when absolutely necessary 😄 
Rules seem fine 👍

src/ffs/types.ts
```diff
@@ -41,7 +41,7 @@ export type Links = { [name: string]: Link }
 // ----
 
 export type NonEmptyPath = [string, ...string[]]
-export type SyncHook = (cid: CID) => any
+export type SyncHook = (cid: CID) => unknown
```
icidasset, line 44: I did not know `unknown` was a thing 🤷

dholms: yeah just recently found this out also! my understanding is that it's like `any` but doesn't completely blow out the type system

PR #33, opened 2020-04-23
Remove

dholms:
## Problem
No `rm` interface for nested files/dirs

## Solution
Add `rm` method to Tree & FFS


dholms:
Accidental duplicate

PR #34, opened 2020-04-23
UCAN 🦜

icidasset:
Some initial work for the UCAN JWTs.
Allows people to create an account using Drive and use their file system.

- Fetches the API DID and caches it for 3 hours
- Adds additional UCAN properties to the JWT
- Fixes issue with `updateRoot` function (ie. don't use sync)

### DIFF
--- a/src/common.ts
+++ b/src/common.ts
@@ -1,12 +1,9 @@
-// CONSTANTS
-
+// 🏔
 
 export const API_ENDPOINT = 'https://runfission.com'
 
 
-
 // BASE64
-
 
 export function base64UrlDecode(a: string): string {
   return atob(a).replace(/\_/g, "/").replace(/\-/g, "+")
@@ -95,6 +92,7 @@ export const mapObjAsync = async <T, S>(obj: {[k: string]: T}, fn: (t: T, k?: st
   )
   return newObj
 }
+
 
 export default {
   base64UrlDecode,
--- a/src/common/dns.ts
+++ b/src/common/dns.ts
@@ -1,0 +1,25 @@
+import ipfs from '../ipfs'
+
+
+export function lookupTxtRecord(host: string): Promise<string> {
+  return fetch(`https://cloudflare-dns.com/dns-query?name=${host}&type=TXT`, {
+    headers: {
+      "accept": "application/dns-json"
+    }
+  })
+  .then(r => r.json())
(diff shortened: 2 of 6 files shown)


src/common.ts
```diff
@@ -1,13 +1,37 @@
 import { KeyStore } from 'keystore-idb/types'
```
dholms, line 1: This line can be :hocho:

src/common.ts
```diff
@@ -1,13 +1,37 @@
...
+// API
+
+
+const apididcache = {
+  didKey: "",
+  lastFetched: 0,
+}
+
+
+export async function apiDidKey(): Promise<string> {
```
dholms, line 21: I think we have different conceptions of what goes in `common`. 

I've mostly been using it for "vanilla javascript helpers", typechecks, string encoding, etc. You can check that out [here](https://github.com/fission-suite/ts-sdk/pull/34/files)

I'm a bit worried about this becoming too much of a "grab bag" file with a bit of everything in it.

I'd say we either:
- split up `common` into submodules (ie `common/base64`, `common/type-checks`, `common/dns`)
- ditch the `common` name all together. `util` might be a better name for what I've been using `common` for, and dns-lookups & caching and what not could exist in its own file

Thoughts? I don't have a strong opinion, but figure we should organize this before ti gets to hodge-podgey

dholms:
Looks good overall! Just my comments re: `common` file

Also, rebase so that the Travis tests pass!

src/common.ts
icidasset: Yeah, with all that combined, it does get to be a lot. I named it `common` because these are functions that are used across multiple modules (ie. modules have, or could have, these things in common). I would use `util` if it's something more specific to a specific module. Does that make sense? 🤔 

I'll split it up into submodules 👍

src/common/dns.ts
```diff
@@ -0,0 +1,25 @@
...
+
+
+export async function lookupDnsLink(host: string): Promise<string> {
+  let t
+
+  try {
+    t = await ipfs.dns(host)
+  } catch (_) {
+    t = await lookupTxtRecord(host)
+  }
```
icidasset, line 22: @dholms I've extracted the api and dns stuff. Also, I've used Cloudflare as a backup for the ipfs-dnslink lookup here ☝️ (in case ipfs.io is down again)

PR #35, opened 2020-04-24
Split up common

dholms:
> immediately merges a PR with a bunch of misc junk in `common`

:stuck_out_tongue: 

## Solution
Split up common.
Renamed a few functions so you can do kinda semantic imports from `common` ie
```ts
import { base64 } from './common
base64.urlDecode(something)
```


src/common/base64.ts
```diff
@@ -0,0 +1,17 @@
+export function urlDecode(a: string): string {
+  return atob(a).replace(/_/g, "/").replace(/-/g, "+")
```
dholms, line 2: this line was
```ts
  return atob(a).replace(/\_/g, "/").replace(/\-/g, "+")
```
but eslint told me those escape characters weren't necessary. still seems to work fine

src/common/index.ts
```diff
@@ -0,0 +1,26 @@
+import apiMod from './api'
```
dholms, line 1: silly naming here cause eslint doesn't support the syntax
```ts
export * as api from './api'
```
& i couldn't just disable it

src/common/api.ts
```diff
@@ -21,3 +21,7 @@ export async function didKey(): Promise<string> {
 
   return didCache.didKey
 }
+
+export default {
+  didKey
+}
```
icidasset, line 27: Just out of curiosity, why are we doing this again?

dholms: let's us do 
```ts
import { api } from '../common'
// or
import api from '../common/api'
//instead of 
import * as api from '../common/api'
```

Not necessary, I just personally like the semantics better. Also works better if we ever need to mock it (jest mocks only really seem to work at the module level)

PR #37, opened 2020-04-24
fixed bug around removing root dir

dholms:
## Problem
Removing a file directly off the tree root does not work

## Solution
- Fix bug in `rmNested`
- give better feedback (throw error) if trying to `addChild` with no path


icidasset:
Closing in favor of #38 which had these changes as well.

PR #38, opened 2020-04-24
Fix circular dependency

dholms:
## Problem
Circular dependency arising from `common` refactor

## Solution
- fix imports to eliminate circular dependency
- quick reorg of some functions. including renaming `fs/utils` to `fs/operations`


icidasset:
FYI @dholms , this PR has the changes from #37 in it as well.

PR #39, opened 2020-05-06
Add a pretty tree alongside the public tree

icidasset:
Currently we have the following structure:

```shell
/index/public
/index/public/dir/index/item/index
/index/private
```

This PR adds:

```shell
/index/pretty
/index/pretty/dir/item
```

It does this by having a copy of the public tree, which always has the version `v0`.

---

I did a quick test, which you can find here:
https://icidasset-test-3.fission.name/index/pretty/Testing%20Public

### DIFF
--- a/src/fs/filesystem.ts
+++ b/src/fs/filesystem.ts
@@ -9,6 +9,15 @@ import keystore from '../keystore'
 import user from '../user'
 
 
+type ConstructorParams = {
+  root: PublicTree,
+  publicTree: PublicTree,
+  prettyTree: PublicTree,
+  privateTree: PrivateTree,
+  key: string
+}
+
+
 export class FileSystem {
 
   root: PublicTree
@@ -19,13 +28,7 @@ export class FileSystem {
 
   private key: string
 
-  constructor(
-    root: PublicTree,
-    publicTree: PublicTree,
-    prettyTree: PublicTree,
-    privateTree: PrivateTree,
-    key: string
-  ) {
+  constructor({ root, publicTree, prettyTree, privateTree, key }: ConstructorParams) {
     this.root = root
     this.publicTree = publicTree
     this.prettyTree = prettyTree
@@ -38,13 +41,19 @@ export class FileSystem {
     const { keyName = 'filesystem-root', version = semver.latest } = opts
 
     const root = await PublicTree.empty(version)
-    const publicTreeInstance = await PublicTree.empty(version)
-    const prettyTreeInstance = await PublicTree.empty(semver.v0)
+    const publicTree = await PublicTree.empty(version)
+    const prettyTree = await PublicTree.empty(semver.v0)
 
-    const privateTreeInstance = await PrivateTree.empty(version)
+    const privateTree = await PrivateTree.empty(version)
     const key = await keystore.getKeyByName(keyName)
 
-    return new FileSystem(root, publicTreeInstance, prettyTreeInstance, privateTreeInstance, key)
+    return new FileSystem({
+      root,
+      publicTree,
+      prettyTree,
+      privateTree,
+      key
+    })
   }
 
   static async fromCID(cid: CID, opts: FileSystemOptions = {}): Promise<FileSystem | null> {
@@ -61,7 +70,13 @@ export class FileSystem {
 
     if (publicTree === null || privateTree === null) return null
 
-    return new FileSystem(root, publicTree, prettyTree, privateTree, key)
+    return new FileSystem({
+      root,
+      publicTree,
+      prettyTree,
+      privateTree,
+      key
+    })
   }
 
   static async forUser(username: string, opts: FileSystemOptions = {}): Promise<FileSystem | null> {
@@ -76,12 +91,19 @@ export class FileSystem {
     const { keyName = 'filesystem-root', version = semver.latest } = opts
 
     const root = await PublicTree.empty(version)
-    const pubTreeInstance = await PublicTree.fromCID(cid)
-    const pretTreeInstance = await PublicTree.fromCID(cid, semver.v0)
-    const privTreeInstance = await PrivateTree.empty(version)
+    const publicTree = await PublicTree.fromCID(cid)
+    const prettyTree = await PublicTree.fromCID(cid, semver.v0)
+    const privateTree = await PrivateTree.empty(version)
 
     const key = await keystore.getKeyByName(keyName)
-    return new FileSystem(root, pubTreeInstance, pretTreeInstance, privTreeInstance, key)
+
+    return new FileSystem({
+      root,
+      publicTree,
+      prettyTree,
+      privateTree,
+      key
+    })
   }
 
   async ls(path: string): Promise<Links> {
--- a/src/fs/filesystem.ts
+++ b/src/fs/filesystem.ts
@@ -40,7 +40,7 @@ export class FileSystem {
   static async empty(opts: FileSystemOptions = {}): Promise<FileSystem> {
     const { keyName = 'filesystem-root', version = semver.latest } = opts
 
-    const root = await PublicTree.empty(version)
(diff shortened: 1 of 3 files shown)


expede:
@icidasset 

`/index/pretty/dir/item`

Any thoughts on just making it `/pretty/dir/item`?

i.e. the top level them becomes

```
boris.fission.name
  |
  +--Public
  |     |
  |     +--index
  |          |
  |          +--dir
  |              |
  |              +--index
  |                    |
  |                    +--item
  |                         |
  |                         +--data
  |                             |
  |                             +--+
  |                                |
  +--Pretty                        |
  |     |                          |
  |     +--dir                     |
  |          |                     |
  |          +--item               |
  |               |                |
  |               +-------------Qm12345
  |
  +--Public (encrypted)
```

src/fs/filesystem.ts
```diff
@@ -13,54 +13,75 @@ export class FileSystem {
 
   root: PublicTree
   publicTree: PublicTree
+  prettyTree: PublicTree
   privateTree: PrivateTree
   syncHooks: Array<SyncHook>
```
expede, line 18: 👍

src/fs/filesystem.ts
```diff
@@ -13,54 +13,75 @@ export class FileSystem {
...
+
     const privLink = root.findLink('private')
     const key = await keystore.getKeyByName(keyName)
     const privateTree = privLink ? await PrivateTree.fromCIDWithKey(privLink.cid, key) : null
-    if (publicTree === null || privateTree === null) {
-      return null
-    }
-    return new FileSystem(root, publicTree, privateTree, key)
+
+    if (publicTree === null || privateTree === null) return null
```
expede, line 62: 👍

src/fs/filesystem.ts
```diff
@@ -13,54 +13,75 @@ export class FileSystem {
...
     const { keyName = 'filesystem-root', version = semver.latest } = opts
+
     const root = await PublicTree.empty(version)
     const pubTreeInstance = await PublicTree.fromCID(cid)
+    const pretTreeInstance = await PublicTree.fromCID(cid, semver.v0)
     const privTreeInstance = await PrivateTree.empty(version)
+
     const key = await keystore.getKeyByName(keyName)
-    return new FileSystem(root, pubTreeInstance, privTreeInstance, key)
+    return new FileSystem(root, pubTreeInstance, pretTreeInstance, privTreeInstance, key)
```
expede, line 84: Let's just type out the words. Yes, it'll be longer, but more readable 4 months from now when we have to dig into this fresh again.

Also FWIW, I don't think we need `Instance` on every one of these.

```js
return new FileSystem(root, publicTree, prettyTree, privateTree, key)
```

Also can't hurt to make the constructor take a map for named arguments.

src/fs/filesystem.ts
```diff
@@ -109,12 +130,15 @@ export class FileSystem {
 
   async sync(): Promise<CID> {
     const pubCID = await this.publicTree.put()
+    const pretCID = await this.prettyTree.put()
     const privCID = await this.privateTree.putEncrypted(this.key)
     const pubLink = link.make('public', pubCID, false)
+    const pretLink = link.make('pretty', pretCID, false)
     const privLink = link.make('private', privCID, false)
```
expede, line 137: See above about full names. This will get unwieldy as we add stuff.

src/fs/filesystem.ts
```diff
@@ -138,27 +162,44 @@ export class FileSystem {
 
   async runOnTree<a>(
     path: string,
-    updateTree: boolean,
+    updateTree: boolean, // ie. do a mutation
```
expede, line 165: I'm actually unclear — do a mutation where? On IPFS, locally, remote IPFS?

May be clearer as `updateOnIPFS` or `update

There's a code smell here for separation of concerns. Command/query separation.

src/fs/filesystem.ts
```diff
@@ -138,27 +162,44 @@ export class FileSystem {
...
       result = await fn(this.publicTree, relPath)
+
       if (updateTree && PublicTree.instanceOf(result)) {
+        resultPretty = await fn(this.prettyTree, relPath)
+
         this.publicTree = result
+        this.prettyTree = resultPretty as unknown as PublicTree
       }
+
     } else if (head === 'private') {
```
expede, line 185: This looks like a job for OO classes

src/fs/filesystem.ts
```diff
@@ -138,27 +162,44 @@ export class FileSystem {
...
         this.privateTree = result
       }
-      return result
+
+    } else if (head === 'pretty' && updateTree) {
+      throw new Error("The pretty path is read only")
+
+    } else if (head === 'pretty') {
+      result = await fn(this.prettyTree, relPath)
+
```
expede, line 197: Same with all of this

src/fs/public/tree.ts
```diff
@@ -38,9 +38,9 @@ class PublicTree implements Tree {
     return new PublicTree({}, version)
   }
 
-  static async fromCID(cid: CID): Promise<PublicTree> {
-    const version = await normalizer.getVersion(cid)
-    const { links }  = await normalizer.getTreeData(cid)
+  static async fromCID(cid: CID, version?: SemVer): Promise<PublicTree> {
+    version = version || await normalizer.getVersion(cid)
```
expede, line 42: 👍

icidasset:
@expede Haven't looked at that, I assume Daniel set it up this way for a reason. But seeing that the root is just another `PublicTree`, I guess I could just have that set to `v0` as well (to get rid of that `index` prefix). Unless... do we need any metadata for the root? 🤔

src/fs/filesystem.ts
icidasset: > Also FWIW, I don't think we need Instance on every one of these.

Why not? 🤔 

> Also can't hurt to make the constructor take a map for named arguments.

Agreed 👍

expede:
> do we need any metadata for the root?

We probably do (for consistency) on `Public`, but don't need it on `Pretty`

expede:
> I guess I could just have that set to v0 as well

💯 👍

src/fs/filesystem.ts
icidasset: A mutation both locally and on IPFS, and if the user has set up a `syncHook`, the DNSLink as well. Yeah might refactor this later.

src/fs/filesystem.ts
icidasset: Yeah possibly, not sure I'd fix this here though. This `head` part is part of the path the user provides, for example, `fs.ls("private/dir")`. We might need to revisit this in another PR and do something like `fs.ls("dir", { mode: fs.modes.private })`.

icidasset:
> We probably do (for consistency) on Public, but don't need it on Pretty (see ASCII tree at the top of the comment thread)

Oh yeah for sure, I meant metadata on the top-level root.
In other words, do we need `/headers` next to `/public`, `/pretty` and `/private`?

expede:
> In other words, do we need /headers next to /public, /pretty and /private?

Ah, I see what you're asking. No, we don't need them as siblings, only as children.

```
   Public
   |    |
index  header
```

---

The other thing to perhaps mention (and that I don't know where Daniel landed on) is if there's a separate `header` directory, or if `index` is in the same layer.

Layout 1:

```
   Public
   |    |
index  header
 |     |    |
...  sizes  tags
       |    |
      ...  ...
```

Layout 2:

```
       Public
         |
   +-----+------+
   |     |      |
index  sizes  tags
 |       |      |
...     ...    ...
```

icidasset:
Ok 👍 Will change that.
We currently have layout 2 (no `headers` directory).

expede:
> We currently have layout 2 (no headers directory).

Cool, I think that makes sense 👍

PR #44, opened 2020-05-27
added maybe type

dholms:
## Problem
A lot of our code had an optional param `key?: string` which would optionally encrypt data before uploading. This could cause issues of unwittingly not passing the key variable along with no warning from the compiler.

## Solution
Added a `Maybe` type so that this is explicit

_haskell intensifies_


dholms:
yeahhh I thought about having a `const Nothing = null` variable and maybe a `type Just<T> = T` but I figured this would just be bulky and wouldn't give us much since there's no pattern matching or binds in TS... :man_shrugging:

PR #49, opened 2020-06-11
UX Overhaul

icidasset:
Improves the user experience of the SDK, besides the file system.

Implements the following API, except those two filesystem functions (need feedback on that):
https://talk.fission.codes/t/initial-public-sdk-interface/688

Notes:
- Includes most of the in-code documentation, type docs would ideally be generated by typescript.
- Should technically work in a web worker, except the two functions indicated that only work on the main thread.
- Example usage: https://github.com/fission-suite/drive/blob/new-sdk/src/Javascript/index.js#L34


src/common/index.ts
```diff
@@ -1,21 +1,10 @@
...
+export { base64, blob }
 
-export default {
-  base64,
-  blob,
-  ...constants,
-  ...types,
-  ...typeChecks,
-  ...util
-}
```
icidasset, line 21: I removed these in a bunch of places, not in `fs/*` or `ipfs/*` (I think). Because I feel like we're doing additional work here, we should probably do one or the other. Either use `export function ...` or `export default { ... }`, no?

src/fs/normalizer/header.ts
```diff
@@ -76,10 +73,15 @@ export const getChildKey = async (cid: CID, key: string): Promise<string> => {
  * ```
  */
 export const put = async (index: CID, header: Header, key: Maybe<string>): Promise<CID> => {
-  const noUndefined = _.pickBy(header, isDefined)
+  const noUndefined = Array.from(
```
dholms, line 76: I.... thought I tried this and typescript didn't like whatever I did :thinking: 

Ah well, if it works then I'm happy to eliminate the dependency!

src/common/index.ts
dholms: :man_shrugging: I personally like the flexibility of either being able to import an entire module or just a function/variable from it. But you're right, it is more work, and it's very easy for your `default` export to get out of sync with the module. I don't have a strong opinion on it. I'm good with leaving it off unless we run into a situation where it would help

src/data-root.ts
```diff
@@ -0,0 +1,46 @@
...
+import { UCAN_STORAGE_KEY } from './common'
+
+
+/**
+ * Get the CID of a user's data root.
+ */
+export async function dataRoot(username: string): Promise<CID> {
+  try {
+    // TODO: This'll be `files.${username}.fission.name` later
+    return await dns.lookupDnsLink(`${username}.fission.name`)
```
dholms, line 15: such a small nitpick, but: you don't need the `await` here

src/data-root.ts
```diff
@@ -0,0 +1,46 @@
...
+  } catch(err) {
+    throw new Error("Could not locate user root in dns")
+  }
+}
+
+/**
+ * Update a user's data root.
+ */
+export const updateDataRoot = async (props: {
```
dholms, line 24: thoughts on making the params `(cid: CID, props: {...})`, since `cid` is required?

Otherwise, I'd at least bump it up to the top of the `props` so it's obvious that it's the most important param

src/data-root.ts
```diff
@@ -0,0 +1,46 @@
...
+  apiEndpoint?: string
+  apiDid?: string
+  cid: CID | string
+}): Promise<any> => {
```
dholms, line 28: Maybe give this a type? `any`s kinda blow apart TS type system. I'm trying to steer clear of them except for in runtime type checks (and even then I'd like to start using `unknown`, but I don't think I'm doing it correctly cause TS always complains :stuck_out_tongue:)

dholms: actually, does this even need to return anything?

src/dns/index.ts
```diff
@@ -9,9 +15,15 @@ export function lookupTxtRecord(host: string): Promise<string> {
   })
   .then(r => r.json())
   .then(r => r.Answer[0].data)
+  .then(r => r && r.replace(/^"|"$/g, ""))
```
dholms, line 18: I'm notoriously terrible at understanding regex :laughing: 
maybe add a short comment of what this does (I think it strips quotes, right?)

src/lobby/index.ts
```diff
@@ -0,0 +1,64 @@
...
+
+import { USERNAME_BLOCKLIST } from './blocklist'
+
+
+/**
+ * Create a user account.
+ */
+export const createAccount = async (
+  userProps: { email: string, username: string },
+  apiEndpoint?: string,
```
dholms, line 13: maybe wrap this in a `props` object similar to `updateDataRoot`. so a user could supply `apiDid` and not `apiEndpoint` (altho I'm not sure if that'd ever happen :thinking:)

src/lobby/index.ts
```diff
@@ -0,0 +1,64 @@
...
+): Promise<any> => {
+  apiDid = apiDid || await core.apiDid()
+  apiEndpoint = apiEndpoint || core.apiEndpoint()
+
+  const jwt = await core.ucan({
+    audience: apiDid,
+    issuer: await core.did(),
+  })
+
+  return fetch(`${apiEndpoint}/user`, {
```
dholms, line 24: Something I've been thinking about is that we're really prioritizing browsers right now. (which makes sense, we're building a browser-native sdk), but there's nothing stopping a Node application from using our sdk in the future. All of our crypto stuff is only available in browsers so we're not Node-compatible by any means right now, but eventually we'll probably want to switch this over to `axios` or something similar. 

Anyways nothing to do here atm, I'm just rambling :stuck_out_tongue:

src/lobby/index.ts
```diff
@@ -0,0 +1,64 @@
...
+
+
+/**
+ * Create a user account.
+ */
+export const createAccount = async (
+  userProps: { email: string, username: string },
+  apiEndpoint?: string,
+  apiDid?: string
+): Promise<any> => {
```
dholms, line 15: I'd wrap this in a type as well^^ or just eliminate the return value

dholms:
Left a bunch of comments but they're mostly really small things. Overall this looks great. And a very friendly api!

src/common/index.ts
icidasset: Yeah, I'd just use `import * as module from` to import an entire module. That way you don't need to maintain that list. `import { somethin }, * as module from` works too.

src/fs/normalizer/header.ts
icidasset: I had to add `val as any` a few lines further down 🤷

src/data-root.ts
icidasset: I'm using `async` to force it into throwing an error. Could have used `.catch()` alternatively, but felt like doing this 🤷

src/data-root.ts
icidasset: I keep forgetting that the target is javascript and not typescript 😅 When using typescript, the current implementation should require you to pass { cid = ... } right? (because there's no `?` after the key)

src/data-root.ts
icidasset: No it doesn't. I wanted to return `Promise<void>` or something, but I don't know how to do that. Any ideas?

src/dns/index.ts
icidasset: Hehe will do. It strips the quotes from the start and end, yeah.

src/lobby/index.ts
icidasset: Oh good catch, will do.

src/common/index.ts
dholms: makes sense to me. I'm happy switching over to this convention :ok_hand:

src/fs/normalizer/header.ts
dholms: ah yup didn't see that. hmmm I really like to avoid `any`s, altho this one is fairly contained. I wonder why TS won't cooperate :thinking:

src/data-root.ts
dholms: ah didn't realize. good looks :+1:

src/data-root.ts
dholms: > I keep forgetting that the target is javascript and not typescript

yup same here haha :sweat_smile: 

> When using typescript, the current implementation should require you to pass { cid = ... } right?

yup! but I still think it might be a bit more intuitive as it's param, since most of our sdk is structured that way

src/data-root.ts
dholms: I'm pretty sure you can do just that^^ 

if it throws a fit just an empty
```ts
return
```
at the end of the fn should do it

src/fs/normalizer/header.ts
icidasset: Managed to replace `as any` with `as FileContent`, which fixes it 👍

dholms: awww yisss

src/data-root.ts
icidasset: Oh ok, neato, fixed it by using `await` and no return.

src/data-root.ts
icidasset: Updated 👍

PR #50, opened 2020-06-14
Add keys tree

dholms:
## Problem
Right now the FS is being loaded by looking up a `keyName` in `keystore-idb`. This will not work across browsers and IDB could be cleared by the browser and the user would lose access to their files

## Solution
- Add a `/keys` tree to the root of the filesystem
- Add a `shareWith` function to the filesystem that encrypts the root key of a filesystem for a DID's public key and adds it to `/keys/${did}`
- Share filesystem with self when creating
- when loading a filesystem `fromCID`, check if a key has been shared with the user's DID, decrypt it, and load the private tree with that key
- reorganized some of the functions in `core` as we're adding more DID functions

## Open questions
Each user has a read key and a write key. We're storing the write key in dns. How do we publicize public read keys? Two separate DNS entries?

How to we give sub access to a user? Instead of just a key stored in `/keys`, we can also include the related cid. Add this feature now? Or in the future?

### DIFF
--- a/src/fs/filesystem.ts
+++ b/src/fs/filesystem.ts
@@ -1,20 +1,21 @@
-import PublicTreeBare from './bare/tree'
+import BareTree from './bare/tree'
 import PublicTree from './v1/PublicTree'
 import PrivateTree from './v1/PrivateTree'
-import { File, Tree, Links, SyncHook, FileSystemOptions, HeaderTree } from './types'
+import { File, Tree, Links, SyncHook, FileSystemOptions, HeaderTree, PinMap } from './types'
 import { CID, FileContent } from '../ipfs'
 import { dataRoot } from '../data-root'
 
 import * as keystore from '../keystore'
-import pathUtil from './path'
-import { asyncWaterfall } from '../common/util'
+import * as pathUtil from './path'
+import { pinMapToLinks } from './pins'
 
 
 type ConstructorParams = {
   root: Tree
   publicTree: HeaderTree
-  prettyTree: PublicTreeBare
+  prettyTree: BareTree
   privateTree: HeaderTree
+  pinTree: BareTree
 }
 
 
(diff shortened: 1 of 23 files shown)


src/fs/filesystem.ts
```diff
@@ -157,11 +187,20 @@ export class FileSystem {
...
+    this.keyTree.add(did, encryptedRootKey)
+    return this.sync()
+  }
+
   async sync(): Promise<CID> {
     this.root = await asyncWaterfall(this.root, [
       (t: Tree) => t.addChild('public', this.publicTree),
       (t: Tree) => t.addChild('pretty', this.prettyTree),
-      (t: Tree) => t.addChild('private', this.privateTree)
+      (t: Tree) => t.addChild('private', this.privateTree),
```
dholms, line 202: the more trees we add the more needlessly expensive `sync`s get...

Mostly a note to myself: I'd like to rewrite this logic soon

dholms: Okay changed this around to update the root node every time a child is updated so that `sync` only has to `put` the root

src/fs/v1/PublicFile.ts
```diff
@@ -43,7 +43,7 @@ export class PublicFile extends BaseFile implements HeaderFile {
   }
 
   protected async putWithKey(key: Maybe<string>) {
-    const { cid, size } = await basic.putFile(this.content, this.header.key)
+    const { cid, size } = await basic.putFileEncoded(this.content, this.header.key)
```
dholms, line 46: wait this will add different file content to the `public` and `pretty` trees since one is encoded and one isn't :thinking: 

Not to self to fix this

dholms:
@expede this one is really about half done. It currently only shares the root of a filesystem, not a private subtree. And it only has logic for the "shared be me" key tree not the "shared with me" tree.

I'd use this branch for context & ideas, but don't merge it without a decent rewrite (which will probably take place after restructuring the FS)

dholms:
Closing due to FS rewrite

PR #51, opened 2020-06-16
0.8.0

icidasset:
Changes:
- UCAN v0.2.0 (`scp` -> `rsc`, https://talk.fission.codes/t/ucan-v0-2-0-now-with-scoped-resources/700)
- Allows passing `fission-suite/get-ipfs` options to `ipfs.getIpfs()`
- Automatic DID lookup for other API instances (eg. our staging instance)
- Removed the `apiDid` option from various functions
- Moved the api functions to `src/common`
- Allows passing the domain to the `dataRoot` function (eg. our staging instance)
   And as a consequence `isUsernameAvailable` as well, which uses the `dataRoot` function.
- Added the `ucanRootIssuer` function
- Couple typo fixes and small stuff here and there


src/ipfs/config.ts
```diff
@@ -13,11 +14,9 @@ export const setIpfs = (userIpfs: unknown): void => {
...
 }
 
-export const getIpfs = async (): Promise<IPFS> => {
-  if (ipfs) {
-    return ipfs
-  }
-  ipfs = (await getIpfsWithCfg(IPFS_CFG)) as unknown as IPFS
+export const getIpfs = async (options = {}): Promise<IPFS> => {
+  if (ipfs) return ipfs
+  ipfs = (await getIpfsWithCfg({ ...IPFS_CFG, ...options })) as unknown as IPFS
```
icidasset, line 19: I'm using the following in Drive now to setup IPFS:
```js
  await sdk.ipfs.getIpfs({
    permissions: [
      "cat",
      "dag.tree",
      "files.catPullStream",
      "files.catReadableStream",
      "id",
      "ls",
      "swarm.connect",
      "version",
    ],

    jsIpfs: "./web_modules/ipfs.min.js"
  })
```
@dholms Do you think that's fine? I'm a bit worried that `getIpfs` will be called on "script load" and my options won't go through, but I think that's not the case?

src/lobby/index.ts
```diff
@@ -14,32 +15,37 @@ export const createAccount = async (
   },
   options: {
     apiEndpoint?: string
-    apiDid?: string
   } = {}
-): Promise<void> => {
-  const apiDid = options.apiDid || await core.apiDid()
-  const apiEndpoint = options.apiEndpoint || core.apiEndpoint()
+): Promise<{ success: boolean }> => {
```
dholms, line 19: why wrap this in an object?

src/ipfs/config.ts
dholms: Yes this should be fine :+1: 

Two notes:
- permissions are only necessary for ipfs companion. the js-ipfs daemon is automatically permissioned (since it's sandboxed by the browser)
- `ipfs` will be set by whatever calls `getIpfs` first. So if you want a custom config passed, just make sure you call `sdk.ipfs.getIpfs` before doing any filesystem actions/ anything from the sdk that uses ipfs

src/common/api.ts
```diff
@@ -0,0 +1,45 @@
+import * as dns from '../dns'
```
dholms, line 1: Why move all of this to `common` instead of `core`? Especially with our other `did` stuff still in `core`? 

I actually split out a lot of `did` stuff into `core/did` in the `keys` PR that I have up right now: https://github.com/fission-suite/ts-sdk/pull/50

src/lobby/index.ts
icidasset: May add an error message later.

dholms: ah cool :+1:

src/common/api.ts
icidasset: I moved it to `./common` because it's not so much to do with DIDs, it's just a lookup for a DID, specifically related to Fission APIs.

Your code looks good 👍

dholms: okay cool makes sense :+1:

PR #54, opened 2020-06-17
Pins for private file tree

dholms:
## Problem
- Our server needs to pin a user's private file tree, but it can't crawl it because it's..... private
  - right now this is accomplished with loose pins which are cumbersome to deal with
- when a subtree is shared with a user, they should retain access to it even if the CID changes. So they need to be able to look up the current CID of the tree by using a given filepath

## Solution
- Add a single-layer /pins node off of the FS root
- That node will include a list of all nodes in the private file system. The names of the nodes will be a hash of the name plus their DID which acts as a salt: sha(filename + did)
  
An example snippet of this:

{ name: sha("private/index/file" + did), cid: "Qm123..." },
{ name: sha("private/index/file/index" + did), cid: "Qm456..." },
{ name: sha("private/index/file/isFile" + did), cid: "Qm789..." },
{ name: sha("private/index/file/mtime" + did), cid: "Qmabc..." },
...

Closes https://github.com/fission-suite/ts-sdk/issues/52

### DIFF
--- a/src/common/crypto.ts
+++ b/src/common/crypto.ts
@@ -1,0 +1,16 @@
+import utils from 'keystore-idb/utils'
+
+export const sha256 = async (buf: ArrayBuffer): Promise<ArrayBuffer> => {
+  return window.crypto.subtle.digest(
+    {
+        name: "SHA-256",
+    },
+    buf
+  )
+}
+
+export const sha256Str = async(str: string): Promise<string> => {
+  const buf = utils.strToArrBuf(str, 8)
+  const hash = await sha256(buf)
+  return utils.arrBufToBase64(hash)
+}
--- a/src/core/did.ts
+++ b/src/core/did.ts
@@ -1,0 +1,14 @@
+import * as dns from '../dns'
+
+export const rootDidForUser = async (
+  username: string,
+  domain = 'fission.name'
+): Promise<string> => {
+  try {
+    const maybeDid = await dns.lookupTxtRecord(`_did.${username}.${domain}`)
+    if(maybeDid !== null) return maybeDid
+  } catch (_err) { 
+    // throw error below
+  }
+  throw new Error("Could not locate user DID in dns")
+}
--- a/src/fs/filesystem.ts
+++ b/src/fs/filesystem.ts
@@ -1,23 +1,26 @@
-import { throttle } from 'throttle-debounce';
+import { throttle } from 'throttle-debounce'
 
-import PublicTreeBare from './bare/tree'
+import BareTree from './bare/tree'
 import PublicTree from './v1/PublicTree'
 import PrivateTree from './v1/PrivateTree'
-import { File, Tree, Links, SyncHook, FileSystemOptions, HeaderTree } from './types'
+import { File, Tree, Links, SyncHook, FileSystemOptions, HeaderTree, PinMap } from './types'
 import check from './types/check'
-import pathUtil from './path'
+import * as pathUtil from './path'
 
 import * as dataRoot from '../data-root'
 import * as keystore from '../keystore'
 import { CID, FileContent } from '../ipfs'
 import { asyncWaterfall } from '../common/util'
+import { pinMapToLinks } from './pins'
 
 
 type ConstructorParams = {
   root: Tree
   publicTree: HeaderTree
-  prettyTree: PublicTreeBare
+  prettyTree: BareTree
   privateTree: HeaderTree
+  pinsTree: BareTree
+  rootDid: string
 }
 
 
@@ -25,19 +28,21 @@ export class FileSystem {
 
   root: Tree
   publicTree: HeaderTree
-  prettyTree: PublicTreeBare
+  prettyTree: BareTree
   privateTree: HeaderTree
-
+  pinsTree: BareTree
+  rootDid: string
   syncHooks: Array<SyncHook>
   syncWhenOnline: CID | null
 
 
-  constructor({ root, publicTree, prettyTree, privateTree }: ConstructorParams) {
+  constructor({ root, publicTree, prettyTree, privateTree, pinsTree, rootDid }: ConstructorParams) {
     this.root = root
     this.publicTree = publicTree
     this.prettyTree = prettyTree
     this.privateTree = privateTree
-
+    this.pinsTree = pinsTree
+    this.rootDid = rootDid
     this.syncHooks = []
     this.syncWhenOnline = null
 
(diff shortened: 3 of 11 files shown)


src/fs/filesystem.ts
```diff
@@ -85,27 +93,6 @@ export class FileSystem {
     return FileSystem.fromCID(cid, opts)
   }
 
-  /**
-   * Upgrade public IPFS folder to FileSystem
```
dholms, line 89: I'm removing this for now. It doesn't currently work with HeaderTrees & there were some questions on if we'd even want to offer this functionality.

src/fs/filesystem.ts
```diff
@@ -145,24 +132,14 @@ export class FileSystem {
...
-    ]
+  async updatePinTree(pins: PinMap): Promise<void> {
+    const pinLinks = await pinMapToLinks('private', pins) // @@TODO: add salt
+    console.log('pinLinks: ', pinLinks)
+    this.pinTree = BareTree.fromLinks(pinLinks)
+    await this.root.addChild('pins', this.pinTree)
   }
 
   async sync(): Promise<CID> {
-    this.root = await asyncWaterfall(this.root, [
```
dholms, line 160: `sync` was getting more expensive with each branch we added so I moved the updates to `root` to the locations in the code where a given tree was actually being updated

src/fs/filesystem.ts
```diff
@@ -201,14 +178,19 @@ export class FileSystem {
         resultPretty = await fn(this.prettyTree, relPath)
 
         this.publicTree = result
-        this.prettyTree = resultPretty as unknown as PublicTreeBare
+        this.prettyTree = resultPretty as unknown as BareTree
```
dholms, line 181: This method keep getting bigger, but I'm not sure of a better way to do it. Will keep thinking :thinking:

src/fs/filesystem.ts
```diff
@@ -145,24 +132,14 @@ export class FileSystem {
...
-    const privateResult = await this.privateTree.putWithPins()
-    const publicResult = await this.publicTree.putWithPins()
-    const rootCID = await this.sync()
-    return [
-      ...privateResult.pins,
-      ...publicResult.pins,
-      rootCID
-    ]
+  async updatePinTree(pins: PinMap): Promise<void> {
+    const pinLinks = await pinMapToLinks('private', pins) // @@TODO: add salt
```
dholms, line 136: Still need to add salt, but first the Filesystem needs to be aware of a user's DID. Which means that we probably can't just instantiate it from a `cid`...

dholms: ok added salt
Now if you instantiate an FFS `fromUsername`, it also stores your DID which it can use for a salt. If you just instantiate an FFS `fromCID` or `empty` then takes an optional param `rootDID` and defaults to an empty string

src/common/crypto.ts
```diff
@@ -0,0 +1,16 @@
...
+  return window.crypto.subtle.digest(
+    {
+        name: "SHA-256",
+    },
+    buf
+  )
+}
+
+export const sha256Str = async(str: string): Promise<string> => {
+  const buf = utils.strToArrBuf(str, 8)
```
icidasset, line 13: Should we create `utils.strToUtf8Buf(str)` and `utils.strToUtf16Buf(str)` instead? Might be more helpful?

src/core/did.ts
```diff
@@ -0,0 +1,12 @@
...
+
+export const rootDIDForUser = async (
+  username: string,
+  domain = 'fission.name'
+): Promise<string> => {
+  try {
+    return dns.lookupTxtRecord(`_did.${username}.${domain}`)
+  } catch (err) {
+    throw new Error("Could not locate user DID in dns")
+  }
```
icidasset, line 11: Nice, was just about to add this as well 👍 
Just an FYI, I just improved `dns.lookupTxtRecord` a bit, now joins multiple TXTs, but now it can return `null` as well.
So instead of try catch, you'll need to do a `null` check.

Nitpick, should we standardise on uppercase DID or camelcase in func/var names? 🤔 I've been doing camelcase like `lookupTxtRecord` instead of `lookupTXTRecord`. Slightly prefer camelcase, but fine with either.

PS. Just realised I haven't been using the `Maybe` type, sorry about that, need to get on that 👀

src/fs/network/header.ts
```diff
@@ -87,12 +77,21 @@ export const put = async (
...
+  const pins = {
     ...pinsForHeader,
-    ...pinMapToList(header.pins),
-    cid
-  ]
+    index: {
+      ...header.pins,
+      '': pinsForHeader.index
+    },
+    '': cid
```
icidasset, line 92: What's this thing here with the empty string as object key?

src/fs/filesystem.ts
icidasset: Did we end up devising a way to get content of a CID into the private part of the filesystem?

dholms: Not yet. We'll just have to do that manually. The code that was there hasn't worked since we upgraded to `HeaderTree`s. If a user wants to upgrade some IPLD content to FLOOFS, I think we'll need to recursively walk that content and re-add it as FLOOFS objs

src/common/crypto.ts
dholms: So:
- `utils.strToUtf8Buf(str) = utils.strToArrBuf(str, 8)` and
- `utils.strToUtf16Buf(str) = utils.strToArrBuf(str, 16)`?

Sure! I don't see a huge difference, but it might read a bit better

src/core/did.ts
dholms: good looks^^ I'll fix that on rebase

> Nitpick, should we standardise on uppercase DID or camelcase in func/var names? thinking I've been doing camelcase like lookupTxtRecord instead of lookupTXTRecord. Slightly prefer camelcase, but fine with either

Yeah I think I like camelcase better actually. I'll start standardizing around that :ok_hand:

src/fs/network/header.ts
dholms: This is a map of all the CIDs needed to pin some objects so it might look like 
```ts
{
  mtime: "Qm...",
  size: "Qm...",
  index: { ... },
  "": "Qm..."
}
```
so the empty key just means its the CID for _that_ object. We could also just do a single slash (`/`)

_buuutt_ I think this stuff is going to be rewritten for the FLOOFS restructure that @expede and I are talking about: https://talk.fission.codes/t/floofs-restructuring/718/4

dholms:
@expede this one is essentially good to go (after approval). I addressed Stevens comments and & pushed some changes for them. 

The pinlist that we're using here is sort of a half-implemented version of the restructuring that you're going to be doing. This PR uses a `pinsTree` to track private tree pins. But with the restructure, that functionality is going to be rolled into the structure of the privateTree itself. 

So again I'll leave it up to you if you'd rather merge this, or use it for context + ideas and close the PR

src/fs/filesystem.ts
dholms: @expede, this is the method that I was talking about that gets more convoluted as we add more branches to the tree. I think we need some better abstraction here

src/common/crypto.ts
icidasset: Yeah that'd be great. My reasoning is that, this way you can't use any `charSize` that doesn't make sense 🤷

dholms: yeah that makes sense to me

PR #55, opened 2020-06-19
Encrypt metadata with node's key

dholms:
## Problem
Currently metadata for a node is encrypted with that node's parent key (the key used to encrypt that node), not the node's own key (the key contained as a header value in that node). Thus, when a user is given access to a file, they don't get access to the metadata that goes with it

## Solution
Encrypt metadata with the node's ownKey instead of the parentKey

Closes https://github.com/fission-suite/ts-sdk/issues/53

### DIFF
--- a/src/fs/network/header.ts
+++ b/src/fs/network/header.ts
@@ -39,6 +39,7 @@ export const getVersion = async (cid: CID, key: Maybe<string>): Promise<SemVer> 
   return checkValue(version, isSemVer)
 }
 
+
 /**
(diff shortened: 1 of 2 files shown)


dholms:
@expede this one is good to go. The included code will probably be rewritten in the restructuring that we're talking about, but the concept of encrypting a node's metadata with that node's key (instead of the parent key) will stick around. I'll leave it up to you whether it's more helpful to merge or close this one

dholms:
closing due to FS rewrite

expede:
@icidasset can you confirm that this is complete?

PR #58, opened 2020-06-29
Brooke's Familiarizing Thread 🐈🧙‍♀️

expede:
I'm purely using this branch as a way to make notes and get familiar with the code. Also a good way to communicate async about why things are the way they are.

I want to clarify that most/many of these questions are asked humbly, and are not criticisms. Really just looking for clarity on design decisions long before potentially suggesting anything.

---

Current game plan:

* Break out multiple (abstract) interfaces
* Parse, don't validate
* ~~`async empty()` -> `constructor({...} = {})`~~
  * ~~Drop the async here, too~~
    * Yeah, need the `async` because key generation. Womp womp.
  * Rearrange AES keys to _not reference_ parent
    * Multiple options here, including passing them in as an explicit parameter, changing the keys, &c
* Abstractions
  * Likely tree operations, such as `walk`, `map`, and `fold`
  * Serialization, serialization functions (plural)
    * Break out as a _separate_ step
    * Render to plain JSON -- helpful for model testing
* Helpers
  * Why not write an `asyncReduce` or `reduce(fn, {async: true}, collection)`?
    * Underneath, it _is just a `for` loop_
* POSSIBLY
  * Fold `pretty` and `FLOOFS` into the same structure, recursively
  * Make the entire tree fully recursive, with each branch having an optional key?
    * Eh, probably not. Very flexible, but fewer invariants. Worth exploring, though.
    * Hmm, yeah. Actually we can have optional pointers from each abstract layer.
    * May not be able to enforce recursive encryption in types, though 🤔
  * Lazy building
    * This may cause more complexity than it solves. Why not just hydrate on pull?
      * I guess it could get large, but like... is that so bad? 10ks of objects I guess 🤔
      * These are isomorphic, so it does keep space under controkl
    * Should be able to `.toJSON` any inode to deterministically get a JSON cache representation
  * _Maybe_ drop the recursive index.json from the public side? It should be pretty small overall. Not sure though.

```haskell
-- Something like:

data FileSystem
  = Abstract Layer (Either Pretty Private)
  | PrivateLayer Private
```

Yes, actually something interesting in that direction:

```haskell
-- Term `Inode` comes from "index node" https://www.wikiwand.com/en/Inode
-- The following is modelled structurally rather than OO. YMMV.

data FileSystem
  | Root (Inode Public) (Inode Private) KeyExchange
  | Public (Inode Public)
  | Private (Inode Private)

data Inode a = Inode
  { meta    :: Metadata a
  , content :: Either RawFile { Text => Inode a }
  }

-- Abusing notation since this is a sketch. Would be a GADT or data family IRL.

data Metadata a = Metadata
  { bytes    :: Natural
  , event    :: Event
  , previous :: Inode a
  , special  :: a -- AES key, pretty CID, and whatnot
  }
```

```haskell
-- partially alternate design

data PublicInode = PublicInode
  { metadata :: Metadata
  , content  :: Either RawFile { Text => PublicInode }
  , direct   :: (FileContent, CID) -- recursive "pretty" path
  }

-- Now the two parts stack directly
```

---

Primitive operations:

* Network
  * Push
  * Pull
* Local
  * Read
  * Write
* Serialization
  * Serialize to IPLD
  * Parse from IPLD
  * Serialize to JSON
  * Parse to JSON
* Inode
  * Mutate Inode links
    * Add link
    * Remove link
    * Rename link
      * Property: same as remove and add updated
  * Query Inode -- actually probably a path transform would be the _primitive_, to call into
                -- or heck, `find` (a path selector)
    * Index (ls)
    * Show (cat)
    * Inspect (meta)
    * Get path segments (e.g. for building IPLD path or pretty URL)
  * Mutate Inode Metadata
    * Add
    * Remove
    * Update
  * Query Metadata
    * Read
  * File
    * Update content
    * Read raw

### DIFF
--- a/src/common/util.ts
+++ b/src/common/util.ts
@@ -56,4 +56,6 @@ export const asyncWaterfall = async <T>(val: T, operations: ((val: T) => Promise
     acc = await operations[i](acc)
   }
   return acc
+
+  // hmmm why not run this in parallel? Promise.all(operations)
 }
--- a/src/common/util.ts
+++ b/src/common/util.ts
@@ -57,5 +57,7 @@ export const asyncWaterfall = async <T>(val: T, operations: ((val: T) => Promise
   }
   return acc
 
-  // hmmm why not run this in parallel? Promise.all(operations)
+  // hmmm why not run this in parallel? Promise.all(operations.map(acc))
+  // Oooooh it's accumulating. Interesting. _Do_ these have a dependency between them?
+  // This function could be a bottleneck, but unsure
 }
--- a/src/common/util.ts
+++ b/src/common/util.ts
@@ -50,6 +50,7 @@ export const arrContains = <T>(arr: T[], val: T): boolean => {
 }
 
 export const asyncWaterfall = async <T>(val: T, operations: ((val: T) => Promise<T>)[]): Promise<T> => {
+  // Why not reduce?
   let acc = val
   for(let i=0; i<operations.length; i++){
     acc = await operations[i](acc)
--- a/src/fs/bare/file.ts
+++ b/src/fs/bare/file.ts
@@ -6,7 +6,7 @@ import BaseFile from '../base/file'
 export class BareFile extends BaseFile {
 
   static create(content: FileContent): BareFile {
-    return new BareFile(content) // why not `super`?
+    return new BareFile(content) // why not the constructor?
   }
 
   static async fromCID(cid: CID): Promise<BareFile> {
--- a/src/fs/bare/file.ts
+++ b/src/fs/bare/file.ts
@@ -6,12 +6,12 @@ import BaseFile from '../base/file'
 export class BareFile extends BaseFile {
 
   static create(content: FileContent): BareFile {
-    return new BareFile(content)
+    return new BareFile(content) // why not `super`?
   }
 
   static async fromCID(cid: CID): Promise<BareFile> {
     const content = await basic.getFile(cid, null)
-    return new BareFile(content)
+    return new BareFile(content) // why not `super`?
   }
 
   async put(): Promise<CID> {
--- a/src/fs/bare/tree.ts
+++ b/src/fs/bare/tree.ts
@@ -17,8 +17,9 @@ class BareTree extends BaseTree {
     this.links = links
   }
 
+  // why is this async? consistent interface? if so, maybe be explicit what we're implementing
   static async empty(): Promise<BareTree> {
-    return new BareTree({})
+    return new BareTree({}) // Why not just part of the constructor as a defautl value?
   }
 
   static async fromCID(cid: CID): Promise<BareTree> {
--- a/src/fs/filesystem.ts
+++ b/src/fs/filesystem.ts
@@ -10,6 +10,7 @@ import pathUtil from './path'
 import { asyncWaterfall } from '../common/util'
 
 
+// This is by far my favourite way of handling TS params
 type ConstructorParams = {
   root: Tree
   publicTree: HeaderTree
@@ -21,8 +22,8 @@ type ConstructorParams = {
 export class FileSystem {
 
   root: Tree
-  publicTree: HeaderTree
-  prettyTree: PublicTreeBare
+  publicTree: HeaderTree // Header? UPDATE: Oooh, a tree with headers. Gotcha.
+  prettyTree: PublicTreeBare // Bare sounds like it's an empty binary tree -- this is not!
   privateTree: HeaderTree
   syncHooks: Array<SyncHook>
 
--- a/src/fs/filesystem.ts
+++ b/src/fs/filesystem.ts
@@ -160,6 +160,7 @@ export class FileSystem {
     ]
   }
 
+  // Maybe rename to pushRemote or other familiar metaphor
   async sync(): Promise<CID> {
     // waterfall? Why not in parallel?
     this.root = await asyncWaterfall(this.root, [
--- a/src/fs/filesystem.ts
+++ b/src/fs/filesystem.ts
@@ -171,6 +171,7 @@ export class FileSystem {
 
     const cid = await this.root.put()
 
+    // What are we using these hooks for?
     this.syncHooks.forEach(hook => {
       hook(cid)
     })
@@ -179,6 +180,7 @@ export class FileSystem {
   }
 
   // Legit don't know what this means: synchronize vs synchronous
+  // I don't see us actually using the sync hooks anywhere?
   addSyncHook(hook: SyncHook): Array<SyncHook> {
     this.syncHooks = [...this.syncHooks, hook]
     return this.syncHooks
--- a/src/fs/filesystem.ts
+++ b/src/fs/filesystem.ts
@@ -81,6 +81,7 @@ export class FileSystem {
     })
   }
 
+  // Nice
   static async forUser(username: string, opts: FileSystemOptions = {}): Promise<FileSystem | null> {
     const cid = await dataRoot(username)
     return FileSystem.fromCID(cid, opts)
@@ -107,6 +108,7 @@ export class FileSystem {
     })
   }
 
+  // Not sure if this is in an interface (will look), but it should! Docs, dependencies, &c.
   async ls(path: string): Promise<Links> {
     return this.runOnTree(path, false, (tree, relPath) => {
       return tree.ls(relPath)
@@ -117,7 +119,7 @@ export class FileSystem {
     await this.runOnTree(path, true, (tree, relPath) => {
       return tree.mkdir(relPath)
     })
-    return this.sync()
+    return this.sync() // Not a huge fan of the term "sync". Had to think about it a few moments to realize that it wasn't "synchronous"
   }
 
   async add(path: string, content: FileContent): Promise<CID> {
@@ -150,6 +152,7 @@ export class FileSystem {
     const privateResult = await this.privateTree.putWithPins()
     const publicResult = await this.publicTree.putWithPins()
     const rootCID = await this.sync()
+    // Minor / inconsequential -- Can be rephrased with destructuring
     return [
       ...privateResult.pins,
       ...publicResult.pins,
@@ -158,6 +161,7 @@ export class FileSystem {
   }
 
   async sync(): Promise<CID> {
+    // waterfall? Why not in parallel?
     this.root = await asyncWaterfall(this.root, [
       (t: Tree): Promise<Tree> => t.addChild('public', this.publicTree),
       (t: Tree): Promise<Tree> => t.addChild('pretty', this.prettyTree),
@@ -173,6 +177,7 @@ export class FileSystem {
     return cid
   }
 
+  // Legit don't know what this means: synchronize vs synchronous
   addSyncHook(hook: SyncHook): Array<SyncHook> {
     this.syncHooks = [...this.syncHooks, hook]
     return this.syncHooks
@@ -192,6 +197,7 @@ export class FileSystem {
     const head = parts[0]
     const relPath = pathUtil.join(parts.slice(1))
 
+    // Why `a` not FileSystem? TS has subtyping. I easily could be missing something.
     let result: a
     let resultPretty: a
 
@@ -212,6 +218,7 @@ export class FileSystem {
         this.privateTree = result
       }
 
+      // NOTE TO SELF: pretty ~ reduction ~ view pattern
     } else if (head === 'pretty' && updateTree) {
       throw new Error("The pretty path is read only")
 
--- a/src/fs/link.ts
+++ b/src/fs/link.ts
@@ -10,6 +10,7 @@ export const toDAGLink = (link: BasicLink): DAGLink => {
   return new dagPB.DAGLink(name, size, cid)
 }
 
+// Consistent naming would be helpful -- fromUnixFSFile
 export const fromFSFile = (fsObj: UnixFSFile): Link => {
   const { name = '', cid, size, mtime, type } = fsObj
   return {
@@ -21,6 +22,7 @@ export const fromFSFile = (fsObj: UnixFSFile): Link => {
   }
 }
 
+// NodeMeta?
 export const fromNodeMap = (nodes: NodeMap): Links => {
   return mapObj(nodes, val => {
     const { name = '', cid, size, mtime, isFile } = val
@@ -28,6 +30,7 @@ export const fromNodeMap = (nodes: NodeMap): Links => {
   })
 }
 
+// Why not `new Link(...)`?
 export const make = (name: string, cid: string, isFile: boolean, size?: number): Link => {
   return {
     name,
@@ -38,6 +41,7 @@ export const make = (name: string, cid: string, isFile: boolean, size?: number):
   }
 }
 
+// Link.join([...])?
 export const arrToMap = (arr: Link[]): Links => {
   return arr.reduce((acc, cur) => {
     acc[cur.name] = cur
--- a/src/fs/path.ts
+++ b/src/fs/path.ts
@@ -1,7 +1,8 @@
 import { NonEmptyPath } from './types'
 
+// segments
 export const splitParts = (path: string): string[] => {
-  return path.split('/').filter(p => p.length > 0)
+  return path.split('/').filter(p => p.length > 0) // why not `p == ''`
 }
 
 export const join = (parts: string[]): string => {
@@ -25,7 +26,7 @@ export const takeHead = (path: string): HeadParts => {
 
 export const splitNonEmpty = (path: string): NonEmptyPath | null => {
   const parts = splitParts(path)
-  if (parts.length < 1) {
+  if (parts.length < 1) { // === 0?
     return null
   }
   return parts as NonEmptyPath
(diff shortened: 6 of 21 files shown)


src/common/util.ts
```diff
@@ -50,6 +50,7 @@ export const arrContains = <T>(arr: T[], val: T): boolean => {
 }
 
 export const asyncWaterfall = async <T>(val: T, operations: ((val: T) => Promise<T>)[]): Promise<T> => {
+  // Why not reduce?
```
expede, line 53: @dholms why the manual iteration here?

src/keystore/config.ts
```diff
@@ -4,6 +4,7 @@ import { KeyStore, CryptoSystem } from 'keystore-idb/types'
 const KEYSTORE_CFG = { type: CryptoSystem.RSA }
 
 
+// Is the keystore a singleton?
```
expede, line 7: This pattern is in a few places. Are we trying to explicitly limit the number of keystores?

src/common/util.ts
dholms: the alternative is an `await/async` reduce. Totally possible, but I thought that this was clearer & also apparently more performant (https://stackoverflow.com/a/41243567)

dholms: it'd have to be `await/async` reduce which is possible, but I thought that this was clearer (and is apparently more performant: https://stackoverflow.com/a/49198393)

src/fs/bare/file.ts
```diff
@@ -6,12 +6,12 @@ import BaseFile from '../base/file'
 export class BareFile extends BaseFile {
 
   static create(content: FileContent): BareFile {
-    return new BareFile(content)
+    return new BareFile(content) // why not `super`?
```
dholms: how do you mean?

src/fs/bare/tree.ts
```diff
@@ -17,8 +17,9 @@ class BareTree extends BaseTree {
     this.links = links
   }
 
+  // why is this async? consistent interface? if so, maybe be explicit what we're implementing
   static async empty(): Promise<BareTree> {
-    return new BareTree({})
+    return new BareTree({}) // Why not just part of the constructor as a defautl value?
```
dholms, line 22: Just for the consistent static interface between trees

src/fs/filesystem.ts
```diff
@@ -157,6 +161,7 @@ export class FileSystem {
   }
 
   async sync(): Promise<CID> {
+    // waterfall? Why not in parallel?
     this.root = await asyncWaterfall(this.root, [
```
dholms, line 165: `addChild` returns an the updated tree. I don't think that we'd want to run these exact operations in parallel because they could get out of sync.
But I agree about the concern of a bottleneck. Instead we could either:
- `put` each child, and then manually update the links in the `Tree` all at once
- add a tree method for adding multiple children in parallel (which would essentially do the above^ but in the `Tree` implementation instead of here)

src/common/util.ts
expede: > this was clearer

Interesting. I find `reduce` to be much more descriptive of what's happening, since it's a more structured abstraction. `for` loops admit all kinds of control flow trickery not allowed in a `reduce`. To each their own 🤷‍♀️

> apparently more performant

I did some further digging. The performance improvement is from not needing to call `await` on the accumulator (because it's an argument marked by `async`). This should be negligible because the network far outstrips the `await` resolution. Another fascinating edge case in the JS semantics!

src/fs/filesystem.ts
```diff
@@ -172,6 +177,7 @@ export class FileSystem {
     return cid
   }
 
+  // Legit don't know what this means: synchronize vs synchronous
```
dholms, line 180: synchronize

yeah I didn't realize the naming overlap here, I'm happy to change that up

src/fs/filesystem.ts
expede: > because they could get out of sync

How? Why would there be dependencies between these?

src/fs/filesystem.ts
```diff
@@ -191,6 +197,7 @@ export class FileSystem {
     const head = parts[0]
     const relPath = pathUtil.join(parts.slice(1))
 
+    // Why `a` not FileSystem? TS has subtyping. I easily could be missing something.
```
dholms, line 200: Methods return Trees, Files, CIDs, FileContent, etc.  This generic is just "whatever the method returns"

Note: I'm not very happy with this function, and it's getting more complex as the code grows. I'm definitely looking for ways to improve it

src/fs/link.ts
```diff
@@ -21,13 +22,15 @@ export const fromFSFile = (fsObj: UnixFSFile): Link => {
...
 
+// NodeMeta?
 export const fromNodeMap = (nodes: NodeMap): Links => {
   return mapObj(nodes, val => {
     const { name = '', cid, size, mtime, isFile } = val
     return { name, cid, size, mtime, isFile }
   })
 }
 
+// Why not `new Link(...)`?
```
dholms, line 33: I like it :+1:

src/fs/bare/file.ts
expede: > Outdated

Sorry, I had copy/pasted, and then changed it to

> return new BareFile(content) // why not the constructor?

src/fs/path.ts
```diff
@@ -1,7 +1,8 @@
 import { NonEmptyPath } from './types'
 
+// segments
 export const splitParts = (path: string): string[] => {
-  return path.split('/').filter(p => p.length > 0)
+  return path.split('/').filter(p => p.length > 0) // why not `p == ''`
```
dholms, line 5: either is fine!

src/fs/bare/tree.ts
expede: Looking at the other instances, we may be able to fold all of these into a constructor with named (object) arguments. Need to explore a bit more first, but it's in the running.

src/fs/filesystem.ts
expede: Okay cool cool! Thanks 😄

src/keystore/config.ts
dholms: You only need one keystore for the sdk, and this lets you access that keystore from anywhere in the code without passing it around as a parameter to everything.

I don't like singletons very much (even tho I use them twice in the sdk :sweat_smile:), but it seemed like a good fit here. Open to alternatives!

src/common/util.ts
dholms: Yeah I generally agree on `reduce` being more descriptive & like it better. But with await/async, the accumulator being a promise & this for loop being super self contained, this just seemed like the less headachey way to do it

src/fs/v1/PublicTree.ts
```diff
@@ -28,6 +33,7 @@ export class PublicTree extends BaseTree implements HeaderTree {
     }
   }
 
+  // Wait wait... why async?
   static async empty (parentKey: Maybe<string>): Promise<HeaderTree> {
```
expede, line 37: @dholms I'm unclear why an empty tree would need to be a promise. I realize that different classes will have different representations, but in all cases we would know the CID in advance, and it should always return a consistent structure, no?

dholms: The `PrivateTree.empty` static method is async (because generating a new symm key through web crypto is async). So I made `PublicTree.empty` async as well so that the interface would be the same

src/fs/filesystem.ts
dholms: Let's say you add 2 files: `deeply/nested/file1` and `deeply/nested/file2` at the same time

You tell the root node to add both files, for each it recurses down to `deeply/nested` & adds the file, then recursively updates the nodes on the way back. You're going to end up splitting the tree because `file1` & `file2` aren't getting added to the same `nested` object, they're being added to 2 separate objects that both represent `nested`. 

You either need to update `nested` for both `file1` and `file2` in the same operation. Or you need to make sure that the operation that is adding `file2` to `nested` is aware that `nested` has been updated to include `file1`.

This could be fixed in the current implementation by adding an `addChildren` method that takes multiple children for adding children in parallel. Or by making `Tree`s more stateful so that each operation will be aware of the other one happening, altho we might run into some weird race conditions with something like this...

I think that the plan to serialize the FS to a list before adding it to the network work going to make this a lot simpler

expede: > at the same time

How would you even do that? Don't you need to call some kind of "add" on it twice?

I'm still unclear why you'd need this to be stateful at all. Each subtree is independent, even in your description above. You recurse through them recursively with a base case. They then build a new nested object. We should be able to blow away the entire structure and build a new one fresh if we need to, and get the same result, no?

expede: > I think that the plan to serialize the FS to a list before adding it to the network work going to make this a lot simpler

I'm very heavily leaning to not using a list for the public side, only the private one

src/fs/v1/PublicTree.ts
expede: Ooooh I see. Gotcha thanks 👍 That's annoying 😛

expede: From a quick search — it sounds like the JS convention for this is `init`. I like `empty` because of the monoidal implications, but it _is_ running a side effect in the browser. Anyhow, naming is hard, right?

src/fs/filesystem.ts
expede: > it recurses down to deeply/nested & adds the file

Actually, is this happening with structural recursion, or message passing? 👀

src/fs/filesystem.ts
```diff
@@ -165,13 +171,16 @@ export class FileSystem {
 
     const cid = await this.root.put()
 
+    // What are we using these hooks for?
     this.syncHooks.forEach(hook => {
```
expede, line 175: @dholms I don't actually see these used anywhere, aside from adding and removing them. Do we need this code?

src/fs/v1/PublicTree.ts
dholms: yup :stuck_out_tongue: 
I chose `empty` because it was making it clear that the file didn't have any children (in opposition to `fromCID` or `fromHeader` which is pulling some info from another source so it might have children). But I'm fine with `init` as well

src/fs/filesystem.ts
dholms: Yup for users to hook into. Steven uses sync hooks in drive

expede: Interesting. Any idea for what?

expede: Like, not sure if Drive should care if the files are coming off the network, right? Maybe something we can later clean up with a `Session` object?

src/fs/filesystem.ts
dholms: > How would you even do that?
by "at the same time", I meant "in parallel".

> I'm still unclear why you'd need this to be stateful at all. Each subtree is independent, even in your description above. 

The folder `/deeply/nested` is being touched by two separate operations though. So if they're running in parallel, they can't be treated independently since they're both altering the same directory

> You recurse through them recursively with a base case. They then build a new nested object. We should be able to blow away the entire structure and build a new one fresh if we need to, and get the same result, no?
The two `add` operations have to be aware of one another tho (for anything that is nested)
Either:
- `file1` and `file2` need to be added to node at the same time (this can be done sequentially in memory, but the `add` to the network would include both changes at the same time)
- or whatever function is adding `file2` to `deeply/nested` must be aware that `file1` was previously added to it.
Otherwise the tree will split and one function will think that `deeply/nested` is a dir that contains one file: `file1` and the other function will think that its a dir that contains one file: `file2`

Am I understanding your point correctly? It seems like what we're each saying isn't quite clicking :sweat_smile:

dholms: > I'm very heavily leaning to not using a list for the public side, only the private one. We can almost certainly get that simplicity by making the structure less stateful / more independent
ok cool :+1: I'm interested to hear why the change in the thought, cause last I remember, we were leaning towards keeping the implementations the same. But maybe this would be better to do over a different medium (Talk/Zoom)

dholms: > Actually, is this happening with structural recursion, or message passing?
ummm not quite sure the comparison here... I'm pretty sure it's using both, but I think I'm probably misunderstanding the question :sweat_smile:

expede: > The folder /deeply/nested is being touched by two separate operations though. So if they're running in parallel, they can't be treated independently since they're both altering the same directory

Maybe this will clarify my thinking / possibly an incorrect assumption? 👇

1. You can't touch two nodes at once, since the memory will get lock by the runtime. The main thread is also single threaded.
2. Why this would be a problem either with immutable structural recursion or mutation. In either case, the parent should get updated. It's not possible for them to happen at the same time, and the operations commute so you should get the same result running `f . g` or `g . f`. Given this property, you can run these operations concurrently, and get the same result.

expede:
NOTE TO SELF: base64URL libarry

src/fs/filesystem.ts
icidasset: Here's the sync hook that Drive uses:
```js
async function syncHook(cid) {
  // stores the root cid locally, used as a cache, among other things.
  // a time difference currently determines when to use cache, or when to use sdk.dataRoot("username")
  // (see below)
  app.ports.ipfsReplaceResolvedAddress.send({ cid })

  // lastFsOperation determines when to use cache or when to use sdk.dataRoot("username")
  // → used on app start
  localStorage.setItem("fissionDrive.lastFsOperation", Date.now().toString())

  // update the user's data root
  await sdk.updateDataRoot(cid, {
    apiEndpoint: endpoints.api
  })
}
```

So yeah, used for "caching" local-first strategy and setting the user's data root. I also figured some users may wanted to hook into this, in case they wanted to do other stuff with the root CID of the file system.

> Like, not sure if Drive should care if the files are coming off the network, right?

It shouldn't care no. Not sure what it has to do with this though, can you clarify?

expede: > So yeah, used for "caching" local-first strategy and setting the user's data root. 

I'm very confused then. Doesn't the user need to call `Fission.sync()` (or something)? They'll get a promise, and know when it's resolved, along with a bunch of inspect-able data. What am I missing?

> It shouldn't care no. Not sure what it has to do with this though, can you clarify?

I'm clearly still confused about the use case. My understanding is that there are hooks for when the CID gets pushed or pulled from the remote. Not sure why the user would need to hook into that,  it feels very leaky, and my intuition is telling me that there's a cleaner design. But again, I clearly don't understand the use case fully.

icidasset: > Doesn't the user need to call Fission.sync() (or something)?

No, the sync hooks are called automatically when something changes in the file system.
For example, you do `fs.add()` and then after that it'll call the sync hook for you.

> My understanding is that there are hooks for when the CID gets pushed or pulled from the remote.

Like explained above, that's not what sync hooks are for. As shown in my Drive sync-hook example, the user is currently in total control over the data-root updating (that's what you meant by "push-pull from the remote", right?)

src/keystore/config.ts
expede: I definitely get that this is convenient. Purely for exploration, would there be a problem with something like:

```ts
const {keystore, filesystem} = Fission.setup({...})
```

Where the filesystem also has a link to that keystore on setup. This is more flexible (say you want to load multiple filesystems, or keep two applications totally separate). But maybe the tradeoff for convenience is too high? Not sure, very much looking for your thoughts here.

src/fs/filesystem.ts
expede: > For example, you do fs.add() and then after that it'll call the sync hook for you.

This still feels very heavy and leaky. Lets do a call 😅

expede: From the call: 

* There probably aren't many other use cases for these hooks. Let's bake this in.
  * Debouncing and whatnot is likely going to get a lot easier when we have versioning
    * Because then we'll have Merkle proofs for which is the latest version
  * Default `add` probably should hit the server (so users don't forget)
    * Need to then either make a `addLocal` method, or add options `add(foo, {remote: false})`
      * My gut is telling me the config object
      * We're going to have ephemeral P2P updates and stuff, so _probably_ easier that way
      * Further thinking required, though
* Configuration (e.g. HTTP endpoint) is necessary
  * Either a Config object, or part of the `Fission.Session`

src/fs/path.ts
expede: > Specifically, the > and < and >= and <= operators all operate either on two numeric values, or two strings, preferring numeric values. That is, if one value is a number, then the other is treated as a number. If a non-number can't be cleanly converted to a number (that is, if it ends up as NaN), then the result of the comparison is undefined. (That's a little problematic, because undefined will look like false in the context of an if statement.)

It's an edge case, but `===` is probably preferred.

```
'4' > 0
true
```

TS will stop this, but downstream people can call the SDK untyped

expede:
Performance: as an experiment, I created 500k simple (but unique) objects on my laptop, and added them to an array via a loop. It took under a second.

dholms:
> Performance: as an experiment, I created 500k simple (but unique) objects on my laptop, and added them to an array via a loop. It took under a second.
to a local daemon or an in-browser daemon?

src/keystore/config.ts
dholms: :thinking: :thinking: Looking back at the ts-sdk, we really don't even use that singleton very much (especially in the filesystem, which is all symm keys). 

In that case would we still have the singleton? one-off methods such as `core.did()` need access to a canonical keystore for determining the user's public key

src/fs/path.ts
dholms: ahh gotta love JS...

ok yup that makes sense :+1:

PR #61, opened 2020-07-15
Add mv method to FileSystem

dholms:
## Problem
No way to move files around within filesystem

## Solution
Add `mv` method to FileSystem

Note that the unix `mv` command actually has several implementations where the second parameter can either represent a destination folder or the name of the actual destination file.

I only included the second implementation here for simplicity's sake.

so for instance if you want to move `private/photos/headshot.jpg` to `public.photos/headshot.jpg`, you wil do
`fs.mv('private/photos/headshot.jpg', public/photos/headshot.jpg` 
_NOT_
 `fs.mv('private/photos/headshot.jpg', public/photos')`

### DIFF
--- a/src/fs/filesystem.ts
+++ b/src/fs/filesystem.ts
@@ -157,7 +157,8 @@ export class FileSystem {
     if (check.isFile(destination)) {
       throw new Error(`Can not \`mv\` to a file: ${destPath}`)
     }
-    return this.addChild(to, node)
+    await this.addChild(to, node)
+    return this.rm(from)
   }
 
   async addChild(path: string, toAdd: Tree | FileContent): Promise<CID> {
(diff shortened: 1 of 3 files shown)


src/fs/filesystem.ts
```diff
@@ -145,6 +146,28 @@ export class FileSystem {
     })
   }
 
+  async mv(from: string, to: string): Promise<CID> {
+    const node = await this.get(from)
+    if (node === null) {
+      throw new Error(`Path does not exist: ${from}`)
+    } 
+    const toParts = pathUtil.splitParts(to)
+    const destPath = pathUtil.join(toParts.slice(0, toParts.length - 2)) // remove file/dir name
```
icidasset, line 155: I think this should be `toParts.length - 1`, because the `end` (aka. 2nd param) is non-inclusive.
See docs: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/slice#Parameters

dholms: yup you're totally right! I actually just stumbled on that myself as I was testing :+1:

PR #68, opened 2020-07-20
Automatically update user's data root

### LINKED ISSUE #66: Remove syncHook and move dataRoot updating to a submodule
A user of the SDK shouldn't have to know about the "data root".

To do:
- [x] The SDK should automatically update the data root after a file-system change
- [x] The API endpoint should be configurable
- [x] This data-root-updating call to the API should be debounced/throttled
- [x] Should not be called when offline
- [x] Should be called automatically when the user comes back online (after changes were made offline)

icidasset:
Removes the `syncHooks` interface and automatically updates the user's data root when making changes to the file system. The data-root updating call is throttled (only make the api call every 5 seconds) and will be called automatically when the user comes back online (and made changes while offline).

Also adds `setup` module for easier customisation of endpoints.
See the [README](https://github.com/fission-suite/ts-sdk/blob/no-synchooks/README.md#fission-sdk) for more info.

Closes #66


src/setup/internal.ts
```diff
@@ -1,6 +1,16 @@
...
-  ipfs: {}
+export type Endpoints = {
+  api: string
+  lobby: string
+  user: string
 }
 
 
-export default setup
+export const setup = {
```
dholms, line 8: should this be `let` since it's mutable?

src/setup.ts
```diff
@@ -1,6 +1,51 @@
-import internalSetup from './setup/internal'
+import { Endpoints, setup as internalSetup } from './setup/internal'
 
 
-export function ipfs(s: { [key: string]: unknown }): void {
-  internalSetup.ipfs = s
+type SomeEndpoints = {
```
dholms, line 4: typescript has a nifty type modifier `Partial` where you can do `Partial<Endpoints>` and not have to define a new type:
https://www.typescriptlang.org/docs/handbook/utility-types.html#partialt

src/fs/filesystem.ts
```diff
@@ -25,16 +27,37 @@ export class FileSystem {
...
     this.root = root
     this.publicTree = publicTree
     this.prettyTree = prettyTree
     this.privateTree = privateTree
+
     this.syncHooks = []
+    this.syncWhenOnline = null
+
+    // Update the user's data root when making changes
+    const syncHook = throttle(5000, cid => {
```
dholms, line 45: dig the throttle :+1: 
but could we wind up in a situation where a user fires off a number of updates in quick succession, and only the first gets synced?

src/fs/filesystem.ts
```diff
@@ -25,16 +27,37 @@ export class FileSystem {
...
+      if (window.navigator.onLine) return updateDataRoot(cid)
+      this.syncWhenOnline = cid
+    })
+
+    this.syncHooks.push(syncHook)
+
+    // Sync when coming back online
+    window.addEventListener('online', () => this.whenOnline())
```
dholms, line 53: didn't know this was a browser event. cool :+1:

src/setup/internal.ts
icidasset: I think `const` is correct, since we don't want the top reference to change (which `let` would allow). So, in other words, `const` allows `setup.key = "change"`, but doesn't allow `setup = otherObject`. Which is what we want I think?

src/setup.ts
icidasset: Oh nice, will change 👍

src/fs/filesystem.ts
icidasset: I don't believe so, it should fire the first call and the last call, and there will always be 5 seconds in between (so last call will be delayed when necessary). See readme for more info: https://www.npmjs.com/package/throttle-debounce#throttle

src/setup/internal.ts
dholms: ah yup you're right :+1:

dholms: misread the setup methods & thought you were updating the top reference

src/fs/filesystem.ts
dholms: ah got it got it, that's what the `noTrailing` param that it mentions would do. cool this looks fine then :+1:

PR #69, opened 2020-07-22
WNFS rewrite

dholms:
## Problem
The format of the public side of the filesystem does not match what's in the whitepaper

## Solution
- cbor encode metadata as a single file
- get size from IPLD not as a metadata value
- remove pin lists
- replace `fileIndex` (which stores all the header info of every child) wtih `skeleton` which only stores CIDs and `children` which stores metadata of immediate children


dholms:
pre-rebase: c440881cc6ae757ca64ae475462a71f3f39b55b9

src/fs/filesystem.ts
```diff
@@ -128,34 +127,41 @@ export class FileSystem {
...
+    prettyTree.onUpdate = result => fs.updateRootLink(Branch.Pretty, result)
+    privateTree.onUpdate = result => fs.updateRootLink(Branch.Private, result)
+
+    return fs
   }
 
+  static async forUser(username: string, opts: FileSystemOptions = {}): Promise<FileSystem | null> {
+    const cid = await dataRoot.lookup(username)
+    return cid ? FileSystem.fromCID(cid, opts) : null
+  }
```
icidasset, line 164: I think this sneaked back in with the rebase?
We have `src/loadFileSystem.ts` now that replaces this.

src/fs/filesystem.ts
```diff
@@ -98,27 +111,39 @@ export class FileSystem {
...
+
+    const root = await BareTree.empty()
 
-    await root.addChild('public', publicTree)
-    await root.addChild('pretty', prettyTree)
-    await root.addChild('private', privateTree)
-    await root.addChild('pins', pinsTree)
+    // await root.addChild('public', publicTree)
+    // await root.addChild('pretty', prettyTree)
+    // await root.addChild('private', privateTree)
```
expede, line 125: Still needed?

src/fs/filesystem.ts
```diff
@@ -172,94 +208,83 @@ export class FileSystem {
...
+  async updateRootLink(branch: Branch, update: AddResult): Promise<void> {
+    this.root.updateLink(link.make(branch, update.cid, false, update.size))
+  }
+
 
   // POSIX INTERFACE
   // ---------------
 
-  async add(path: string, content: FileContent): Promise<CID> {
+  async mkdir(path: string): Promise<CID> {
```
expede, line 219: Actually about line 216 — anything that you can make a logic section about (like “POSIX interface”) can be a TS `interface`

src/fs/protocol/private/mmpt.ts
```diff
@@ -0,0 +1,100 @@
+import { AddResult, CID } from "../../../ipfs"
+import * as basic from '../basic'
+import * as link from '../../link'
+import { Links } from '../../types'
+
+const nibbles = { "0": true, "1": true, "2": true, "3": true, "4": true, "5": true, "6": true, "7": true,
+                  "8": true, "9": true, "a": true, "b": true, "c": true, "d": true, "e": true, "f": true,
+                } as {[key: string]: boolean}
```
expede, line 8: Wait, what is this for? Why all true?

expede: Oh testing membership? Why not array membership? Speed?

src/fs/filesystem.ts
dholms: oops thanks for pointing that out! :ok_hand:

src/fs/filesystem.ts
dholms: nope, thanks! forgot to cleanup this file :sweat_smile:

src/fs/protocol/private/index.ts
```diff
@@ -0,0 +1,79 @@
...
+  cid: CID
+  name: PrivateName
+}
+
+export const findLatestRevision = async (mmpt: MMPT, bareName: BareNameFilter, key: string, lastKnownRevision: number): Promise<Maybe<Revision>> => {
+  let lowerBound = lastKnownRevision, upperBound = null
+  let i = 1
+  let lastRevision: Maybe<Revision> = null
+  while(upperBound === null){
+    const toCheck = lastKnownRevision + Math.pow(i, 2)
```
expede, line 50: I think this exponent is actually backwards. We want powers of two, not squares. The order is confusing, but from MDN:

```js
console.log(Math.pow(4, 0.5));
// expected output: 2
```

src/fs/protocol/private/mmpt.ts
dholms: yup for testing membership. I just figured this was faster than looping an array :man_shrugging:

src/fs/protocol/private/namefilter.ts
```diff
@@ -0,0 +1,131 @@
...
+  }
+  return true
+}
+
+// counts the number of 1s in a uint32
+// from: https://graphics.stanford.edu/~seander/bithacks.html#CountBitsSetParallel
+const bitCount32 = (num: number): number => {
+  num = num - ((num >> 1) & 0x55555555)
+  num = (num & 0x33333333) + ((num >> 2) & 0x33333333)
+  return ((num + (num >> 4) & 0xF0F0F0F) * 0x1010101) >> 24
```
expede, line 130: Minor, but any thoughts on not reassigning the variable? There are sometimes subtle bugs when there’s lots of reassignment.

expede: Also `const`s generally get optimized away, so the performance impact would be negligible even if we weren’t operating on a handful of bytes.

expede: ASIDE: this method only works on 32-bit integers, and the complexity goes up massively from there (chunking, twos compliments, rolling count windows, and so on). If we ever need to do larger numbers, for our current use case, we can either chunk or use one of the slightly slower (there is no way that this is the bottleneck) but are more general.

src/fs/protocol/private/index.ts
dholms: ah yup that's totally right :+1:

PR #70, opened 2020-07-22
Improve file system interface

icidasset:
_As discussed here: https://talk.fission.codes/t/file-system-sdk-story/790_

**The purpose of this PR is to improve the dev experience of the SDK.**

Changes:
- Improves README, should show everything you need, to do regular tasks
- `sdk.isAuthenticated` is now `sdk.initialise`
- `sdk.initialise` now returns an instance of the file system (unless disabled, see readme)
- Adds `sdk.loadFileSystem` (called from initialise ☝️ ). This function is responsible for caching the file system locally and making a file system if the user hasn't got one yet.
- Adds the `fs.appPath.private`, and public, functions (this is a stopgap until we have subgraphs, hard/soft links)
- Adds `fs.exists`
- Adds `fs.read` as an alias for `fs.cat`
- Adds `fs.write` (overwrite a file, create if it doesn't exist yet)
- Adds a "CID log" in which we keep track of file-system root cids (see `loadFileSystem` for more info)

Sample app code:
https://github.com/icidasset/quotes/blob/v2/src/Main.js


src/filesystem.ts
```diff
@@ -0,0 +1,57 @@
...
+
+// ㊙️
+
+
+async function addSampleData(fs: FileSystem): Promise<void> {
+  await fs.mkdir("private/Apps")
+  await fs.mkdir("private/Audio")
+  await fs.mkdir("private/Documents")
+  await fs.mkdir("private/Photos")
+  await fs.mkdir("private/Video")
```
icidasset, line 56: Is there a better way of doing this?

I tried doing:
```
await this.runOnTree("private", true, async (tree, relPath) => {
    await tree.mkdir("Apps")
    await tree.mkdir("Documents")
    await tree.mkdir("Photos")
})
```
But that doesn't seem to work.

README.md
```diff
@@ -233,14 +193,24 @@ Returns a list of links at a given directory path
...
 ```ts
-// public
-const links = await wnfs.ls("public/some/directory/path")
-// private
-const links = await wnfs.ls("private/some/directory/path")
+linksObject = await wnfs.ls("public/some/directory/path") // public
+linksObject = await wnfs.ls("private/some/directory/path") // private
+
+// convert to list
+links = Object.entries(linksObject)
```
icidasset, line 204: Why is this an object? 🤔

src/filesystem.ts
```diff
@@ -0,0 +1,57 @@
...
+  // Determine username
+  username = username || (await authenticatedUsername() || undefined)
+  if (!username) throw new Error("User hasn't authenticated yet")
+
+  const lastLocalChange = await localforage.getItem(FS_TIMESTAMP) as number || 0
+  const currentTime = Date.now()
+
+  // If our last file-system change was over 15 minutes ago
+  if (currentTime - lastLocalChange > 30 * 60 * 1000) {
+    cid = await dataRoot.lookup(username)
```
icidasset, line 27: This is done to avoid data loss, because either the DNS doesn't update fast enough, or the system/browser caches the DNS too long.

This could most likely be done better, so if anyone has any quick fixes for this, let me know.

src/filesystem.ts
```diff
@@ -0,0 +1,57 @@
...
+
+  // Determine username
+  username = username || (await authenticatedUsername() || undefined)
+  if (!username) throw new Error("User hasn't authenticated yet")
+
+  const lastLocalChange = await localforage.getItem(FS_TIMESTAMP) as number || 0
+  const currentTime = Date.now()
+
+  // If our last file-system change was over 15 minutes ago
+  if (currentTime - lastLocalChange > 30 * 60 * 1000) {
```
icidasset, line 26: This is done to avoid data loss, because either the DNS doesn't update fast enough, or the system/browser caches the DNS too long.

This could most likely be done better, so if anyone has any quick fixes for this, let me know.

src/fs/filesystem.ts
```diff
@@ -41,8 +57,20 @@ export class FileSystem {
...
+      private(appUuid: string, suffix?: string | Array<string>): string {
+        return appPath("private", appUuid, suffix)
+      }
+    }
+
     // Update the user's data root when making changes
-    const syncHook = throttle(5000, cid => {
+    const syncHook = throttle(5000, async cid => {
+      await localforage.setItem(FS_TIMESTAMP, Date.now())
+      await localforage.setItem(FS_CID, cid.toString())
```
icidasset, line 72: Probably should add this to a different sync hook that's not throttled.

src/filesystem.ts
expede: Yeah, time based caches cause all kinds of headaches. They’re generally considered a no-no in distributed systems (which the internet is).

In the long term, we’ll be able to check if you’re ahead of DNS because the CID will be in the history of your current file system.

In the meantime, why not track the root CIDs that are out of date in an array? i.e. When you load the page, record the FS root CID. Then when you update it, write that one down as well. Keep them in localStorage. When you try to pull from DNS, check that the hash isn’t in your local “out of date” cache. Now you’re not relying on time as a rough proxy, and can know instantly if there’s something new by using the Merkle proof (the method outlined here).

expede: Also, every time you successfully pull and update our local copy (i.e. DNS was ahead of your local), you can clear this cache. It doesn’t need to grow infinitely

icidasset: > when we implement FS versioning
👍 

> why not track the root CIDs that are out of date in an array?
Oh I like that, will use that.

src/filesystem.ts
```diff
@@ -0,0 +1,70 @@
...
+  } else if (logIdx > 0) {
+    // DNS is outdated
+    cidLog.removeOlderCids(logIdx)
+    cid = await cidLog.newest()
+
+  } else {
+    // DNS is newer
+    cid = dataCid
+
+  }
```
icidasset, line 45: Implemented that CID log @expede 
Here you can see all the conditions it checks for.

expede: This feels like a lot of steps. Why not something like:

```js
if (cidLog.includes(dnsRoot)) {
  // Local ahead of remote
  cidLog.push(dnsRoot)
} else {
  // Remote ahead of local
  cidLog = [dnsRoot]
  localRoot = dnsRoot
}
```

expede: Specifically, we shouldn’t care about ordering, so `=== 0` vs `> 0` shouldn’t matter. Because DNS is eventually consistent, you may get out of order reads from it. Set membership is probably safer/easier, unless I’m missing something

expede: Speaking of sets, may as well use a `Set` rather than an array, since we don’t care about order

src/filesystem.ts
expede: Lenses would actually work really well here, since the structure is persistent :P This is part of why it may be nice to make our internal representation more inductive — update some subgraph, and reattach to the rest of the graph.

For now you may need to it as you’ve written, but this kind of thing is in the short-term roadmap 👍

src/filesystem.ts
icidasset: I don't think a `Set` works here. What if your local file system is multiple CIDs ahead and you refresh the page, how do you know which is the last one? You are correct that I could eliminate one branch here and just use `log.includes()`

`setDNS(dnsRoot)` would conflict with other code yeah. Because a DNS update will always be done for every file-system change, it might just not be persisted yet.

> Because DNS is eventually consistent, you may get out of order reads from it.

Which is something that we can take advantage of when have versioning and/or crdt stuff? Not right now I assume?

expede: > What if your local file system is multiple CIDs ahead and you refresh the page, how do you know which is the last one

It shouldn’t matter because all the cache says is “I’ve seen this one before. The one that I have in localStorage is thus up to date.” The cache only has the single responsibility: to check if this is something new. The filesystem can’t be rewound. With an array, we have an implicit assumption that the first one is the current head of the history. Until we support versioning, we don’t care about the order earlier than this.

I think maybe the main difference between our approaches is I’m assuming that the current filesystem root is kept in a different variable than this cache. You _can_ overload them, but then you need to check for the there states.

---

> Which is something that we can take advantage of when we have versioning and/or crdt stuff? Not right now I assume?

It’s never going to be an asset, only a liability. What CRDTs, Merkle history, and whatnot give us is a way of not being dependent on time or order — our data dependencies are only the things that matter to the operation of the program. DNS will sometimes return old results, which is weird for humans to think about because we want a nice linear ordering with a clear “now”, but distributed systems (DNS, CRDTs, GenServer retry) are a bit more “quantum”.

icidasset: > I think maybe the main difference between our approaches is I’m assuming that the current filesystem root is kept in a different variable than this cache.

Oh ok, yes that's indeed the difference. Makes sense now 👍 

> I love talking about this stuff, but writing a small textbook here isn’t the more effective way to communicate the topic in full. Always happy to hop on a call to explain with more hand gestures :P

Yes please 😄 Lets talk about it later during our call. I understand the situation and liabilities, but don't know about the techniques to use to fix this.

PR #88, opened 2020-08-19
Add permission system and enable multiple apps

### LINKED ISSUE #73: Ability to use file system across multiple domains
Depends on fission-suite/fission#348 and the file system refactor.

### LINKED ISSUE #82: Doesn't work offline
Needs to check if the user is online before doing any DNS lookups.

icidasset:
#### Changes

- `initialise` now accepts two additional options, named "prerequisites":
   ```javascript
   const { prerequisites, scenario, state } = await wn.initialise({
     // Will ask the user permission to store
     // your apps data in `private/Apps/Nullsoft/Winamp`
     app: {
       name: "Winamp",
       creator: "Nullsoft"
     },

     // Ask the user permission for additional filesystem paths
     fs: {
       privatePaths: [ "Music" ],
       publicPaths: [ "Mixtapes" ]
     }
   })
   ```
- Those prerequisites are passed to the `wn.redirectToLobby` function.  
  (So the auth lobby has the correct parameters to determine the permissions to ask the user)
- Adds the ability to use multiple apps with one file system (closes #73)
- The SDK now handles multiple UCANs
- Works offline (fixes #82)
- Added `initialize` (american spelling) as an alias for `initialise`
- Adds ability to set the potency `ptc` of a UCAN

#### Breaking changes

- File system actions (ie. POSIX interface methods) throw an error when they are missing the necessary permissions (read UCAN)
- Does no longer automatically call the `fs.publicise()` method. You have to call this yourself from now on.
- Changed the `fs.appPath` function (see README)
- Renamed `scenario.isAuthenticated` to `scenario.isAuthorised`
- Renamed `scenario.continuum` to `scenario.continuation`
- The first parameter to `redirectToLobby` has now become the second parameter.
- Replaced `deauthenticate` with `leave`, which now redirects to the auth lobby automatically, so you can "sign out" there as well.


README.md
```diff
@@ -28,7 +28,7 @@ The Fission SDK offers tools for:
 import * as sdk from 'webnative'
 
 // Browser/UMD build
-self.webnative
+const sdk = self.webnative
```
dholms, line 31: Should we call this `webnative` instead of `sdk`? I use sdk internally cause of brevity, but it would probably be good to have people thinking of this as "webnative" not "the webnative sdk"

README.md
```diff
@@ -102,7 +135,30 @@ WNFS exposes a familiar POSIX-style interface:
 - `mv`: move a file or directory
 - `read`: alias for `cat`
 - `rm`: remove a file or directory
-- `write`: write to a file
+- `write`: alias for `add`
+
+
+## Publicise
+
+The `publicise` function synchronises your file system with the Fission API and IPFS. We don't do this automatically because if you add a large set of data, you only want to do this after everything is added. Otherwise it would be too slow and we would have too many network requests to the API.
```
dholms, line 143: think I used american spelling here :thinking: feel free to switch it/alias on rebase

src/apps.ts
```diff
@@ -14,15 +14,15 @@ export type App = {
...
  */
 export async function index(): Promise<Array<App>> {
   const apiEndpoint = setup.endpoints.api
 
   const jwt = await ucan.build({
     audience: await api.did(),
-    issuer: await did.local(),
-    proof: await localforage.getItem(UCAN_STORAGE_KEY)
+    issuer: await did.ucan(),
+    proof: undefined, // TODO
```
dholms, line 24: still todo?

README.md
dholms: ah you already did that. nvm!

src/filesystem.ts
```diff
@@ -50,12 +60,13 @@ export async function loadFileSystem(username?: string): Promise<FileSystem> {
...
-  fs = cid ? await FileSystem.fromCID(cid) : null
+  const keyName = READ_KEY_FROM_LOBBY_NAME
+  fs = cid ? await FileSystem.fromCID(cid, { keyName, prerequisites }) : null
   if (fs) return fs
 
   // Otherwise make a new one
-  fs = await FileSystem.empty()
-  await addSampleData(fs)
+  fs = await FileSystem.empty({ keyName, prerequisites })
+  // TODO: await addSampleData(fs)
```
dholms, line 69: do we want this back in?

src/fs/filesystem.ts
```diff
@@ -273,17 +286,34 @@ export class FileSystem implements UnixTree {
...
    */
-  async publicize(): Promise<CID> {
+  async publicise(): Promise<CID> {
     const cid = await this.root.put()
+    const proofs = [...this.proofs]
 
-    this.syncHooks.forEach(hook => hook(cid))
+    this.proofs = []
+
+    proofs.forEach(proof => {
```
dholms, line 302: Wanna make sure I'm following the logic correctly

We're storing 1 proof for each change to the file system between publishing & then sending an "udpate data root" req for each proof but with the same CID, right?

In order to process a request, wouldn't the server need all of the proofs for a given change? Together, the proofs are sufficient to change the root CID, but alone I think that each would be insufficient.

Also do we want to perform some sort of dedup here? For instance if a user has `"*"` permission & makes 20 updates before `publicise`ing, we don't need to send 20 requests

src/ucan/internal.ts
```diff
@@ -0,0 +1,123 @@
...
+
+
+let dictionary: Record<string, Ucan> = {}
+
+
+// FUNCTIONS
+
+
+/**
+ * You didn't see anything 👀
```
dholms, line 19: :laughing:

src/ucan/internal.ts
```diff
@@ -0,0 +1,123 @@
...
+): boolean {
+  const prefix = dictionaryFilesystemPrefix(username)
+
+  // Root access
+  const rootUcan = dictionary["*"]
+  if (rootUcan && !ucan.isExpired(rootUcan)) return true
+
+  // Check prerequisites
+  if (app) {
+    const u = dictionary[`${prefix}private/Apps/${app.creator}/${app.name}`]
```
dholms, line 89: Nodes on the private side need to be addressed by their "namefilter" (something like `boris.fission.name/private/a0a7de4f90a....`) not their path in the filesystem.

I see two difficulties here: 
- you can't know a namefilter without having read access to it. So how do you request access? Request access for paths & then the auth lobby (which does have read access already) will translate that into namefilters for the UCAN composition?
- we need a merkle proof with every change that gets sent along on the private side

Ya know this might be worth having a slightly longer form discussion on the forum. I imagine I'll be helping out on this at least with the merkle proof side

dholms: Actually we'll probably want to store these in the dictionary with the path as the key like you're doing here since the private names don't immediately reveal info about the scope.

It looks like the dictionary is keyed by the `rsc` on the UCAN. In the case of private nodes that `rsc` will be the name filter. 

UCANs have to discuss private nodes using their name filters, but at the sdk level, we may want to discuss private nodes using their path (either "absolute path" from the root of the filesystem, or "relative path" from the app folder)

dholms: Alright gonna post some thoughts in Talk

dholms: https://talk.fission.codes/t/ucans-on-the-private-side/988

src/filesystem.ts
icidasset: Thanks, yeah, forgot this one.

src/apps.ts
icidasset: Yeah, waiting on some API changes.
Need to add some more info here 👀

README.md
icidasset: Yeah, good question. I agree 👍 Just `wn` instead? 🤔

src/fs/filesystem.ts
icidasset: > We're storing 1 proof for each change to the file system between publishing & then sending an "udpate data root" req for each proof but with the same CID, right?
Yes

> In order to process a request, wouldn't the server need all of the proofs for a given change?
I think a single change only needs one proof, no?
That's what I gathered from what Brooke told me.
But I could be totally wrong on this.
I don't entirely understand how the server would validate a filesystem/dnslink UCAN 🤔 

> Also do we want to perform some sort of dedup here?
Yeah good point, will add this 👍

src/ucan/internal.ts
```diff
@@ -0,0 +1,123 @@
...
+
+/**
+ * Look up a UCAN with a file system path.
+ */
+export async function lookupFilesystemUcan(path: string): Promise<Ucan | null> {
+  const pathParts = pathUtil.splitParts(path)
+  const username = await common.authenticatedUsername()
+  const prefix = username ? dictionaryFilesystemPrefix(username) : ""
+
+  if (dictionary["*"]) {
```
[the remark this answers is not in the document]
icidasset: Don't think so, since you can have the `*` resource. Prefix would only be necessary if the resource of a UCAN would be an object.

src/apps.ts
dholms: ah cool cool :ok_hand:

README.md
dholms: I like it :+1:

src/ucan/internal.ts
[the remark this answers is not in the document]
dholms: Okay yeah I think this makes sense since we're only ever interacting with one FS at a time right now :ok_hand: 

(Hopefully soon!) apps will be able to interact with multiple filesystems. If we store all available UCANs in the same dictionary, we'll need to preface them with the username. If we have a different dictionary per FS this will be fine

src/fs/filesystem.ts
dholms: > I think a single change only needs one proof, no?

It depends....
Imagine that I have permission for `private/Music` & `private/Photos`, but nothing else. Locally I add something to each folder & then try to sync the CID. 

I think the code right now is submitting two requests:
- req 1 contains the updated CID & a UCAN that proves access to modify `private/Music`
- req 2 contains the updated CID & a UCAN that proves access to modify `private/Photos`

But for the server, this is a single transaction & it needs proof that you can modify both. So we oughta be sending:
- a single req with the updated CID & a UCAN that proves access to modify both `private/Music` & `private/Photos`

Does that make sense/match your understanding?

src/ucan/internal.ts
icidasset: Oh right, forgot about those name filters.
Are name filters one way? (ie. from path -> name filter)
Or can I get a path from a name filter as well?

dholms: I realize this is a lot to lump on at once + namefitlers weren't even a construct in the code you were working with until I got my PR in yesterday :sweat_smile: 

I've been in the land of namefilters for the past couple of weeks so this is all obviously at the top of my mind right now. More than happy to answer any questions & give a hand!

I think the Merkle proofs are out of the scope of this PR (and will probably fall on me to do :thinking:)

But we probably do want to get Private UCAN's talking in terms of namefilters. Check out Brooke's & my discussion on the talk forum. The client will still pass the auth lobby the canonical paths to private folders. The auth lobby will translate those into namefilters & issue UCANs for them. The sdk will talk about them in terms of their human-readable path, but the actual `rsc` field will be a namefilter

expede: > Are name filters one way? (ie. from path -> name filter)
> Or can I get a path from a name filter as well?

They’re not reversible (well, it’s POSSIBLE but completely impractical). Also, the paths here are the AES keys, not the human readable names

dholms: oops didn't see your response when I posted another reply!

yup as brooke said name filters are made from AES keys so you can path -> namefilter, but you need read access to the FS so it's more like:  `(path, key, filesystem) -> namefilter` (which is why only the auth lobby can calculate this)

to go namefilter -> path, I think you would have to walk the tree from the root until you find the node in question :thinking: so yeah pretty impractical

expede: > to go namefilter -> path, I think you would have to walk the tree from the root until you find the node in question 🤔 so yeah pretty impractical

wrt the above quote, but actually a question for @icidasset: is there a specific use case in mind, or just figuring out the bounds of the system?

src/fs/filesystem.ts
icidasset: > But for the server, this is a single transaction & it needs proof that you can modify both.

I'm not sure I understand why this would be transformed into a single transaction 🤔 
Why not separate? cc @expede

src/ucan/internal.ts
icidasset: > But we probably do want to get Private UCAN's talking in terms of namefilters. Check out Brooke's & my discussion on the talk forum. The client will still pass the auth lobby the canonical paths to private folders. The auth lobby will translate those into namefilters & issue UCANs for them.

Ah makes sense now, thanks. Yeah, let's implement that in a different PR 👍 

> question for @icidasset: is there a specific use case in mind, or just figuring out the bounds of the system?

Was confused about what I needed to do here. But it sounds like I can implement this later in the auth lobby and then adjust the SDK accordingly.

src/fs/filesystem.ts
icidasset: Oh wait I think I understand. It's a single transaction because we do `cid = await this.root.put()` in the sdk and then update the data root with that cid?

icidasset: Ok so, can we do this with separate UCANs @expede ? Or do we need a UCAN with multiple scopes (ie. UCAN v1.1)?

dholms: > It's a single transaction because we do cid = await this.root.put() in the sdk and then update the data root with that cid?

Yup exactly. that `cid` is representative of a Filesystem that contains _both_ updates. So in order to update the user's data root to the new `cid`, you need to prove that you have the permission to do both updates.

> Ok so, can we do this with separate UCANs @expede ? Or do we need a UCAN with multiple scopes (ie. UCAN v1.1)?

Both would work I believe, but my understanding is that UCAN's with multiple scopes are targeted at solving just this issue.  Actually I'm gonna go back & read some of the talk forum posts, I remember a discussion on the tradeoffs here. Reviewing this PR has been really for me in getting UCAN stuff more solid in my brain :+1:

expede: Here’s the spec: https://app.gitbook.com/@runfission/s/whitepaper/access-control/ucan/jwt-authentication

expede: > Ok so, can we do this with separate UCANs @expede ?

In theory yes, but we won’t be supporting it, since it’s a subset of what’s possible with UCAN 0.4

src/ucan/internal.ts
dholms: Yup most of this change is in the auth lobby :+1: 

I think the one change in this code is in `compileDictionary`, the dictionary is being keyed by the UCAN's `rsc`. Right now that's the private path, but it _will be_ the name filter. I think we want it to remain keyed by the private path. But yeah this probably doesn't make much sense to do until we're sending back namefilter UCANs from the auth lobby :thinking:

icidasset: Sounds good, let's do another PR 👍

src/fs/filesystem.ts
icidasset: I'll address the UCAN v0.4 changes in another PR.
The foundation for those changes is now there.

The auth lobby will return a UCAN with the `*` rsc for now, so it works with this SDK version and the current API.

icidasset:
@dholms I've changed it so that the proofs needed for the file system are stored in an object, with the signature as the key. So we have some deduplication there now.

The auth lobby will return a UCAN with the `*` resource for now, so it still works with the current API implementation of UCAN, and also this SDK version.

Also made the other changes 👀 
Published as v0.17.0-alpha-1

src/fs/filesystem.ts
```diff
@@ -273,17 +281,34 @@ export class FileSystem implements UnixTree {
...
    */
-  async publicize(): Promise<CID> {
+  async publicise(): Promise<CID> {
     const cid = await this.root.put()
+    const proofs = Array.from(Object.entries(this.proofs))
+
+    this.proofs = {}
 
-    this.syncHooks.forEach(hook => hook(cid))
+    proofs.forEach(([_, proof]) => {
```
dholms, line 297: Totally good with putting off the UCAN 0.4 changes until next PR :ok_hand:  but I think we should adjust this logic slightly in the meantime. With UCAN 0.4, we're going to collate all the required proofs into 1 UCAN & then send one "Update Data Root" request to the server (not one per proof). Right now we're sending 1 request _per proof_. They'll get approved (because the server doesn't have the correct parsing code yet), but there's no real sense in bombarding it with multiple requests to update the same CID. Let me know if I'm misunderstanding the purpose of calling the synchooks once per proof!

tl;dr all good, but let's just call the sync hooks once per `publicise`

dholms:
Sweet sounds good! Excited to get this merged :tada: :tada: 

Just the one note & I think this is good to go!

src/fs/filesystem.ts
icidasset: > there's no real sense in bombarding it with multiple requests to update the same CID.
Totally, that's why adjusted the auth lobby to only return one UCAN.
So it'll only use that one UCAN here, and thus only make one request.

dholms: ah got it. Auth lobby only sends 1 UCAN & we're deduping this so only sending one request. makes sense!

PR #93, opened 2020-08-24
bugfix: if no private tree when loading WNFS, create new

dholms:
## Problem
When loading WNFS from a CID & a private branch does not exist, it will create a new private MMPT, but will attempt to load the root of the private tree as if it exists, causing an error

## Solution
Create a new MMPT & create a new private tree root

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "webnative",
-  "version": "0.16.1",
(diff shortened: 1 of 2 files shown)


icidasset (approved):
Thanks! I think I just bumped into this 👀

PR #95, opened 2020-08-25
Reconnect browser daemon

dholms:
## Problem
The server hangs when trying to update a user's data root. I believe this is because the server disconnects from a peer & takes a while to re-find it to download the content

## Solution
Concurrently reconnect to server while attempting to update the data root


dholms:
A couple data points:
The POST reqs for data root are taking ~2 - 2.5s now. Certainly could be faster, but also significantly better than 0.5 - 1.5 min, or not going through at all

Some testing I did before hand showed that trying to load some simple text in a gateway was nearly immediate if I had just started the browser node (with the remote node as a bootstrap peer) or reconnected first. But it would take a long time, and often time out, if I did not do so. This makes me think that this is definitely a connectivity issue

icidasset:
Nice 👍  I wonder if we set up our gateway as a "delegate node", if the problem will improve or not 🤔

dholms:
Yeah that's not a bad idea ^^ it will almost certainly help with getting content from other users' filesystems. But it would also probably help keep the nodes connected

dholms:
was running into some strange linting errors, so updated ESLint (which prompted some other fixes as well)

PR #97, opened 2020-08-27
Cidv1

dholms:
## Problem
We're still using CIDv0. CIDv1 are the new standard, no longer experimental & are case-insensitive so can be used in domains

## Solution
Upgrade IPFS CIDs to v1
- [x] files
- [x] dag nodes

### DIFF
--- a/src/ipfs/basic.ts
+++ b/src/ipfs/basic.ts
@@ -7,7 +7,7 @@ import { DAG_NODE_DATA } from './constants'
 
 export const add = async (content: FileContent): Promise<AddResult> => {
   const ipfs = await getIpfs()
-  const result = await ipfs.add(content)
+  const result = await ipfs.add(content, { cidVersion: 1 })
 
   return {
     cid: result.cid.toString(),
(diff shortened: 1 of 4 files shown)


src/ipfs/basic.ts
```diff
@@ -50,9 +51,10 @@ export const dagGet = async (cid: CID): Promise<DAGNode> => {
 
 export const dagPut = async (node: DAGNode): Promise<AddResult> => {
   const ipfs = await getIpfs()
-  // using this format so that we get v0 CIDs. ipfs gateway seems to have issues w/ v1 CIDs
+  // using this format because Gateway doesn't like `dag-cbor` nodes. 
+  // I think this is because UnixFS requires `dag-pb` & the gateway requires UnixFS for directory traversal
   const cidObj = await ipfs.dag.put(node, { format: 'dag-pb', hashAlg: 'sha2-256' })
-  const cid = cidObj.toString()
+  const cid = CIDTool.base32(cidObj.toString())
```
icidasset, line 57: Can we do `cidObj.toString('base32')`?
`js-ipfs` uses this package for cids right? https://www.npmjs.com/package/cids

dholms: oh nice thanks for pointing that out! 
I actually had to do `cidObj.toV1().toString()` but I could remove the `cid-tool` dependency :ok_hand:

icidasset (approved):

Issue #77, opened 2020-07-30
File system content not always available on Fission gateway


icidasset:
When setting the data root, we'll want to wait on the API to tell us the content was successfully fetched from our local js-ipfs node. In other words, the data from our file system lives on our local js-ipfs node and we need to wait until that is transferred to the Fission gateway.

More notes:
- The user cannot close the window before this action is finished, show a dialogue when trying to close the window.
- If this were not solved, you can have an infinite loader on your app and possibly lose your file system

expede:
@icidasset I believe that this should fix it: https://github.com/fission-suite/fission/pull/396

I believe that you’ll get a 200 after the data the data fetch is complete — it’s not fire-and-forget internally.Please let me know if you still have problems

icidasset:
I don't think it's fully working as intended.

Here's how I'm testing this:
- I open https://quotes.fission.app/
- Wait for DNSLink-update message to appear in the console
- Double check if the update-data-root request has succeeded
- Open https://quotes.fission.app/ in another browser
- Check if the CID is the same
- Wait for it to load file system
- 🛑 Infinite spinner

Notes:
If open the same file system in Drive in the same browser, it does load a bit quicker than in other browsers (where it often doesn't load at all).

expede:
> If open the same file system in Drive in the same browser, it does load a bit quicker than in other browsers (where it often doesn't load at all).

This sounds like it’s client side. WebRTC or the WSS connection, maybe?

icidasset:
> If you have that window open in your browser, there should always be a route to the data since it’s live

Yeah exactly, that's what I don't understand.
The js-ipfs instance where the data lives is still live, so why does the other js-ipfs instance not find it? 🤔

expede:
Just double checked the server code — we wait for a response from the gateway, so you shouldn’t see a response from us until it’s working. I’m going to double check that we don’t like... ignore errors or something deeper in the code, but on the surface it looks pretty much okay.

expede:
Yeah, we pass back any errors from the gateway

expede:
> > If you have that window open in your browser, there should always be a route to the data since it’s live
> 
> Yeah exactly, that's what I don't understand.
> The js-ipfs instance where the data lives is still live, so why does the other js-ipfs instance not find it? 🤔

[...]

PR #102, opened 2020-09-03
Improve discoverability of other nodes by using a signaling server

icidasset:

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "webnative",
-  "version": "0.17.1",
+  "version": "0.17.2",
   "description": "Fission Typescript SDK",
   "keywords": [],
   "main": "index.cjs.js",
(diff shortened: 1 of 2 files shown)


dholms (approved):
Sweet!

Just out of curiosity, does this just improve connectivity between browser nodes? Or does it help connectivity to our gateway node as well?

icidasset:
Just between browser nodes. I don't think connectivity to our gateway is an issue, since whenever I check `swarm.peers` this peer id of the gateway is in there. Thoughts?

Also, see https://github.com/fission-suite/webnative/issues/77#issuecomment-686537305 for my thoughts on debugging all of this reliability stuff 👀

dholms:
Cool that's what I thought. Just making sure we were on the same page :ok_hand: 

:eyes: checking out that comment now

PR #105, opened 2020-09-04
Configure for Nix

expede:


expede:
Yup! Boris opened an issue #107

PR #106, opened 2020-09-04
Return size on public dirs

dholms:
## Problem
`ipfs.ls` does not return size for directories. I could've sworn that this worked in the past :thinking: but it doesn't return size in the in-browser daemon or a local go-ipfs node so this seems intentional

## Solution
Combine `ipfs.ls` output with `ipfs.dag.get` output to get the entire object that we're looking for


icidasset (approved):
👍 

Besides this, I guess the only reason we're still using `ls` is for the `mtime` property?
Is `mtime` something we can keep track of ourselves?

icidasset:
Aight sounds good, was just wondering if it'd be possible to avoid `ipfs.ls` for the shared worker thing. But I'll see if I can ask the ipfs team to add support for the `ls` method.

PR #109, opened 2020-09-12
Race google & cloudflare lookups

dholms:
## Problem
Cloudflare can intermittently be slow & unresponsive

## Solution
When doing DNS lookups, race Google & Cloudflare DNS-over-HTTPS and return the first to finish

### DIFF
--- a/src/common/async.ts
+++ b/src/common/async.ts
@@ -1,0 +1,9 @@
+/**
+ * Race an array of promises, returning whichever finishes first
+ */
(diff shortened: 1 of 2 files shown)


bmann:
I wonder if this is CF doing something funky with looking at dnslink?

I need to at least post in their Community Forums about this, since my experience with wrong paths and very slow lookups means something is misconfigured.

icidasset (approved):
👍

PR #114, opened 2020-09-18
Hit fission server for dataroot

dholms:
## Problem
DNS takes a while to propagate, and especially on initial lookups, it can take > 15 min

## Solution
Hit new server route for a user's DataRoot, if that fails, then go to DNS

### DIFF
--- a/src/data-root.ts
+++ b/src/data-root.ts
@@ -38,7 +38,7 @@ export async function lookupOnFisson(
   username: string
 ): Promise<CID | null> {
   try {
-    const resp = await fetch(`${setup.endpoints.api}/user/data/${username}`)
+    const resp = await fetch(`${setup.endpoints.api}/user/data/${username}`, { cache: 'reload' }) // don't use cache
     const cid = await resp.json()
     if (!check.isCID(cid)) {
       throw new Error("Did not receive a CID")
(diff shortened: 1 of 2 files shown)


src/data-root.ts
```diff
@@ -25,6 +29,27 @@ export async function lookup(
...
+ * Get the CID of a user's data root from the Fission server.
+ *
+ * @param username The username of the user that we want to get the data root of.
+ */
+export async function lookupOnFisson(
+  username: string
+): Promise<CID | null> {
+  try {
+    const resp = await fetch(`${setup.endpoints.api}/user/data/${username}`)
+    const cid = await resp.json()
```
icidasset, line 42: Does `resp.json()` work for plain-text responses? Or is `cid` an object?

icidasset (approved):
One question looks good otherwise

src/data-root.ts
dholms: yup works for plain-text responses :+1: 

We could do `resp.text()` but that adds quotes on either side so it'd have to be something like
`(await resp.text()).replace('"', '')`

dholms: I just figured this was cleaner

icidasset: Nice, didn't know that. Carry on 😄

dholms: haha neither did I :smile: found out while coding this^

PR #115, opened 2020-09-18
Retry update-data-root requests when 504 status

### LINKED ISSUE #101: Replace `ipfs.object.stat` with `ipfs.files.stat`
Object API is being (slowly) deprecated.
Need to prefix the path with `/ipfs/` though.
See conversation https://github.com/ipfs/js-ipfs/issues/3252#issuecomment-684954307

### LINKED ISSUE #112: Retry updating the data-root when 504 response from server
The API can return a 504 status when the IPFS gateway is still fetching the data from your in-browser js-ipfs node. We should retry this until it succeeds.

Additionally we should cancel long-running data-root requests, whenever a new request is made.

### LINKED ISSUE #116: Pin directories and files retrieved using `ls` and `cat`
And other querying methods.

icidasset:
Keeps retrying update-data-root requests until a 2xx status is received, when getting a 504 status.
Also cancels the previous update-data-root requests if still running.
Closes #112

Replaces `ipfs.object.stat` with `ipfs.files.stat`
(Closes #101)

Pins everything we query.
(Closes #116)

### DIFF
--- a/src/data-root.ts
+++ b/src/data-root.ts
@@ -73,6 +73,9 @@ export async function update(
 ): Promise<void> {
   const apiEndpoint = setup.endpoints.api
 
+  // Debug
+  debug.log("🚀 Updating your DNSLink:", cid)
+
   // Cancel previous updates
   if (fetchController) fetchController.abort()
   fetchController = new AbortController()
--- a/src/data-root.ts
+++ b/src/data-root.ts
@@ -98,7 +98,7 @@ export async function update(
     },
     retries: 100,
     retryDelay: 5000,
-    retryOn: [ 503, 504 ],
+    retryOn: [ 502, 503, 504 ],
 
   }, {
     method: 'PATCH',
--- a/src/data-root.ts
+++ b/src/data-root.ts
@@ -7,13 +7,9 @@ import * as ipfs from './ipfs'
 import { CID } from './ipfs'
 import { Maybe, api } from './common'
 import { setup } from './setup/internal'
-import makeRetryFetch from 'fetch-retry'
 
 
-// `fetch` with `retries`, `retryDelay`, `retryOn` options
-const fetchWithRetry = makeRetryFetch(fetch)
-
-// Only one data-root update at a time
+// Controller for data-root-update fetches
 let fetchController: Maybe<AbortController> = null
 
 /**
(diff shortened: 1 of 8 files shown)


src/data-root.ts
```diff
@@ -4,9 +4,17 @@ import * as did from './did'
...
 import { CID } from './ipfs'
 import { setup } from './setup/internal'
+import makeRetryFetch from 'fetch-retry'
+
+
+// `fetch` with `retries`, `retryDelay`, `retryOn` options
+const fetchWithRetry = makeRetryFetch(fetch)
+
+// Only one data-root update at a time
+let fetchController: Maybe<AbortController> = null
```
dholms, line 17: Cool didn't know about AbortControllers :+1:

src/data-root.ts
```diff
@@ -52,11 +67,12 @@ export async function update(
...
     ipfs.reconnect(),
-    fetch(`${apiEndpoint}/user/data/${cid}`, {
+    fetchWithRetry(`${apiEndpoint}/user/data/${cid}`, {
       method: 'PATCH',
-      headers: {
-        'authorization': `Bearer ${jwt}`
-      }
+      headers: { 'authorization': `Bearer ${jwt}` },
+      retries: 100,
+      retryOn: [ 503, 504 ],
```
dholms, line 74: 503s as well?

src/ipfs/basic.ts
```diff
@@ -18,6 +18,7 @@ export const add = async (content: FileContent): Promise<AddResult> => {
 export const catRaw = async (cid: CID): Promise<Buffer[]> => {
   const ipfs = await getIpfs()
   const chunks = []
+  await ipfs.pin.add(cid)
```
dholms, line 21: :+1:

src/ucan.ts
```diff
@@ -66,7 +66,7 @@ export const WNFS_PREFIX = "floofs"
 export async function build({
   audience,
   issuer,
-  lifetimeInSeconds = 30,
+  lifetimeInSeconds = 900,
```
dholms, line 69: Don't we want UCANs to be short lasting? Like basically a new one per request to prevent replay attacks

icidasset: Yeah true. I'll have to find another way than using `fetch-retry` 🤔 Will revise 👀

src/data-root.ts
icidasset: Yeah, I imagine we could get a 503 "Service Unavailable" if IPFS is currently unavailable (eg. ipfs server is restarting). Is that a thing?

dholms: Ah perhaps (honestly not sure how our server deals with it). Seems good either way tho :ok_hand:

src/data-root.ts
```diff
@@ -37,26 +41,69 @@ export async function update(
...
+
+        // TODO: Waiting on API change.
+        //       Should be `username.fission.name/*`
+        resource: ucan.decode(proof).payload.rsc
+      })
+
+      return { 'authorization': `Bearer ${jwt}` }
+    },
+    retries: 100,
+    retryDelay: 5000,
```
dholms, line 68: why a delay? wouldn't we want to retry immediately?

dholms (approved):
I'm confused about the 5s delay, but besides that, this looks good :+1: :tada:

src/data-root.ts
icidasset: I thought a delay couldn't hurt, to give the server a bit of breathing room, since the requests are already lasting so long 🤷 A request could easily last 5 minutes, so waiting another 5 seconds doesn't change much I guess 😄

dholms: ah kk :+1:

src/ipfs/basic.ts
```diff
@@ -37,6 +38,7 @@ export const cat = async (cid: CID): Promise<string> => {
 export const ls = async (cid: CID): Promise<UnixFSFile[]> => {
   const ipfs = await getIpfs()
   const links = []
+  await ipfs.pin.add(cid)
```
icidasset, line 41: Hmm, I think I'll remove this one.
Every CID we `ls` we also `dagGet`.

Issue #107, opened 2020-09-04
Move docs to separate branch


bmann:
We should move generated docs to a separate branch. Open to other solutions :)

expede:
Yep — not wrong

The nice thing with the current setup is that the docs get generated on _every_ build. We could also gitignore the docs, and deploy them with a GH Action

bmann:
@expede yeah that was the other option I was thinking.

PR #122, opened 2020-09-28
Ignore Docs

dholms:
## Problem
Docs get checked into git every time we build & publish

## Solution
Remove docs from repo & gitignore the folder. These will still get build & published any time we publish to npm!

Closes https://github.com/fission-suite/webnative/issues/107


icidasset (approved):
👍 👍 👍

I don't really know how much people need these docs, besides what we have in the fission guide. Like, I know it's useful to see which parameters functions exactly take. But there's also the entire internal structure of the file system in there, I'm guessing people won't need that?

Would need some feedback on that from people before we generate a webnative typescript documentation website.

dholms:
Yeah I don't know about all the internal docs, but I imagine some people will want some more detailed, in-depth docs. But like you said, let's wait on feedback before putting up a documentation website :+1:

PR #123, opened 2020-09-28
Fix type check for public tree

dholms:
## Problem
Public write methods aren't getting run on the `pretty` tree. This is because the `PublicTree.instanceOf` method is not formatted correctly.

## Solution
Update `PublicTree.instanceof` to correctly identify public trees. Also update the user of `info`/`header` to be more consistent


icidasset (approved):
👌

PR #124, opened 2020-09-28
Bugfix: Public tree update files

dholms:
## Problem
Updating files on the public tree fails because we incorrectly identity files as directories. This is because files _are_ directories at the protocol level, but _not_ at the platform level 

## Solution
Check if a given node is a file in the node's skeleton (the platform-level description) instead of the node's links (the protocol level description)


icidasset (approved):
🙆‍♂️

PR #125, opened 2020-09-28
Versioning

dholms:
...

### DIFF
--- a/src/fs/protocol/basic.ts
+++ b/src/fs/protocol/basic.ts
@@ -41,6 +41,8 @@ export const getLinks = async (cid: CID): Promise<Links> => {
 }
 
 export const putLinks = async (links: Links): Promise<AddResult> => {
-  const dagLinks = Object.values(links).map(link.toDAGLink)
+  const dagLinks = Object.values(links)
+    .filter(l => l !== undefined)
+    .map(link.toDAGLink)
   return ipfs.dagPutLinks(dagLinks)
 }
--- a/src/fs/protocol/public/index.ts
+++ b/src/fs/protocol/public/index.ts
@@ -7,7 +7,7 @@ import { Metadata } from '../../metadata'
 import { isString } from '../../../common/type-checks'
 import * as check from '../../types/check'
 
-import { isValue } from '../../../common'
+import { isValue, Maybe } from '../../../common'
 import * as ipfs from '../../../ipfs'
 import { CID, FileContent } from '../../../ipfs'
 import * as link from '../../link'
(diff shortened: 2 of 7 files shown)


dholms:
The whitepaper has a `previous` field on a `PrivateDirectory`. Is this necessary?

```
data PrivateDirectory = PrivateDirectory
  { metadata       :: Metadata
  , bareNameFilter :: BareNameFilter
  , revision       :: Natural -- Version counter for this exact AES key
  , previous       :: EncryptedLink
  , links          :: Map Text PrivateLink
  , skeleton       :: Skeleton
  }
```

The `previous` on the `PublicDirectory` has 2 functions:
- keep the CID in the tree so it remains pinned
- have an easy reference back to the previous version so we can go arbitrarily back in time

On the private side, the CID is already in the MMPT so it will remain pinned. & we can easily go arbitrarily back in time by calculating the name filter for the given revision.

Is there another reason that I'm missing?

dholms:
```haskell
data Metadata = Metadata
  { history       :: Maybe History
  , unixMeta      :: UnixMeta
  , floofsVersion :: SemVer
  }
  
data History = History
  { previous :: VirtualNode
  , event    :: Event
  }
```

Is `History` stored at the protocol level? Or calculated when needed at the platform level? It seems like this could get _big_ if we're CBOR encoding the entire history for each node. 

This sentence later on leads me to believe we don't want to do that: "Users do not expect history to be a lightening fast operation. It is still accessible by looking at the concrete (uncached) virtual node" (about the skeleton cache)

However, the other `Metadata` values actually are stored at the protocol level, so I wanted to make sure

expede:
> Is History stored at the protocol level? Or calculated when needed at the platform level? It seems like this could get big if we're CBOR encoding the entire history for each node.

TL;DR It’s not the entire history

```haskell
data History = History
  { previous :: VirtualNode
  , event    :: Event
  }
```

`History` is an extension of the previous link to _include the event that caused that state transition._ In other words, it’s a tuple that make the link back richer. `Event` is not defined in the current whitepaper, since it was written WAY before we got to this stage. It would look something like this (but let’s chat about this on our 1:1 this week):

```haskell
data Event
  = Genesis  -- Initial creation
  | ModifyContent -- Change this file
  | Meta Action FieldName
  | Child Action Path -- Change directory children

data Action
  = Add
  | Modify
  | Delete
```

expede:
> On the private side, the CID is already in the MMPT so it will remain pinned. & we can easily go arbitrarily back in time by calculating the name filter for the given revision.
>
> Is there another reason that I'm missing?

Nope, you’re correct! If we can calculate it at runtime, that gives us a single source of truth. Let’s do that 👍

dholms:
Closed in favor of https://github.com/fission-suite/webnative/pull/126

PR #126, opened 2020-09-29
Concurrent FS operations & basic versioning

dholms:
## Problem
The Filesystem does not allow concurrent operations. Two write operations at the same time will overwrite one another.

## Solution
- make the MMPT concurrent
- make the filesystem concurrent
- also roll in changes for versioning on the public side (https://github.com/fission-suite/webnative/pull/125) to avoid rebases etc

### DIFF
--- a/src/fs/base/tree.ts
+++ b/src/fs/base/tree.ts
@@ -2,10 +2,11 @@
 
 /** @internal */
 import * as pathUtil from '../path'
-import { Tree, File, NonEmptyPath, UnixTree, BaseLinks } from '../types'
+import { Tree, File, UnixTree, BaseLinks, UpdateCallback } from '../types'
 import { SemVer } from '../semver'
 import * as check from '../types/check'
 import { AddResult, CID, FileContent } from '../../ipfs'
+import { Maybe } from '../../common'
 
 
 abstract class BaseTree implements Tree, UnixTree {
@@ -31,17 +32,6 @@ abstract class BaseTree implements Tree, UnixTree {
     return dir.getLinks()
   }
 
-  async mkdir(path: string): Promise<this> {
-    const exists = await this.exists(path)
-    if (exists) {
-      throw new Error(`Path already exists: ${path}`)
-    }
-    const toAdd = await this.emptyChildTree()
-    await this.addChild(path, toAdd)
-    await this.putDetailed()
-    return this
-  }
-
   async cat(path: string): Promise<FileContent> {
     const file = await this.get(path)
     if (file === null) {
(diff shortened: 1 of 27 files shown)


icidasset (commented):
I don't entirely know what's going on, but I think it looks good 😳 👍

expede:
@icidasset I haven't had a chance to look over the code yet, but the idea is to use a "proper" OO architecture (think message passing in Elixir minus the supervisors & actors) to push messages to the tree's leaves, and they respond to the parent wth their updated CID. Because of nice tree properties, this should allow us to just fire a bunch of updates at the tree, and it'll automatically be able to handle the concurrency without data loss as the leaves no longer make assumptions about the CIDs of their parents (which overwrite other changes).

TL;DR it "should" solve timing bugs that cause data loss

src/fs/base/tree.ts
```diff
@@ -52,105 +42,121 @@ abstract class BaseTree implements Tree, UnixTree {
...
+    //   throw new Error(`Path does not exist: ${to}`)
+    // }
+    // const parent = await this.get(parentPath || '')
+    // if (check.isFile(parent)) {
+    //   throw new Error(`Can not \`mv\` to a file: ${parentPath || ''}`)
+    // }
+
+    // await this.updateDirectChild(node, tail, null)
+
+    // return this.rm(from)
```
expede, line 132: Why disabled temporarily? Ahead of versioning stuff? WIP?

icidasset: Found a bug with moving private stuff. Daniel said he'd fix it after these 2 PRs.

dholms: yup^^
> I've realized that it does not work properly on the private side due to 1) lack of signposting & 2) not adjusting the namefilter on move
> Both of these can be fixed, but I'm thinking I might do that after I get concurrency & versioning/events working

I figured it was more important to get these changes out & start testing them. & signposting is going to be related to the events stuff anyway, so I figured I'd implement it then

icidasset:
Status update:
- ✅ `previous` header on public files and trees is correctly added and updated when needed
- ✅ The file system root has been changed from a `BareTree` to a v1 `PublicTree`, so it can have a `previous` header as well.
- ✅ You can get a previous version of a directory/tree with `tree.getRevision(x)`
- 🐞 The method is also present on private files, but it doesn't work yet. Not entirely sure why, something to do with the MMPT I'm guessing. It seems like revisions are not stored in the MMPT, or only the latest revision is kept, don't know.
- 🐞 I've also added a `listRevisions` method, but it probably doesn't work yet.
- 📝 Todo, need to make a "module" for the revision functions, since they're mostly the same in both classes (file and tree classes)

PR #128, opened 2020-09-30
Clear all data when leaving

icidasset:
Wasn't clearing all the data before, which was causing some issues when creating a new account. More specifically, the sdk tried to load a different file system because the CID log was still in the indexeddb. Additionally I cleared the keystore as well, so the user starts with fresh keys.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "webnative",
-  "version": "0.19.7",
+  "version": "0.19.8-alpha-1",
   "description": "Fission Typescript SDK",
   "keywords": [],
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "webnative",
-  "version": "0.19.7",
+  "version": "0.19.8-alpha-1",
```
dholms, line 3: alpha?

dholms (approved):
ran into this earlier so thx for fixing :raised_hands: 

guessing the `alpha` is just from testing, so just remember to bump it before merging/publishing :+1:

PR #132, opened 2020-10-01
UCAN expiration timestamp should not exceed the proof's expiration timestamp

### LINKED ISSUE #130: Shouldn't be able to make invalid UCANs
### Problem

Currently you can make a UCAN that expires later than its proof. This shouldn't be possible. Additionally, less important, but still should be validated as well, the `nbf` should be greater than or equal to the `nbf` of the proof. To be precise regarding the `exp` field, it should be less than or equal to the `exp` of the proof.

### Solution

When making a UCAN, get the `exp` and `nbf` of the proof (if it is present), and then adjust the default values if needed.

icidasset:
Closes #130

### DIFF
--- a/src/ucan.ts
+++ b/src/ucan.ts
@@ -20,7 +20,7 @@ export type Ucan = {
     aud: string,
     exp: number,
     iss: string,
-    nbf: string,
+    nbf: number,
     prf: string | undefined,
     ptc: string | undefined | null,
     rsc: Resource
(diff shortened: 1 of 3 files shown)


src/ucan.ts
```diff
@@ -81,18 +81,30 @@ export async function build({
...
   }
 
+  // Timestamps
+  let exp = currentTimeInSeconds + lifetimeInSeconds
+  let nbf = currentTimeInSeconds - 60
+
+  if (proof) {
+    const prf = decode(proof).payload
+
+    exp = exp <= prf.exp ? exp : prf.exp
```
dholms, line 98: `Math.min`/`Math.max` might read a little bit clearer here

dholms (approved):
:+1:

PR #139, opened 2020-10-19
UCAN v0.5

icidasset:
Implements UCAN according to spec: https://whitepaper.fission.codes/access-control/ucan


src/apps/index.ts
```diff
@@ -24,12 +25,12 @@ export async function index(): Promise<Array<App>> {
...
 
-  const jwt = await ucan.build({
+  const jwt = ucan.encode(await ucan.build({
+    attenuations: [],
     audience: await api.did(),
     issuer: await did.ucan(),
-    proof: ucan.encode(localUcan), 
-    potency: null
-  })
+    proof: ucan.encode(localUcan)
```
expede, line 32: Proofs are also an array in 0.4. Not sure if `ucan.encode` handles that yet, but I'm guessing no.

src/ucan/app.ts
```diff
@@ -0,0 +1,5 @@
+export const PREFIX = "app"
+
+export type Capability
+  = "DESTROY"
+  | "REVISE"
```
expede, line 5: `DESTROY` is now `SUPER_USER` for consistency (as of like 20 minutes ago 😅)

expede: SUPER_USER also gives you the ability to do things like transfer ownership, which is essentially destroy+create (i.e. "move" to another user)

src/apps/index.ts
icidasset: Oh thanks, I missed that.

icidasset:
We'll use https://github.com/fission-suite/ucan instead

PR #41, opened 2020-05-26
Testing

dholms:
Tests for Filesystem

PR #147, opened 2020-12-02
Test setup

walkah:
Setting up for unit & integration (via puppeteer) testing. 

This is a revival of #41 (which got way out of date)


babel.config.js
```diff
@@ -0,0 +1,6 @@
+module.exports = {
+  presets: [
+    ['@babel/preset-env', { targets: { node: 'current' } }],
+    '@babel/preset-typescript'
+  ]
+}
```
icidasset, line 6: Does jest need this to compile individual typescript files, and/or the tests?

src/common/base64.test.ts
```diff
@@ -0,0 +1,10 @@
+/**
+ * @jest-environment jsdom
+ */
+import * as base64 from './base64'
+
+test('can decode an encoded string', () => {
+  const testString = 'ab/c-d=='
+  const encoded = base64.urlEncode(testString)
+  expect(base64.urlDecode(encoded)).toEqual(testString)
```
icidasset, line 9: Does it automatically import `test()` and `expect()`?

icidasset (approved):
Ship iiit 🙏 😄

babel.config.js
walkah: yep, exactly

src/common/base64.test.ts
walkah: it does, which not everyone is a fan of (this is one of the things `ava` explicitly does differently)

PR #154, opened 2020-12-16
Use a shared worker

### LINKED ISSUE #79: Setup shared worker
So that the IPFS instance is shared across tabs.
https://github.com/ipfs/js-ipfs/tree/master/examples/browser-sharing-node-across-tabs
https://github.com/ipfs/js-ipfs/tree/master/packages/ipfs-message-port-client
https://developer.mozilla.org/en-US/docs/Web/API/MessageChannel

icidasset:
Closes  #79

Moves the js-ipfs node to the auth lobby and sets up a tunnel to a shared worker on that lobby, through an iframe. That iframe is inserted automatically when you call `webnative.initialise()`. Additionally this makes webnative work in a web worker.

### DIFF
--- a/src/auth.ts
+++ b/src/auth.ts
@@ -25,13 +25,15 @@ export async function authenticatedUsername(): Promise<string | null> {
  *
  * Removes any trace of the user and redirects to the lobby.
  */
-export async function leave(): Promise<void> {
+export async function leave({ withoutRedirect }: { withoutRedirect: boolean }): Promise<void> {
   await localforage.removeItem(USERNAME_STORAGE_KEY)
   await ucan.clearStorage()
   await cidLog.clear()
   await keystore.clear()
 
-  window.location.href = setup.endpoints.lobby
+  if (!withoutRedirect && globalThis.location) {
+    globalThis.location.href = setup.endpoints.lobby
+  }
 }
 
 /**
--- a/src/ipfs/config.ts
+++ b/src/ipfs/config.ts
@@ -1,48 +1,10 @@
-import loadScript from 'load-script2'
-
+import IpfsMessagePortClient from 'ipfs-message-port-client'
 import { IPFS } from './types'
 import { setup } from '../setup/internal'
 
 
-type IpfsGlobalScope = {
-  Ipfs?: { create: (options: unknown) => IPFS }
-}
-
-type PossibleWorkerGlobalScope = {
-  importScripts?: (url: string) => void
-}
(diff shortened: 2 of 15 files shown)


src/auth.ts
```diff
@@ -25,13 +25,15 @@ export async function authenticatedUsername(): Promise<string | null> {
...
-export async function leave(): Promise<void> {
+export async function leave({ withoutRedirect }: { withoutRedirect: boolean }): Promise<void> {
   await localforage.removeItem(USERNAME_STORAGE_KEY)
   await ucan.clearStorage()
   await cidLog.clear()
   await keystore.clear()
 
-  window.location.href = setup.endpoints.lobby
+  if (!withoutRedirect && globalThis.location) {
+    globalThis.location.href = setup.endpoints.lobby
```
icidasset, line 35: https://caniuse.com/mdn-javascript_builtins_globalthis

I've included a polyfill for this in our `keystore-idb` library which should enable it here as well since it's a global variable.

src/ipfs/config.ts
```diff
@@ -1,94 +1,39 @@
-import loadScript from 'load-script2'
```
walkah, line 1: 🙌 so much cleaner!

src/ipfs/config.ts
```diff
@@ -1,94 +1,39 @@
...
+    iframe.onload = () => {
+      const channel = new MessageChannel()
+      channel.port1.onmessage = ({ ports }) => resolve(ports[0])
+      iframe.contentWindow
+        ? iframe.contentWindow.postMessage("connect", "*", [ channel.port2 ])
+        : reject(new Error("Don't have access to iframe.contentWindow"))
     }
-  }
+
+    iframe.src = `${setup.endpoints.lobby}/ipfs.html`
```
walkah, line 37: So, this loads https://github.com/fission-suite/auth-lobby/blob/158f30c6d38a9d065fa4716c2d8acec684718c70/src/Static/Html/Ipfs.html in an iframe? 

What happens if the user is offline? Does `IpfsMessagePortClient` handle that for us?

walkah (approved):
A couple questions, but looks good - much less heavy lifting for us to handle!

✨ 🚢 🚀

src/ipfs/config.ts
icidasset: Yeah exactly.
The lobby has a service worker which makes it work offline.

PR #161, opened 2021-01-11
v0.21.2

icidasset:
Couple of bugfixes, see CHANGELOG and commits for info.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,6 +56,18 @@ const state = await wn.initialise({
       publicPaths: [ "Mixtapes" ]
     }
   }
+
+}).catch(err => {
+  switch (err) {
+    case wn.InitialisationError.InsecureContext:
+      // We need a secure context to do cryptography
+      // Usually this means we need HTTPS or localhost
+
+    case wn.InitialisationError.UnsupportedBrowser:
+      // Browser not supported.
+      // Example: Firefox private mode can't use indexedDB.
+  }
+
 })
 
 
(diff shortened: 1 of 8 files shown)


src/data-root.ts
```diff
@@ -85,9 +85,6 @@ export async function update(
   if (fetchController) fetchController.abort()
   fetchController = new AbortController()
 
-  // Ensure peer connection
-  await ipfs.reconnect()
```
walkah, line 89: this is no longer needed?

README.md
```diff
@@ -56,6 +56,18 @@ const state = await wn.initialise({
       publicPaths: [ "Mixtapes" ]
     }
   }
+
+}).catch(err => {
```
walkah, line 60: 👍

src/fs/filesystem.ts
```diff
@@ -210,6 +210,9 @@ export class FileSystem {
     if (!sameTree) {
       throw new Error("`mv` is only supported on the same tree for now")
     }
+    if (await this.exists(to)) {
```
walkah, line 213: good addition!

walkah (approved):
Looks good! 🎉 🕺 💃

src/data-root.ts
icidasset: Is handled in https://github.com/fission-suite/auth-lobby/blob/master/src/Javascript/Worker.js#L61 now.

PR #167, opened 2021-01-22
V0.21.3

icidasset:
👀


src/data-root.ts
```diff
@@ -79,11 +79,12 @@ export async function update(
   const apiEndpoint = setup.endpoints.api
 
   // Debug
-  debug.log("🚀 Updating your DNSLink:", cid)
+  debug.log("🌊 Updating your DNSLink:", cid)
```
walkah, line 82: v. important change. 👍 😉

walkah (approved):
image
image

src/data-root.ts
icidasset: Well, we can't have a proper release if the emojis aren't in the same theme 😬

PR #169, opened 2021-01-28
Add resend verification email function

matheus23:
This is for the dashboard.

Not sure whether `auth.ts` is the right file to put this :/ Help! :smile:


icidasset (commented):
Looks good! Would put it `lobby/index.ts` instead 👀

matheus23:
> Looks good! Would put it `lobby/index.ts` instead eyes

Ah! Perfect. Thanks.

icidasset (approved):
🚢  it!

PR #171, opened 2021-01-30
Remove link to docs

bmann:
I don’t think there is a docs folder anymore?


icidasset:
There isn't no, we decided at some point to keep them in a `docs` branch. But we lost that somewhere a long the way. Will generate some HTML docs and put them up.

PR #176, opened 2021-03-04
Leverage multiple read keys and bare name filters given by the lobby

icidasset:
Depends on https://github.com/fission-suite/auth-lobby/pull/50

### DIFF
--- a/src/apps/index.ts
+++ b/src/apps/index.ts
@@ -37,7 +37,10 @@ export async function index(): Promise<Array<App>> {
   })
 
   const data = await response.json();
-  return Object.values(data).map(d => ({ domain: (d as Array<string>)[0] }))
+  return Object
+    .values(data)
+    .filter(v => (v as Array<string>).length > 0)
+    .map(v => ({ domain: (v as Array<string>)[0] }))
 }
 
 /**
--- a/src/apps/index.ts
+++ b/src/apps/index.ts
@@ -1,8 +1,8 @@
-import * as did from './did'
-import * as ucan from './ucan'
-import * as ucanInternal from './ucan/internal'
-import { api, Maybe, isDefined } from './common'
-import { setup } from './setup/internal'
+import * as did from '../did'
+import * as ucan from '../ucan'
+import * as ucanInternal from '../ucan/internal'
+import { api, Maybe, isString } from '../common'
+import { setup } from '../setup/internal'
 
 
 export type App = {
@@ -17,7 +17,7 @@ export type App = {
 export async function index(): Promise<Array<App>> {
   const apiEndpoint = setup.endpoints.api
 
-  const localUcan = await ucanInternal.lookupFilesystemUcan("*")
+  const localUcan = await ucanInternal.lookupAppUcan("*")
   if (localUcan === null) {
     throw "Could not find your local UCAN"
   }
@@ -25,7 +25,7 @@ export async function index(): Promise<Array<App>> {
   const jwt = ucan.encode(await ucan.build({
     audience: await api.did(),
     issuer: await did.ucan(),
-    proof: localUcan, 
+    proof: localUcan,
     potency: null
   }))
 
@@ -37,7 +37,7 @@ export async function index(): Promise<Array<App>> {
   })
 
   const data = await response.json();
-  return data
+  return Object.values(data).map(d => ({ domain: (d as Array<string>)[0] }))
 }
 
 /**
@@ -50,7 +50,7 @@ export async function create(
 ): Promise<App> {
   const apiEndpoint = setup.endpoints.api
 
-  const localUcan = await ucanInternal.lookupFilesystemUcan("*")
+  const localUcan = await ucanInternal.lookupAppUcan("*")
   if (localUcan === null) {
     throw "Could not find your local UCAN"
   }
@@ -58,11 +58,13 @@ export async function create(
   const jwt = ucan.encode(await ucan.build({
     audience: await api.did(),
     issuer: await did.ucan(),
-    proof: localUcan, 
+    proof: localUcan,
     potency: null
   }))
 
-  const url = isDefined(subdomain) ? `${apiEndpoint}/app?${subdomain}` : `${apiEndpoint}/app`
+  const url = isString(subdomain)
+    ? `${apiEndpoint}/app?subdomain=${encodeURIComponent(subdomain)}`
+    : `${apiEndpoint}/app`
 
   const response = await fetch(url, {
     method: 'POST',
@@ -71,20 +73,20 @@ export async function create(
     }
   })
   const data = await response.json();
-  return data
+  return { domain: data }
 }
 
 /**
- * Destroy app by any associated URL
+ * Destroy app by any associated domain
  *
- * @param url The url we want to delete
+ * @param domain The domain associated with the app we want to delete
  */
-export async function deleteByURL(
-  url: string
+export async function deleteByDomain(
+  domain: string
 ): Promise<void> {
   const apiEndpoint = setup.endpoints.api
 
-  const localUcan = await ucanInternal.lookupFilesystemUcan("*")
+  const localUcan = await ucanInternal.lookupAppUcan(domain)
   if (localUcan === null) {
     throw new Error("Could not find your local UCAN")
   }
@@ -92,11 +94,11 @@ export async function deleteByURL(
   const jwt = ucan.encode(await ucan.build({
     audience: await api.did(),
     issuer: await did.ucan(),
-    proof: localUcan, 
+    proof: localUcan,
     potency: null
   }))
 
-  await fetch(`${apiEndpoint}/app/associated/${url}`, {
+  await fetch(`${apiEndpoint}/app/associated/${encodeURIComponent(domain)}`, {
     method: 'DELETE',
     headers: {
       'authorization': `Bearer ${jwt}`
--- a/src/fs/path.ts
+++ b/src/fs/path.ts
@@ -16,9 +16,7 @@ export const joinNoSuffix = (parts: string[]): string => {
 }
 
 export const parent = (path: string): string | null => {
-  const parts = splitParts(path)
-  if (parts.length < 2) return null
-  return join(parts.slice(0, -1))
+  return takeTail(path).parentPath
 }
 
 type HeadParts = {
--- a/src/fs/path.ts
+++ b/src/fs/path.ts
@@ -10,9 +10,15 @@ export const join = (parts: string[]): string => {
 
 export const joinNoSuffix = (parts: string[]): string => {
   const joined = join(parts)
-  return joined[joined.length -1] === '/' 
+  return joined[joined.length -1] === '/'
           ? joined.slice(0, joined.length -1)
           : joined
+}
+
+export const parent = (path: string): string | null => {
+  const parts = splitParts(path)
+  if (parts.length < 2) return null
+  return join(parts.slice(0, -1))
 }
 
 type HeadParts = {
--- a/src/fs/root/tree.ts
+++ b/src/fs/root/tree.ts
@@ -159,6 +159,7 @@ export default class RootTree implements Puttable {
 
   // PRIVATE TREES
   // -------------
+
   findPrivateTree(path: string[]): [string, PrivateTree | null] {
     return findPrivateTree(this.privateTrees, path)
   }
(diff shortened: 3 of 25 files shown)


src/fs/path.ts
```diff
@@ -10,11 +10,17 @@ export const join = (parts: string[]): string => {
 
 export const joinNoSuffix = (parts: string[]): string => {
   const joined = join(parts)
-  return joined[joined.length -1] === '/' 
+  return joined[joined.length -1] === '/'
           ? joined.slice(0, joined.length -1)
           : joined
 }
 
+export const parent = (path: string): string | null => {
```
dholms, line 18: You can also do this with `takeTail`
```ts
return takeTail(path).parentPath
```
but doesn't really matter :shrug: 
I don't love the names `takeTail` and `takeHead` either but couldn't think of better names

src/index.ts
```diff
@@ -290,3 +291,50 @@ function scenarioNotAuthorised(
...
+): Promise<void> {
+  if (!classifiedParam) return
+
+  const ks = await keystore.get()
+  const classifiedInfo = JSON.parse(common.base64.urlDecode(classifiedParam))
+
+  // Extract session key and its iv
+  const iv = utils.base64ToArrBuf(classifiedInfo.iv)
+  const rawSessionKey = await ks.decrypt(classifiedInfo.sessionKey)
+  const sessionKey = await crypto.subtle.importKey(
```
dholms, line 311: can also use `keystore.aes.importKey` if ya like

src/fs/v1/PrivateTree.ts
```diff
@@ -75,17 +79,17 @@ export default class PrivateTree extends BaseTree {
       throw new Error(`Could not parse a valid private tree using the given key`)
     }
 
-    return new PrivateTree({ mmpt, key, header: info })
+    return this.fromInfo(mmpt, key, info)
   }
 
   static async fromName(mmpt: MMPT, name: PrivateName, key: string): Promise<PrivateTree> {
-    const info = await protocol.priv.getByName(mmpt, name, key)
+    const info = await protocol.priv.getByLatestName(mmpt, name, key)
```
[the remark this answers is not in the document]
icidasset: Sounds good, I'll add a new `fromLatestName`, or something like that, instead of changing this function.

src/fs/path.ts
icidasset: Ah cool. Will alias this function that then.

src/index.ts
```diff
@@ -290,3 +291,50 @@ function scenarioNotAuthorised(
...
+    "raw",
+    utils.base64ToArrBuf(rawSessionKey),
+    "AES-GCM",
+    false,
+    [ "encrypt", "decrypt" ]
+  )
+
+  // Decrypt secrets
+  const secrets: Record<string, { key: string, bareNameFilter: string }> =
+    JSON.parse(utils.arrBufToStr(await crypto.subtle.decrypt(
```
dholms, line 321: and we have decryption helper functions in `webnative/keystore`

is there some reason not to use those?

src/fs/root/tree.ts
```diff
@@ -92,13 +100,13 @@ export default class RootTree implements Puttable {
     // Load private bits
     const privateCID = links[Branch.Private]?.cid || null
 
-    let mmpt, privateTree
+    let mmpt, privateTrees
     if (privateCID === null) {
       mmpt = await MMPT.create()
-      privateTree = await PrivateTree.create(mmpt, key, null)
+      privateTrees = await makeNewPrivateTrees(keys, mmpt)
```
dholms, line 106: is there a scenario in which the app is creating the private tree & also needs to permission & create multiple sub-trees?

dholms: oh is this for scaffolding the initial private tree structure? (ie in Drive?)

dholms: wait but it also errors if the parnetNameFilter is not available :thinking: So wouldn't we only be able to create the one?

icidasset: This needs more testing, but essentially it:
1. Sorts the keys alphabetically. So if you have a key for the root, it'll always create the root tree first.
2. Find the parent name filters for creating sub trees. It does this by either looking in indexeddb, or by finding a parent tree and then getting the namefilter that way.

> oh is this for scaffolding the initial private tree structure? (ie in Drive?)

Yes. Also replacing the old `empty`

src/index.ts
icidasset: Yeah I didn't use that because it does an `importKey` and `decrypt` in one function, and here I need to do multiple decrypts with the same key. I only wanted to do one key import.

src/index.ts
icidasset: Ah nice, thx, will look at that.

src/index.ts
dholms: ah got it got it. you can do this with `keystore.aes.encrypt` as well then to avoid multiple imports :ok_hand: 

it accepts either a key string or an imported key as a param

src/fs/root/tree.ts
dholms: ahhh ok cool cool yeah that makes sense :ok_hand:

dholms (approved):
This looks really good :ok_hand: 

I'd clean up the crypto methods a bit (and test making a new tree a bit to make sure that logic is solid), but besides that I'd say it's good to go

icidasset:
Removed the complex `makeNewPrivateTrees` function and now requires you to pass the root aes key to `FileSystem.empty` when creating a new filesystem.

Reason for this being, the lobby generates a root key before we create the filesystem. Because, that key could be shared with another device, and can also occur before creating the filesystem.

icidasset:
Almost there. Still need to add Fission `app` permissions (see `app` https://github.com/fission-suite/auth-lobby#how-does-it-work) and needs more testing.

src/apps/index.ts
```diff
@@ -37,7 +37,7 @@ export async function index(): Promise<Array<App>> {
   })
 
   const data = await response.json();
-  return data
+  return Object.values(data).map(d => ({ domain: (d as Array<string>)[0] }))
```
matheus23, line 40: This won't work for what I get returned from the API:
```json
{
    "427": [
        "elegant-small-old-wolf.fission.app"
    ],
    "1291": [
        "victorious-short-azure-witch.fission.app"
    ],
    "1292": [
        "thin-azure-ghost.fission.app"
    ],
    "1293": [
        "petite-oval-teal-tulip.fission.app"
    ],
    "1294": [
        "web-uploaded-app.fission.app"
    ],
    "1295": []
}
```
The "1295" key contains an empty list.

Personally, I'd rather change the type than the implementation. I feel like there's a reason why the backend returns a dictionary of lists, and maybe we should reflect that in the API of webnative.

Thoughts?

icidasset: Oh ok, thanks, I missed that. What's the reasoning here with the keys and lists? (ie. what is this key exactly and why are we working with lists)

matheus23: It was on my plate to figure this out for a long time now, but I haven't gotten to it, yet. So: I don't know.

I just talked to James. It might make sense to just assume that there's one domain per app (not zero and not more than that), and the empty list *might* be a bug.

We *could* design the API to be more future-proof, though: At some point, there's going to be multiple domains per app.

However, e.g. the dashboard is making the assumption that there's one domain per app at the moment (and filtering out apps without domains), so :man_shrugging: 

I'll take a look at the server.

bmann: There needs to be a "main" domain per app, that always works. That is the `*.fission.app` subdomain. So keep it with that for now, and when we have automated addition of custom domains.

matheus23: Ok, so I took a look.
Here's the code for the `.index()` handler:</summary>
```hs
index :: (MonadDB t m, App.Retriever t, App.Domain.Retriever t) => ServerT API.App.Index m
index Authorization {about = Entity userId _} = runDB do
  apps        <- App.ownedBy userId
  appXDomains <- forM apps findDomains
  let normalized = (\(k, v) -> (fromIntegral (fromSqlKey k), v)) <$> appXDomains
  return (Map.fromList normalized)
  where
    findDomains :: App.Domain.Retriever m => Entity App -> m (AppId, [URL])
    findDomains (Entity appId _) = do
      appDomains <- App.Domain.allForApp appId
      return (appId, toURL <$> appDomains)

    toURL :: Entity AppDomain -> URL
    toURL (Entity _ AppDomain {..}) = URL appDomainDomainName appDomainSubdomain
```

So it's a dictionary of apps by SQL key with a list of associated domains for each.

Also, there's two things you can destroy via the REST API: Apps by domain and apps by ID. If you destroy by domain, you basically just remove the domain entry from the app's list of domains. However, **it's still there**. If you want to, you can re-add a domain to it via the ID you get from `.index()`.
If you destroy it by ID, it's just gone, but that's not an API that webnative exposes at the moment.

More related code in `fission-web-server/library/Fission/Web/Server/Handler/App/Destroy.hs` and `fission-web-server/library/Fission/Web/Server/App/Destroyer/Class.hs`
```hs
handler :: (MonadTime m, MonadThrow m, MonadLogger m, App.Destroyer m) => ServerT API.App.Destroy m
handler = destroyByURL :<|> destroyById

destroyByURL :: (MonadTime m, MonadThrow m, MonadLogger m, App.Destroyer m) => ServerT API.App.Destroy.ByURL m
destroyByURL URL {..} Authorization {about = Entity userId _} = do
  now <- currentTime
  Web.Error.ensureM $ App.destroyByURL userId domainName subdomain now
  return NoContent

destroyById :: (MonadTime m, MonadThrow m, MonadLogger m, App.Destroyer m) => ServerT API.App.Destroy.ById m
destroyById appId Authorization {about = Entity userId _} = do
  now <- currentTime
  Web.Error.ensureM $ App.destroy userId (toSqlKey $ fromIntegral appId) now
  return NoContent
```
and
```hs
class Monad m => Destroyer m where
  destroy ::
       UserId
    -> AppId
    -> UTCTime
    -> m (Either Errors' [URL])

  destroyByURL ::
       UserId
    -> DomainName
    -> Maybe Subdomain
    -> UTCTime
    -> m (Either Errors' [URL])

instance MonadIO m => Destroyer (Transaction m) where
  destroy userId appId now = do
    appDomains <- AppDomain.allForApp appId
    destroyAssociated userId appId appDomains now

  destroyByURL userId domainName maySubdomain now =
    AppDomain.allSiblingsByDomain domainName maySubdomain >>= \case
      Left  errs ->
        return $ relaxedLeft errs

      Right [] ->
        return . openLeft $ NotFound @AppDomain

      Right appDomains@(Entity _ AppDomain {appDomainAppId} : _) ->
        destroyAssociated userId appDomainAppId appDomains now

destroyAssociated ::
  MonadIO m
  => UserId
  -> AppId
  -> [Entity AppDomain]
  -> UTCTime
  -> Transaction m (Either Errors' [URL])
destroyAssociated userId appId appDomains now =
  App.byId userId appId >>= \case
    Left err ->
      return $ relaxedLeft err

    Right (Entity _ app) ->
      if isOwnedBy userId app
        then do
          putMany (toEvent now <$> appDomains)
          deleteCascade appId
          return $ Right (extractURL <$> appDomains)
        else
          return . openLeft $ ActionNotAuthorized @App userId
```

---

I think it makes sense to reflect this API soon? But probably in another PR. I was planning on overhauling the API from what I've learned working with it in the dashboard anyway.

icidasset: Ah ok, sounds good 👍  I can change it in this PR, so basically instead of the `{ domain }` response, you'd like `{ appId, domains }`?

walkah: > I just talked to James. It might make sense to just assume that there's one domain per app (not zero and not more than that), and the empty list _might_ be a bug.

Yeah, as @bmann mentioned, we should always have one stable domain (APP_NAME.fission.app) but we'll have additional custom domains eventually too.

matheus23: > Ah ok, sounds good +1 I can change it in this PR, so basically instead of the { domain } response, you'd like { appId, domains }?

I think it's likely that I'd touch the API with a PR of my own as a result of finishing the current dashboard slice :sweat_smile: 

> Yeah, as @bmann mentioned, we should always have one stable domain (APP_NAME.fission.app) but we'll have additional custom domains eventually too.

Hm. I agree that this would be nice, but at the moment it's pretty easy to break this invariant using the REST API. What do we do? Can we assume this invariant to hold if it's breakable, or do we roll with the way it works right now? Or do we fix the server, too? Or do we enforce the invariant, but only in webnative?

Sorry, that's starting to get quite off-topic. I think this is not a problem with the this PR anymore. :smile:

icidasset: > I think it's likely that I'd touch the API with a PR of my own

Sure thing, I'll just do a quick fix then, filtering out the empty lists.

PR #177, opened 2021-03-05
backport changes from #139 for current spec

walkah:
- backport better ucan header / payload types
- alter encode/decode & signing
- add `ucan.isValid`  and update test

### DIFF
--- a/src/ucan.ts
+++ b/src/ucan.ts
@@ -6,7 +6,11 @@ import { base64 } from './common'
 
 // TYPES
 
-export type Fact = any
+export type SessionKey = {
+  sessionKey: string
+}
+
+export type Fact = SessionKey | Record<string, string>
 
 export type Resource =
   "*" | Record<string, string>
@@ -202,7 +206,7 @@ export function encode(ucan: Ucan): string {
 export function encodePayload(payload: UcanPayload): string {
   return base64.urlEncode(JSON.stringify({
     ...payload,
-    prf: payload.prf.map(encode)
+    prf: payload.prf.slice(1).map(encode) // TODO: 0.3.1 only supports a single proof.
   }))
 }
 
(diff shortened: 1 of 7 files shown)


src/ucan.ts
```diff
@@ -176,12 +177,33 @@ export function decode(ucan: string): Ucan  {
...
+
+/**
+ * Encode the payload of a UCAN.
+ *
+ * @param payload The UcanPayload to encode
+ */
+export function encodePayload(payload: UcanPayload): string {
+  return base64.urlEncode(JSON.stringify({
+    ...payload,
+    prf: payload.prf.map(encode)
```
icidasset, line 205: Does the server support a list at the moment? 👀

icidasset (approved):
Looks great, thanks! 👏 🎉 
Had one question regarding proofs.

src/ucan.ts
walkah: @icidasset oh... good catch! server *is* still single proof: https://github.com/fission-suite/fission/blob/main/fission-core/library/Fission/Web/Auth/Token/JWT/Types.hs#L132

thanks!

walkah:
Huh. The fs concurrency test is failing ... I'll have a look at that too 🤔

src/ucan.ts
```diff
@@ -1,35 +1,38 @@
 import { CryptoSystem } from 'keystore-idb/types'
 
+import * as did from './did'
 import * as keystore from './keystore'
 import { base64 } from './common'
 
-
 // TYPES
 
+export type Fact = any
```
walkah, line 9: 🤔 this should get cleaned up to better match server's Fact type

walkah:
@icidasset after doing a little more testing, I rolled back to single proof - keeps our types more explicit and in-step with server for now. we can update when server is ready. cool?

icidasset:
Aight, sounds good 👍

PR #180, opened 2021-03-11
Fix regex for usernames & allow underscores

dholms:
## Problem
Today, users are restricted to non-underscored usernames and app names (subdomains).

also our regex check isn't properly disallowing bad usernames (worst case our server rejects but would still be nice to give a user a heads up)

## Solution
- Allow underscores in the validator
- disallow usernames that start with underscores so that we don't have conflicts with _did, _dataroot, _dnslink, etc
- fix regex so it doesn't allow special characters (someone plz double check my regex! it appears to work well, but regex has always been a weak spot for me)
- lowercase usernames before checking against block list
- still allow uppercase because that all gets normalized on the server

### DIFF
--- a/src/lobby/username.ts
+++ b/src/lobby/username.ts
@@ -19,6 +19,6 @@ export function isUsernameValid(username: string): boolean {
   return !username.startsWith("-") &&
          !username.endsWith("-") &&
          !username.startsWith("_") &&
-         /^[a-zA-Z1-9_-]*$/.test(username) &&
+         /^[a-zA-Z1-9_-]+$/.test(username) &&
          !USERNAME_BLOCKLIST.includes(username.toLowerCase())
 }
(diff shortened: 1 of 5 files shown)


dholms:
Sidenote: I split username out into it's own module because jest is blowing up when trying to import some node_modules :man_facepalming: 

but that's definitely something we should look at soon

src/lobby/username.ts
```diff
@@ -0,0 +1,24 @@
...
+}
+
+/**
+ * Check if a username is valid.
+ */
+export function isUsernameValid(username: string): boolean {
+  return !username.startsWith("-") &&
+         !username.endsWith("-") &&
+         !username.startsWith("_") &&
+         /^[a-zA-Z1-9_-]*$/.test(username) &&
```
icidasset, line 22: I think the `*` has to be a `+` here, so the string can't be empty.

icidasset (approved):
Nice 👌  Looks good, except for one minor thing 👍 

> jest is blowing up when trying to import some node_modules

Yeah I've had that issue months ago, I still don't know what that's about.

src/lobby/username.ts
dholms: Good looks :+1:

PR #188, opened 2021-03-30
Generate the src/common/version.ts file from the package.json

### LINKED ISSUE #183: Dynamically set VERSION in src/common/version.ts
Attempted to do it the following way:
https://github.com/fission-suite/webnative/commit/8bec07aae58aaf1b42e67ae0526e9d997a38737d#diff-9bdfafd0d1cd513e6f33e5419d76e970775a804993ac44c17368951f25c4ce5b
https://github.com/fission-suite/webnative/commit/8bec07aae58aaf1b42e67ae0526e9d997a38737d#diff-b55cdbef4907b7045f32cc5360d48d262cca5f94062e353089f189f4460039e0

But that caused problems with publishing the library.
Because Typescript copies the `package.json` file to `dist` and NPM doesn't like that?

### Possible solution

> Another idea: generate a tiny .ts file from the pkg version in the rollup config
> side effects in config files for the win
> Or just as another step in the build process
> The generated ts file would then contain the version from package.json

– @matheus23

matheus23:
Fixes #183


matheus23:
Side note: Should we drop node 10 compatibility? Official "end of life" for node 10 is going to be end of this month. (April 30th 2021)

scripts/gen-version.js
```diff
@@ -0,0 +1,11 @@
+const fs = require("fs").promises
+
+fs.readFile("package.json", { encoding: "utf-8" }).then(pkg => {
+    const pkgJson = JSON.parse(pkg)
+    const version = pkgJson.version
+    const versionModule = `export const VERSION = ${JSON.stringify(version, null, 4)}` + "\n"
```
icidasset, line 6: Is there a specific reason the space (ie. `4`) is passed as a param to `JSON.stringify`? Shouldn't this be `0`?

icidasset (approved):
Looks good 👍  Added one question.

> Should we drop node 10 compatibility?

We can upgrade yeah.

scripts/gen-version.js
matheus23: Ideally, it shouldn't ever make a difference whether it's 4 or 0. It would only apply if the stringified thing would be an array or object - and the version key in `package.json` should only ever be a string.

But *in the absurd case it is something else* it'll lead to nicely-formatted code in `version.ts`? :smile:

icidasset: True 😄 For some reason I was thinking this could turn out in in `"    version"`, but indeed, it would probably be `    "version"`

PR #191, opened 2021-04-08
Improve permission path handling

icidasset:
This replaces

```js
initialise({
  permissions: {
    fs: {
      privatePaths: [ "Music" ],
      publicPaths: [ "Mixtapes" ]
    }
  }
}
```

with

```js
initialise({
  permissions: {
    fs: {
      private: { directories: [ "Music" ], files: [] },
      public: { directories: [ "Mixtapes" ], files: [] }
    }
  }
}
```

Then it passes the paths to the lobby in the following way:
- Directory: `foo/bar/`
- File: `foo/bar`

Also cleans up the way functions, that use permissions, deal with paths.

### Why is this necessary?

The SDK and lobby currently assume every filesystem permission is a directory, but we need to support files too.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,10 +47,10 @@ const state = await wn.initialise({
       creator: "Nullsoft"
     },
 
-    // Ask the user permission for additional filesystem paths
+    // Ask the user permission to additional filesystem paths
     fs: {
-      privatePaths: [ "Music" ],
-      publicPaths: [ "Mixtapes" ]
+      private: { directories: [ "Audio/Music" ], files: [] },
+      public: { directories: [ "Audio/Mixtapes" ], files: [] }
     }
   }
 
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
     "package.json",
     "docs"
   ],
-  "typings": "types/index.d.ts",
+  "types": "dist/index.d.ts",
   "author": "Daniel Holmgren <daniel@fission.codes>",
   "repository": {
     "type": "git",
@@ -39,6 +39,7 @@
     "test": "jest",
     "test:watch": "jest --watch",
     "test:prod": "yarn run build && yarn run lint && yarn run test -- --no-cache",
+    "test:types": "cp -RT tests/types/ dist/ && yarn run tsd",
     "test:unit": "jest --watch --testPathPattern=src/",
     "precommit": "lint-staged",
     "prepublishOnly": "yarn build && yarn docs",
@@ -111,5 +112,8 @@
   },
   "resolutions": {
     "**/ipfs-message-port-protocol": "https://ipfs.runfission.com/ipfs/bafybeigx6q4aezve7my76s5vvfuiinbxtepapqvmjf2jbgrozrut6cjape/p/ipfs-message-port-protocol.tar.gz"
+  },
+  "tsd": {
+    "directory": "dist"
   }
 }
--- a/src/auth.ts
+++ b/src/auth.ts
@@ -7,7 +7,7 @@ import * as keystore from './keystore'
 import * as ucan from './ucan/internal'
 import { USERNAME_STORAGE_KEY, Maybe, VERSION } from './common'
 import { FileSystem } from './fs/filesystem'
-import { Permissions } from './ucan/permissions'
+import { Permissions, fileSystemPaths } from './ucan/permissions'
 import { setup } from './setup/internal'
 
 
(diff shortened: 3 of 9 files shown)


README.md
```diff
@@ -47,10 +47,10 @@ const state = await wn.initialise({
       creator: "Nullsoft"
     },
 
-    // Ask the user permission for additional filesystem paths
+    // Ask the user permission to additional filesystem paths
```
matheus23, line 50: doesn't really matter, but I think it should be "Ask the user *for* permission to additional filesystem paths" :slightly_smiling_face:

walkah (approved):
🎉 Much cleaner API 👍

README.md
icidasset: Ah yes that's better 👌

expede:
Why break out files vs directories specifically? Is there an advantage over paths?

expede:
Also some light ambiguity on the wording:

`directories: [ "Music" ], files: []`

Can I get _multiple_ directories, or is this a single path down the DAG? I'm _guessing_ that it's a single line of directories, and possibly many files at the ends? Or can you have multiple directories and multiple files?

expede:
One suggestion for how to get rid of that ambiguity would be to tag it with a `type` field


```ts
  //////////////////
 // SINGLE FILES //
//////////////////
private: {
  type: FILE,
  segments: ["foo", "bar"]
}

// or

private: {
  type: FILE,
  path: "foo/bar"
}

// or

private: {
  filePath: ["foo", "bar"]
}

  ////////////////
 // MANY FILES //
////////////////

private: [
  { file: ["foo", "bar"] },
  { directory: ["baz", "quux"] },
  { file: ["vacation.png"] }
]

// and other variants above
```

icidasset:
> Why break out files vs directories specifically? Is there an advantage over paths or segments? Should we define all paths in this format for consistency, and make it a TS `interface`?

Because we want to support permissions for files too.
If we'd stick with the paths, `foo/bar` would become a file path (as opposed to a directory path, like it is now).
Having to depend on a `/` at the end of a path is confusing, and a lot of people will forget to do just that.

I would like to have a better interface for all the paths in webnative, but not sure yet what that would exactly look like.

> Can I get multiple directories, or is this a single path of segments down the DAG? I'm guessing that it's a single line of directories, and possibly many files at the ends? Or can you have multiple directories and multiple files?

Multiple directories and files.
```
directories: [ "Music/Playlists", "Documents" ], files: [ "Documents/Invoice.pdf" ]
```

expede:
> Because we want to support permissions for files too.
> If we'd stick with the paths, foo/bar would become a file path (as opposed to a directory path, like it is now).
> Having to depend on a / at the end of a path is confusing, and a lot of people will forget to do just that.

Yes understood. Refresh the browser — I updated my comment to clarify that we may want to use tags on paths or segments.

expede:
> Multiple directories and files.
> 
> directories: [ "Music/Playlists", "Documents" ], files: [ "Documents/Invoice.pdf" ]

Ah, I see. So still using string paths. Yup that totally works! If we have examples or docs, ones with `/`s in them would help clarify.

It would also be nice to have a universal way of disambiguating between directories and files, but I honestly don't know if that's common enough of a use case to need to explicitly separate file or directory elsewhere.

icidasset:
I do like these interfaces 👌 
```
{
  type: FILE,
  pathSegments: ["foo", "bar"]
}
```
Would be my preferred option.
Do you think we should tackle that now, or later with the big SDK overhaul?

icidasset:
> It may also be nice to have a universal way of disambiguating between directories and files, but I honestly don't know if that's common enough of a use case to need to explicitly separate files vs directories elsewhere 🤷‍♀️

I don't think so? Need to think about it more. But I'd assume not, because the various filesystem operations already assume one type or another. For example, `ls` assumes directories, `read` assumes files.

expede:
> Do you think we should tackle that now, or later with the big SDK overhaul?

How much effort would it be today? Minutes vs hours vs days?

---

Also possibly worth noting that UCANs follow the format `{TAG: {...permissions}}`, rather than `{type: _, permissions: _}`. They're strictly equivalent, so it really doesn't bother me one way or another, but thought I'd mention it!

icidasset:
> Also possibly worth noting that UCANs follow the format

I like `{TAG: {...permissions}}` as well 👍 
I don't think this is something we should tackle now, would be a couple of days of work.
Rather do this properly from the get-go in the overhaul.

expede:
> Rather do this properly from the get-go in the overhaul.

Okay sounds good.

Should we hold off on this PR as well in that case, since these are also file paths that we may want to have also be consistent? I'm honestly happy either way 👍

expede (approved):
Makes sense; thanks for answering my questions and brainstorming @icidasset

icidasset:
Don't think we should hold off with this, since the type of paths used is the same (ie. just strings as opposed to an interface). Would like to get this in so the lobby can provide permissions for files, which is something Daniel will probably need for his keychain work.

PR #197, opened 2021-04-15
Add support for ED25519 & BLS DIDs

icidasset:
And verification of ED25519 signed data.


src/index.ts
```diff
@@ -246,6 +246,7 @@ export * as ucan from './ucan'
 export * as dns from './dns'
 export * as ipfs from './ipfs'
 export * as keystore from './keystore'
+export * as common from './common'
```
icidasset, line 249: What's the reason you've added this @dholms?

dholms: I wanted to use one of the common methods (`makeUrlUnsafe`) in the tests instead of rewriting logic, and didn't see any harm in exporting the methods. But can go either way with it honeslty :man_shrugging:

icidasset: Ah gotcha 👍  I'll rename this to something else, because `common` doesn't make much sense from the user's perspective.

icidasset: Renamed to `webnative.machinery`

dholms: love it

expede:
Closes #202

PR #203, opened 2021-04-16
Notify the lobby we're using a shared worker

icidasset:
Lets us optimise for the fast path.
Adds the temporary auth data to the shared worker via ipfs, instead of to the user's filesystem.

### DIFF
--- a/src/auth.ts
+++ b/src/auth.ts
@@ -59,6 +59,7 @@ export async function redirectToLobby(
 
   const exchangeDid = await did.exchange()
   const writeDid = await did.write()
+  const sharedRepo = !!document.body.querySelector("iframe[src$=\"/ipfs.html\"]") && typeof SharedWorker === "function"
 
   redirectTo = redirectTo || window.location.href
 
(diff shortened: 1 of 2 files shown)


src/auth.ts
```diff
@@ -59,6 +59,7 @@ export async function redirectToLobby(
 
   const exchangeDid = await did.exchange()
   const writeDid = await did.write()
+  const sharedRepo = !!document.body.querySelector("iframe[src$=\"/ipfs.html\"]") && typeof SharedWorker === "function"
```
matheus23, line 62: I feel like *eventually* we want to have some kind of webnative class/instance/state.
Since the code to initialize the shared worker is in webnative's control, it should be possible to keep track of that as state, instead of relying on this.

But that'd be a bigger refactor, just throwing that over the fence :smile:

matheus23 (approved):

src/auth.ts
walkah: I agree that this feels a little fragile. Perhaps as an interim step we could put an `id` or something on the iframe so at least we have a little more confidence we're checking the right one.

icidasset: Good point, added 👌

walkah (approved):
👍 👀 Looks good to me!

PR #204, opened 2021-04-16
Dependency Injection

dholms:
## Problem
`webnative` was created with the browser in mind and relies on certain browser apis such as WebCrypto and IndexedDB. This prevents it from being used in node and makes it difficult to test

## Solution
Inject all the dependencies!

- add a new method to `setup`: `setDependencies`
- a user can supply implementations for crypto methods, keystore methods, and persistent storage
- abstract these methods into `crypto` and `storage` modules that the rest of the code consumes
- set defaults to currently used browser implementations

Note: The SDK isn't totally node-ready. There's still a bunch of stuff with the web-worker, lobby redirects, ipfs, etc that is browser specific. But this code only gets called when initializing & getting the FS. This PR is mainly directed towards getting the helper functions ready to be used & tested in Node. We'll need to figure out the auth workflow before using the SDK from node.

---
### Build config
I also took this chance to build the CommonJS format for Node by passing `{ browser: false, preferBuiltins: true }` to the `nodeResolve` plugin. Es5 & UMD are still built with browser in mind


src/common/base64.ts
```diff
@@ -1,9 +1,17 @@
-export function urlDecode(a: string): string {
-  return atob(makeUrlUnsafe(a))
+export function decode(base64: string): string {
+  return Buffer.from(base64, 'base64').toString('binary')
```
dholms, line 2: we inject `buffer` in the rollup config. in the browser it just uses Uint8Arrays under the hood. I'm using this pattern in `keystore-idb` and `webnative-filecoin` and quite like it

src/crypto/browser.ts
```diff
@@ -0,0 +1,122 @@
...
+  return new Uint8Array(decryptedBuf)
+}
+
+export const genKeyStr = async (): Promise<string> => {
+  assertBrowser('aes.genKeyStr')
+  const key = await aes.makeKey({ length: SymmKeyLength.B256 })
+  return aes.exportKey(key)
+}
+
+export const decryptGCM = async (encrypted: string, keyStr: string, ivStr: string): Promise<string> => {
```
dholms, line 31: This feels a bit clunky and I'm not a huge fan. But we're using `AES-CTR` in the FS for now and `AES-GCM` for the auth lobby. Once we had an algo description to links in the FS, then we can standardize on `AES-GCM` for backwards-compatibility.

Also: in `keystore-idb`, I just prepend the `iv` to the encrypted message. So there is a slightly different api between these two

src/crypto/browser.ts
```diff
@@ -0,0 +1,122 @@
...
+import utils from 'keystore-idb/utils'
+import aes from 'keystore-idb/aes'
+import { CharSize, SymmKeyLength } from 'keystore-idb/types'
+
+import { assertBrowser } from '../common/browser'
+import * as keystore from '../keystore'
+
+
+export const encrypt = async (data: Uint8Array, keyStr: string): Promise<Uint8Array> => {
+  assertBrowser('aes.encrypt')
```
dholms, line 12: Give user's better error messages if they try to call a browser-specific method in node

src/did.ts
```diff
@@ -226,9 +215,11 @@ const hasPrefix = (prefixedKey: ArrayBuffer, prefix: ArrayBuffer): boolean => {
...
-  if(cryptoSystem === CryptoSystem.RSA) {
-    return KeyType.RSA
+const toKeyType = (str: string): KeyType => {
+  switch(str) {
+    case 'rsa': return KeyType.RSA
+    case 'ed25519': return KeyType.Edwards
+    case 'bls12-381': return KeyType.BLS
   }
-  throw new Error(`Key type '${cryptoSystem}' not supported`)
+  throw new Error(`Key Type ${str} not support`)
```
icidasset, line 224: Typo, "supported"?

src/keystore/index.ts
```diff
@@ -1,2 +1,8 @@
+import * as config from './config'
 export * from './config'
-export * from './basic'
+
+
+export const getKeyByName = async (keyName: string): Promise<string> => {
```
icidasset, line 5: Is this the same as `crypto.keystore.exportSymmKey`?

icidasset (approved):
Looks great!

src/did.ts
dholms: yes ty! :pray:

src/keystore/index.ts
dholms: ah yup :+1: I meant to delete this. good looks

src/setup/dependencies.ts
```diff
@@ -0,0 +1,83 @@
...
+    sign: (message: string, charSize: number) => Promise<string>
+    importSymmKey: (key: string, name: string) => Promise<void>
+    exportSymmKey: (name: string) => Promise<string>
+    keyExists: (keyName:string) => Promise<boolean>
+    getAlg: () => Promise<string>
+    clear: () => Promise<void>
+  }
+  storage: {
+    getItem: <T>(key: string) => Promise<T | null>
+    setItem: <T>(key: string, val: T) => Promise<T>
```
matheus23, line 79: Why is this polymorphic? Shouldn't it be e.g. `getItem: (key: string) => Promise<string | null>`?

matheus23 (approved):
:+1: 

<sub>(Secretly can't wait for a webnative v2 in which we switch from a mutation-based setup - i.e. `webnative.setup.endpoints(..)` and `webnative.setup.setDependencies`,etc. - to an immutable setup, similar to how e.g. js-ipfs is initialized)<sup>

src/common/browser.ts
```diff
@@ -0,0 +1,8 @@
+// Took this one-liner from: https://www.npmjs.com/package/browser-or-node
+export const isBrowser = typeof window !== 'undefined' && typeof window.document !== 'undefined'
+
+export const assertBrowser = (method: string): void => {
```
walkah, line 4: I like this! It's nice to let people know why something isn't working as expected. But, maybe i missed it - are we not going to provide a node implementation as well?

walkah (approved):
Asian Man Thank You - m2WQBH7DhSdgLfnTUi
looking good northern lights - KIxBn0rcMbhII

🎉  thanks for taking this on! We need to co-ordinate versions as @icidasset has some 0.24.0-alpha versions published... probably makes sense to have this all as 0.24.0

src/setup/dependencies.ts
dholms: These are the types that `localforage` gives back. Much of the time you can expect a string but it doesn't have to be a string. For instance, our keys are stored in `localforage` as `CryptoKeys` not as strings

src/common/browser.ts
dholms: You didn't miss it. I just felt that was out of scope for this PR. While working on the filecoin stuff, the philosophy I've been taking with changes to `webnative` & the auth-lobby etc is: do the smallest change possible to allow what you need that isn't hacky and doesn't accrue tech debt.

So in otherwords, this is all I need for the `fil-cosigner` server, it allows developers to use `webnative` in Node (with some extra work) and Brian to right tests. Node defaults coming soon :grin:

PR #205, opened 2021-04-20
Keep existing UCANs in their original state

icidasset:
The idea here is to follow the convention of the fission server, and that's to never re-encode a UCAN. For example, we ran into issues because of a `prf: null` and `prf: undefined` inconsistency. The reason this was happening is because we decoded UCANs from other sources and then re-encoded them. This PR changes that.

Also fixes a issue with `ucan.encode`, namely that it could return an unexpected Promise.

### DIFF
--- a/src/data-root.ts
+++ b/src/data-root.ts
@@ -70,7 +70,7 @@ export async function lookupOnFisson(
  */
 export async function update(
   cid: CID | string,
-  proof: Ucan
+  proof: string
 ): Promise<{ success: boolean }> {
   const apiEndpoint = setup.endpoints.api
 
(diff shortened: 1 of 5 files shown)


icidasset:
Also important to know. The lobby will generate "old style" UCANs with `prf: undefined` for apps using the older SDK (version < 0.24), that way everything continues working.

expede:
> Also important to know. The lobby will generate "old style" UCANs with `prf: undefined` for apps using the older SDK (version < 0.24), that way everything continues working.

What's the updated version? `null`?

icidasset:
> What's the updated version? `null`?

`null` indeed.

PR #206, opened 2021-04-21
Update Niv, Nix shell hook, and gitignore

expede:
Screen Shot 2021-04-21 at 11 41 34 AM

...the same will happen with `lorri`


.gitignore
```diff
@@ -3,3 +3,209 @@
 /docs
 /node_modules
 /yarn-error.log
+
```
expede, line 6: 👇 Protip: https://gitignore.io

shell.nix
```diff
@@ -6,9 +6,12 @@ in
...
   buildInputs = [
-    pkgs.just
     unstable.yarn
     unstable.nodejs-15_x
     unstable.niv
   ];
+
+  shellHook = ''
+    ${unstable.yarn}/bin/yarn install
+  '';
```
expede, line 16: The main event

icidasset:
Same comment as in https://github.com/fission-suite/auth-lobby/pull/68 regarding gitignore file. Looks good otherwise 👍

expede:
Looking at offline support: this one doesn't have the problem from the auth-lobby because no `just` action deleting stuff

icidasset (approved):

Issue #207, opened 2021-04-22
File system "warming up" forever


expede:
For example, the auth lobby with warm up forever and never redirect back to the app

### Reproduced By
* Brooke
* Boris

### Environment
* macOS 11.2 / Safari 14
* macOS 11.2 / Firefox 87

### Ruled Out
* [ ] IPFS Cluster
* [ ] Web API
* [ ] DNS
* [ ] SDK
* [ ] Application code

expede:
I'm uncertain if this is webnative or the auth lobby, but someone mentioned similar in Drive, so starting here. Feel free to move if this isn't the case.

matheus23:
I believe the root issue for this is fixed. What I'm saying is: New accounts should not have this issue unless they use apps with an outdated webnative or malicious apps.

The remaining thing to fix here is to make sure that webnative handles corrupt private filesystems gracefully.

expede:
> or malicious apps

@matheus23 can you explain the threat model?

matheus23:
Not exactly sure what you mean by threat model (are we talking a simple explanation or actual diagrams?).

Here's a short explanation for a start:
* A user creates an account
* The user logs in with fission in an app that's running malicious code, giving it permission to access (parts of) their private file system
* The app code will create a private file header and add it to the user's private file system
* The app code *doesn't* add the file content to the private file system. Thus the CID in the private file *header* is pointing to something nobody in the IPFS network knows about.
* When the users navigates to the auth lobby to permission another app, the auth lobby tries to load the private filesystem, but will fail to do so because it (potentially) tries to fetch a CID that doesn't exist on the IPFS network.

Therefore, we should make webnative more resilient against missing CIDs/malformed file headers or more generally resilient against any kind of file system corruption.

expede:
@matheus23

> Not exactly sure what you mean by threat model

Threat model means "in what way do you we expect people to abuse this?" For example, we're not going to worry about "what if someone has physical access to your phone, unlocked, and all of your passwords". We consider that already game over. Our threat model definitely DOES include "what if you load a malicious application that can steal exportable keys?". Threat model also extends to "what if there's a client that's broken in ways X, Y, and Z".

So, my question translates to roughly "which bits are you concerned about that we actually care about handling?"

gotjoshua:
curious if there are any updates here? (both to the threat model discussion and the warming up issue)

PR #211, opened 2021-04-29
Test the MMPT on race conditions

matheus23:
Note: This test is random. It doesn't _always_ fail. From 1000 added hashes, sometimes it's missing 2, sometimes ~35, sometimes 0. If you want to reproduce, **rerun the test multiple times**

Also, sorry for pulling in the IPFS dependency (only as `devDependency` though). Should I try to get the test running in puppeteer or should we do more of these node-based tests? Especially in the future, given Daniel's work on DI and Brian's work on running tests in node.

### DIFF
--- a/src/fs/protocol/private/mmpt.ts
+++ b/src/fs/protocol/private/mmpt.ts
@@ -2,6 +2,7 @@ import { AddResult, CID } from "../../../ipfs"
 import * as basic from '../basic'
 import * as link from '../../link'
 import { Puttable, SimpleLinks } from '../../types'
+import fs from 'fs'
 
 const nibbles = { "0": true, "1": true, "2": true, "3": true, "4": true, "5": true, "6": true, "7": true,
                   "8": true, "9": true, "a": true, "b": true, "c": true, "d": true, "e": true, "f": true,
@@ -11,6 +12,14 @@ const isNibble = (str: string): boolean => nibbles[str] === true
 type Member = {
   name: string
   cid: string
+}
+
+const filename = `logs-${Date.now()}.log`
+const file = fs.openSync(filename, 'a')
+
+const log = (txt: string) => {
+  // fs.appendFileSync(file, txt, 'utf8')
+  fs.appendFileSync(file, `${txt}\n`, 'utf8')
 }
 
 /**
@@ -26,6 +35,11 @@ export default class MMPT implements Puttable {
   constructor(links: SimpleLinks) {
     this.links = links
     this.children = {}
+  }
+
+  static break(num: number): void {
+    log(`\n----Break ${num}----\n`)
+    return
   }
   
   static create(): MMPT {
@@ -69,7 +83,7 @@ export default class MMPT implements Puttable {
       await this.putAndUpdateLink(nextNameOrSib, nextTree)
     }
 
-    // if one other child with first char of name, then put both into a child tree
+    // if one other child with first char of name, then put both into a new child tree
     else{
       const newTree = this.addEmptyChild(name[0])
       const nextCID = this.links[nextNameOrSib].cid
@@ -84,6 +98,18 @@ export default class MMPT implements Puttable {
 
   async putAndUpdateLink(name: string, child: MMPT): Promise<void> {
     const { cid, size } = await child.putDetailed()
+    log(
+      `
+\n
+----
+name: ${name}
+cid: ${cid}
+links: 
+${Object.values(child.links).map(l => `- ${l.name}: ${l.cid}\n`)}
+----
+\n
+      `
+    )
     this.links[name] = link.make(name, cid, false, size)
   }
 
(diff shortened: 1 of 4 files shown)


matheus23:
Run the test via `yarn test mmpt`

---

If you change a couple of lines in the `MMPT.add()` function definition's last branch, you can see that its condition is invalidated:

```ts
    // if one other child with first char of name, then put both into a child tree
    else{
      const checkRace = () => {
        const nextNameOrSib2 = this.nextTreeOrSiblingName(name)
        if (nextNameOrSib2 != null) {
          console.log("There's a race");
          return;
        }
        if (nextNameOrSib.length === 1) {
          console.log("There's a race");
          return;
        }
      }

      const newTree = this.addEmptyChild(name[0])
      const nextCID = this.links[nextNameOrSib].cid
      this.removeChild(nextNameOrSib)
      await Promise.all([
        newTree.add(name.slice(1), value),
        newTree.add(nextNameOrSib.slice(1), nextCID)
      ])
      checkRace()
      await this.putAndUpdateLink(name[0], newTree)
    }
```

matheus23:
> If you change a couple of lines in the MMPT.add() function definition's last branch, you can see that its condition is invalidated:

Ignore this. I messed up the condition. `nextNameOrSib2 != null` should've been `nextNameOrSib2 == null`.

matheus23:
Closes #207

src/fs/protocol/private/mmpt.ts
```diff
@@ -47,43 +48,53 @@ export default class MMPT implements Puttable {
...
+    const cidBefore = this.links[name]?.cid
+    const { cid, size } = await this.children[name].putDetailed()
+    const cidNow = this.links[name]?.cid
+
+    if (cidBefore != cidNow) {
+      // If there are changes in-between, we have to
+      // re-try updating. Otherwise we can overwrite changes that
+      // a concurrent update made
+      return await this.putAndUpdateChildLink(name)
+    }
```
icidasset, line 96: Trying to understand the MMPT issue.
If the link has updated while updating it, what does that mean exactly?

icidasset (approved):
Had a question, looks good 👍

src/fs/protocol/private/mmpt.ts
matheus23: Ok, here's the best explanation I can give:

This is a race condition, so there's something going wrong when we have two entities concurrently writing to the MMPT. That's possible in javascript due to async-await: As soon as one "execution branch" (not sure how to call this) `await`s some system resource (e.g. network IO or indexedDB IO due to js-ipfs), another branch will continue its execution.
In rare cases, it's possible that two branches execute the same code path, but with slightly different data.

This means that it's possible that two `MMPT.add` calls try to update the same child tree, but try to do that with different data, because they were called for different reasons (different parameters in `MMPT.add()`).

Now, what *can* happen, is that both execute this line:
```js
const { cid, size } = await child.putDetailed()
```
And once the first one enters the `await`, the second one hits the same `await` call. And after the first one finishes, it'll run this line:
```js
this.links[name] = link.make(name, cid, false, size)
```
(and continue with something else until the next `await`)

And then the second branch *also* set's the same link with the same name to another CID. However, since both of these calls happened concurrently, the second time the link was set, it didn't include changes from the first time the link was set, thus dropping one or more branches from the actual DAG that'll be stored in IPFS.

---

The fix for this is to detect a race and the loser of the race will have to run again.

Also, note that in the fix we're using `this.children[name]` instead of the `child` parameter.
This is because `this.children[name]` will always give us the most recent *actual* child of the MMPT. The MMPT actually exists both in IPFS as a tree of CIDs *and* partially as a cached in-memory tree structure with the entries to `this.children` and `this.link` respectively.
During testing, Daniel and I noticed that the in-memory version would never drop anything that was `.add()`ed, but what's behind the `link`s *would*. That's because the in-memory version will always be modified synchronously.

---

Hope this clarifies the issue a bit :)

icidasset: Awesome, thanks!

> This means that it's possible that two MMPT.add calls try to update the same child tree, but try to do that with different data, because they were called for different reasons (different parameters in MMPT.add()).

I was primarily thinking around this, why it would update the same child tree. But I guess that's unavoidable? (I don't entirely remember how the MMPT works in relation to the other pieces). Anyhow, your solution works out perfectly 👍

matheus23: Well, the MMPT is append-only, and the keys used are SHA256 hex-encoded hashes, so it's also super unlikely that two things try to update the same child tree intuitively at first.

However, you have to remember that the IPFS DAG that encodes the tree is immutable. So anytime, say, a 4th-level child MMPT gets another node added, it's parent and grandparent and grand-grand-parent have to be updated, since all the CIDs change.

That's when two calls can suddenly clash and race each other.

Also, interestingly, note here that this is only true for the `this.links` property, but not for `this.children`.

PR #214, opened 2021-05-03
Add a rule to disallow semicolons

bgins:
## Summary

* [x] Add linter rule to disallow semicolons

We tend to not use semicolons. This makes it official.

## Test plan (required)

Run `npm run lint` and you will see some warnings for some errant semicolons, mostly in tests.

## After Merge


matheus23:
What does enabling this rule mean for the few cases in which ASI doesn't work?

bgins:
That's a good question. The `never` setting makes some exceptions: https://eslint.org/docs/rules/semi#never. We have one case that looks like their example:

https://github.com/fission-suite/webnative/blob/f1340a00fbae9c89c322d41dccd7506c2c9ceb9c/src/auth.ts#L37

The `never` rule also allows semicolons in old-school for loops:

https://github.com/fission-suite/webnative/blob/f1340a00fbae9c89c322d41dccd7506c2c9ceb9c/src/common/hex.ts#L10

Are there other cases where we want a semicolon that we should try out?

matheus23 (approved):
This answers my question! :)

> Are there other cases where we want a semicolon that we should try out?

I remember this case from the dashboard source code:

```js
const unhighlight = async event => {
  event.preventDefault()
  this.classList.remove("dropping")
};

["dragleave", "drop"].map(ev => this.addEventListener(ev, unhighlight));
["dragenter", "dragover"].map(ev => this.addEventListener(ev, highlight));
```

But I they've got that covered:

> except to disambiguate statements beginning with `[`, `(`, `/`, `+`, or `-`

So :+1: :+1:

matheus23 (approved):

matheus23:
I think we can just merge this, right? I'll push the button for now.

bgins:
Yep, we can. Thanks! :pray:

PR #215, opened 2021-05-03
Raw ucans

dholms:
## Problem
A user can only request wnfs-related UCANs from the auth lobby

## Solution
Allow a user to request additional `raw` ucans.
They can provide the following fields:
- rsc
- ptc
- exp

The rest can be inferred by the auth lobby


dholms:
pre-rebase checkpoint: 680d618a20146ebad51d26d01fb65132b1fcc140

rollup.config.ts
```diff
@@ -76,30 +76,29 @@ const configUMDMinified = {
...
-  },
+  output: [
+    {
+      file: pkg.main,
+      format: 'cjs',
+      sourcemap: true
+    },
+    {
+      file: pkg.module,
+      format: 'es',
```
dholms, line 95: When packaging up `webnative-filecoin`, it uses the `module` entry not the `main` entry. So I'm building both of these for node.

I'm still hopelessly confused by js module formats lol. would it be wrong for us to just ditch either CJS or ES entirely? UMD for browser build & either cjs/es for node? Would also speed up build times.

Related node: I don't think anything uses or minified build right now. We could either remove it from the build or point `browser` (in `pacakge.json` to the minified build instead of the full umd

src/ucan.ts
```diff
@@ -138,23 +143,28 @@ export async function build({
...
-    const resource = Array.from(Object.entries(rsc))[0]
-    const key = resource[0] + ":" + (
-      resource[0] === WNFS_PREFIX
-        ? resource[1].replace(/^\/+/, "")
-        : resource[1]
-    )
+/**
+ * Creates the label for a given resource in the UCAN dictionary
+ */
+export function resourceLabel(rsc: Resource): string {
```
dholms, line 157: cosign ucans are labeled as: `cosign:did:key:cosigingkey`

rollup.config.ts
matheus23: I think `browser` in the package.json should point to the minified build :+1: 

> I'm still hopelessly confused by js module formats lol. would it be wrong for us to just ditch either CJS or ES entirely? UMD for browser build & either cjs/es for node? Would also speed up build times.

We should have both cjs and mjs. The problem here is that there are users that use webnative with ES modules and users that use webnative with commonjs modules.
I think we should reconsider that once node supports esnext natively and by default & the old node versions that don't support it are out of LTS. That'll take a while, if it happens at all, though :D

matheus23 (approved):

src/ucan/index.ts
```diff
@@ -0,0 +1,3 @@
+export * from './types'
+export * from './token'
+export * as dictionary from './dictionary'
```
dholms, line 3: we export helper methods for looking up ucans in dictionary, but not actual store itself (which is in `ucan/store.ts`)

src/did.ts
```diff
@@ -54,6 +55,18 @@ export async function root(
   throw new Error("Could not locate user DID in DNS.")
 }
 
+/**
+ * Get a user's own root write-key DID.
+ * Stored at `_did.${username}.${endpoints.user}`
+ */
+export async function ownRoot(): Promise<string> {
+  const username = await storage.getItem(USERNAME_STORAGE_KEY)
```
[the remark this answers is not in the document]
icidasset: There's also `await common.authenticatedUsername()` as a helper to get the username (https://github.com/fission-suite/webnative/blob/main/src/common/index.ts#L21).

Regarding getting the actual root DID, you could pick a random UCAN from the store and then get the root issuer with the `ucan.rootIssuer()` function. Feels a bit messy, but still ok I guess, better than making an HTTP request? 🤷

src/ucan/dictionary.ts
```diff
@@ -0,0 +1,111 @@
...
+ */
+export function filesystemPrefix(username: string): string {
+  // const host = `${username}.${setup.endpoints.user}`
+  // TODO: Waiting on API change.
+  //       Should be `${WNFS_PREFIX}:${host}/`
+  return WNFS_PREFIX + ":"
+}
+
+/**
+ * Look up a UCAN for a platform app.
```
icidasset, line 65: Think this needs a new description.

icidasset (approved):
Few small notes, looks great!

src/did.ts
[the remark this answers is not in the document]
dholms: Yeah I thought about that :thinking: but also a user could be authenticated wtih no UCANs.

Hmmm maybe I"ll try the UCAN approach first and then fall back to the HTTP request

PR #219, opened 2021-05-05
Load private files in root tree

dholms:
## Problem
The private side of the sdk expects all entry points to be directories. But we can now permission files as entry points

## Solution
Allow both files & directories as entry points on the private side

### DIFF
--- a/src/fs/filesystem.ts
+++ b/src/fs/filesystem.ts
@@ -345,7 +345,7 @@ export class FileSystem {
       : "query"
 
     if (!this.localOnly) {
-      const proof = await ucanInternal.lookupFilesystemUcan({ directory: parts })
+      const proof = await ucanInternal.lookupFilesystemUcan(path)
       const decodedProof = proof && ucan.decode(proof)
 
       if (!proof || !decodedProof || ucan.isExpired(decodedProof) || !decodedProof.signature) {
--- a/src/fs/filesystem.ts
+++ b/src/fs/filesystem.ts
@@ -335,7 +335,6 @@ export class FileSystem {
     isMutation: boolean,
     fn: (node: UnixTree | File, relPath: Path) => Promise<a>
   ): Promise<a> {
-    console.log('path: ', path)
     const parts = pathing.unwrap(path)
     const head = parts[0]
     const relPath = parts.slice(1)
--- a/src/fs/filesystem.ts
+++ b/src/fs/filesystem.ts
@@ -314,14 +314,15 @@ export class FileSystem {
     const head = parts[0]
     const relPath = parts.slice(1)
 
+    const operation = isMutation
+      ? "make changes to"
+      : "query"
+
     if (!this.localOnly) {
       const proof = await ucanInternal.lookupFilesystemUcan({ directory: parts })
       const decodedProof = proof && ucan.decode(proof)
 
       if (!proof || !decodedProof || ucan.isExpired(decodedProof) || !decodedProof.signature) {
-        const operation = isMutation
-          ? "make changes to"
-          : "query"
         throw new NoPermissionError(`I don't have the necessary permissions to ${operation} the file system at "${pathing.toPosix(path)}"`)
       }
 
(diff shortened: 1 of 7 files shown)


src/fs/root/tree.ts
```diff
@@ -308,8 +310,15 @@ function loadPrivateTrees(
         const bareNameFilter = await findBareNameFilter(map, path)
         if (!bareNameFilter) throw new Error(`Was trying to load the PrivateTree for the path \`${path}\`, but couldn't find the bare name filter for it.`)
 
-        privateTree = await PrivateTree.fromBareNameFilter(mmpt, bareNameFilter, key)
-
+        const maybeInfo = await protocol.priv.getLatestByBareNameFilter(mmpt, bareNameFilter, key)
+        if (maybeInfo === null) throw new Error(`Could not find content in filesystem for path ${path}`)
+        if(check.isPrivateTreeInfo(maybeInfo)) {
```
icidasset, line 315: Is there a reason you prefer this over just using `pathing.isFile(path)`?

src/fs/root/tree.ts
```diff
@@ -256,44 +260,47 @@ type PathKey = { path: DistinctivePath, key: string }
...
-  if (!tree.exists(relativePath)) {
-    if (pathing.isDirectory(path)) await tree.mkdir(relativePath)
-    else await tree.add(relativePath, "")
+  const relativePath = unwrappedPath.slice(pathing.unwrap(nodePath).length)
+  if(PrivateFile.instanceOf(node)) {
+    return relativePath.length === 0 ? node.header.bareNameFilter : null
+  }
+  if (!node.exists(relativePath)) {
+    if (pathing.isDirectory(path)) await node.mkdir(relativePath)
+    else await node.add(relativePath, "")
```
icidasset, line 281: What's the reason we're adding this here? 👀

src/fs/root/tree.ts
dholms: :monocle_face: 

nope I don't think so. I'm just not super used to the new pathing tools so didn't think of it :ok_hand:

src/fs/root/tree.ts
dholms:
```ts
if (!node.exists(relativePath)) {
    if (pathing.isDirectory(path)) await node.mkdir(relativePath)
    else await node.add(relativePath, "")
```
this part already existed before the PR
```ts
if(PrivateFile.instanceOf(node)) {
    return relativePath.length === 0 ? node.header.bareNameFilter : null
```
this part I added and is to prevent trying to access relative paths off of a file (which would not exist)

src/fs/filesystem.ts
```diff
@@ -304,23 +330,25 @@ export class FileSystem {
...
     isMutation: boolean,
-    fn: (tree: UnixTree, relPath: Path) => Promise<a>
+    fn: (node: UnixTree | File, relPath: Path) => Promise<a>
   ): Promise<a> {
+    console.log('path: ', path)
     const parts = pathing.unwrap(path)
     const head = parts[0]
     const relPath = parts.slice(1)
 
+    const operation = isMutation
```
dholms, line 343: nice :ok_hand:

src/fs/filesystem.ts
```diff
@@ -181,17 +183,25 @@ export class FileSystem {
 
   async ls(path: DirectoryPath): Promise<BaseLinks> {
     if (pathing.isFile(path)) throw new Error("`ls` only accepts directory paths")
-    return this.runOnTree(path, false, (tree, relPath) => {
-      return tree.ls(relPath)
+    return this.runOnNode(path, false, (node, relPath) => {
+      if (typeCheck.isFile(node)) {
```
dholms, line 187: ah right... paths let us check the node type ahead of time (as long as everything is structured correctly). So wtih proper pathing, we should never hit this error

loving the new path primitives :ok_hand:

src/fs/root/tree.ts
icidasset: Ah gotcha 👍 
My bad, I should've added comments to this 😅

PR #220, opened 2021-05-05
Implement new flow for exchanging secrets with the auth lobby

matheus23:
Depends on https://github.com/fission-suite/auth-lobby/pull/72

Creates a small iframe to exchange the linking secrets with the auth lobby.
The auth lobby can signal that it supports this flow by providing the query parameter `authoised=via-postmessage` after redirecting back from `redirectToLobby`.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -370,6 +370,7 @@ async function importClassifiedInfo(
 async function getClassifiedViaPostMessage(): Promise<string> {
   const iframe: HTMLIFrameElement = await new Promise((resolve, reject) => {
     const iframe = document.createElement("iframe")
+    iframe.id = "webnative-secret-exchange"
     iframe.style.width = "0"
     iframe.style.height = "0"
     iframe.style.border = "none"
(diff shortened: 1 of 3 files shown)


matheus23:
This PR can be tested on production at:
* https://philipps-auth-test.fission.app/ The lobby part
* https://dashboard-auth-test.fission.app/ An app with the new webnative change to support this

matheus23:
This also addresses #207

src/index.ts
```diff
@@ -353,6 +367,48 @@ async function importClassifiedInfo(
   await ucan.store(ucans)
 }
 
+async function getClassifiedViaPostMessage(): Promise<string> {
+  const iframe: HTMLIFrameElement = await new Promise((resolve, reject) => {
+    const iframe = document.createElement("iframe")
```
icidasset, line 372: Maybe add an `id` attribute so people can recognise what the iframe is for.

icidasset (approved):
One small note, looks good otherwise 🎉

PR #227, opened 2021-05-07
Browser detection should include workers

icidasset:
The browser detection was assuming the `window` object which is not available in web/service workers. This PR replaces that with a check for the `location` object, which is both available on main ui thread (window object) and [worker contexts](https://developer.mozilla.org/en-US/docs/Web/API/WorkerGlobalScope).

Published this as `0.24.1-alpha-2` if you want to test it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,15 @@
 # Changelog
 
+### v0.24.2
+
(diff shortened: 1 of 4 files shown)


dholms (approved):
Looks great :tada: :tada:

PR #228, opened 2021-05-07
Fix DI issues

dholms:
## Problem
- merging dependencies is not a deep merge
- `crypto` & `storage` modules get set off the original implementation and not updated on `setDependencies

## Solution
- deep merge dependencies on set
- make sure `crypto` & `storage` modules use the currently set implementation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+### v0.24.1
+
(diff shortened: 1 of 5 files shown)


bgins (approved):
Looks good! I tried these changes out with the initial Jest DI implementation, and they are working there. :tada:

Issue #201, opened 2021-04-16
WNFS v1 Test Suite


expede:

bgins:
Updating on the status of our test suites.

## Completed

### Unit tests and prop checks
- Most of `src/common` is tested with unit tests and prop checks
- A few unit tests of mmpts, paths, and usernames

### Integration tests
- Filesystem tests for appPaths, concurrency, and creating parent directories automatically
- UCAN tests and UCAN permissions for private paths
- Did tests verifying UCAN signatures and converting public keys to DIDs

## Next steps
- Fix broken tests
  + `src/common/cid-log.test.ts` on `bgins/testing`
  + `tests/fs/concurrency.test.ts`
  + warning from `tests/fs/tree.test.ts` that looks like it may be related to DI
- Dependency injection opens up some opportunities using WebCrypto and storage in tests
  + Some tests in `src/common/` could replace mocking with dependency injection
- Optimize `mmpt.test.ts`
  + Currently takes 2 minutes or more to run, and it sometimes times out
- Add more unit tests and prop checks in `src/fs/`

expede:
@bgins since these seem bucket-able, is it worth making this Issue an epic, and breaking those out into their own and linking them back to this one? Or do they all effect each other and can't be cleanly separated?

bgins:
@expede I think we could go epic with this one. They should all be fairly independent.

Would we add these as issues to the `webnative` repo? Perhaps with a "testing" label?

expede:
Okay great! Yeah, a testing label, and reference them from this issue so we can track them from here.

Thanks!!

bgins:
Opened issue https://github.com/fission-suite/webnative/issues/221 to track optimization of `mmpt.test.ts`.

bgins:
Alright, opened a few more issues for the next steps mentioned above.

- Update common tests with DI https://github.com/fission-suite/webnative/issues/223
- Fix broken tests https://github.com/fission-suite/webnative/issues/224
- Add fs unit tests and prop checks https://github.com/fission-suite/webnative/issues/225

bgins:
Pull request https://github.com/fission-suite/webnative/pull/232 resolves #223, where we wanted to update the `src/common/` tests with DI. It also takes care of the broken `cid-log` test mentioned in #224.

The next step here is adding the `fs` tests mentioned in #225.

icidasset:
Closing in favour of upcoming rs-wnfs integration.

PR #232, opened 2021-05-12
Add in-memory storage DI, node sha256 DI, and common tests

### LINKED ISSUE #223: Update common tests with DI
Some of the tests in `src/common/` rely on mocking. In some cases, we may be able to replace mocking with dependency injection.

We can update:

- `cid-log.test.ts` on `bgins/testing` which mocks `localforage`

We cannot update:

- `api.test.ts` because it uses `fetch` and we do not currently have DI for network requests

bgins:
## Summary

* [x] Add in-memory storage DI implementation
* [x] Add node WebCrypto sha256 implementation
* [x] Add cid-log tests
* [x] Add async race test
* [x] Add identifiers tests

This PR adds an in-memory storage implementation for use in Jest. The implementation mostly matches the `localforage` API, though it does not currently restrict to the types that `localforage` can store.

We are also adding tests for `cid-log.ts`, `async.ts`, and `identifiers.ts` in the `src/common/` directory. Note that the `cid-log` tests use the new storage DI implementation and the `identifiers` tests use the sha256 implementation.

## Test plan (required)

Run `npm run test:unit`. All tests should pass.

## Related issues

Closes #223 
Part of https://github.com/fission-suite/webnative/issues/201

## After Merge


.eslintrc.js
```diff
@@ -17,6 +17,7 @@ module.exports = {
       },
     }],
     "@typescript-eslint/no-use-before-define": ["off"],
-    "@typescript-eslint/semi": ["error", "never"]
+    "@typescript-eslint/semi": ["error", "never"],
+    "@typescript-eslint/ban-ts-comment": 1
```
matheus23, line 21: What does the value `1` mean here? Online I found that you can set this to `"off"` and that turns it off?

src/common/cid-log.test.ts
```diff
@@ -0,0 +1,130 @@
+import * as fc from 'fast-check'
+import CID from 'cids'
+import multihash from 'multihashing-async'
+import * as cidLog from './cid-log'
+import * as storage from '../storage'
+
+async function generateCids(data: Uint8Array[]): Promise<string[]> {
+  const promisedCids = data.map(async bytes => {
+    const mhash = await multihash(bytes, 'sha2-256')
```
matheus23, line 9: Oh my god. I needed this in one of my tests haha.

src/common/async.test.ts
```diff
@@ -0,0 +1,17 @@
+import * as fc from 'fast-check'
+import { race } from './async'
```
matheus23, line 2: Wait. Why do we have our own implementation for [`Promise.race`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/race)?

No action required immediately, but I think we should switch that out for the native variant. No need to test this ourselves then :+1:

matheus23 (approved):
:pray: 
:raised_hands: 
:+1:

.eslintrc.js
bgins: `0` means off, `1` means warn, and `2` means error. We might be able to `@typescript-eslint/eslint-plugin` to a newer version to get something a bit more explicit. The newer interface is `bool` or `"allow-with-description"` that makes you write down why.

src/common/cid-log.test.ts
bgins: Yeah! That's been super helpful for getting some real CIDs :smile:

src/common/async.test.ts
bgins: Great point, I was also noticing that Jest coverage doesn't think we are actually calling `promise.then(resolve, reject)` in the implementation. (I tested it with logging and we are, but it's not ideal that it doesn't show up in coverage.)

walkah: TIL! I didn't realize Promises has `.race` built in.

walkah (approved):
30 rock college - DOxlkyuDxTrna

*test coverage intensifies*

PR #234, opened 2021-05-18
Complete the Jest DI with crypto

matheus23:
This is preparation work for making it possible to write more sophisticated test, making test-driven-development for internal (non-browser-bundle-exposed) APIs possible.


src/common/version.ts
```diff
@@ -1 +1 @@
-export const VERSION = "0.24.0"
+export const VERSION = "0.24.2"
```
matheus23, line 1: No idea why this was stale :/

Maybe we should .gitignore this?

src/setup/jest.ts
```diff
@@ -10,40 +15,306 @@ import { setDependencies } from './dependencies'
...
+  async getSymmKey(keyName: string, cfg?: Partial<Config>): Promise<CryptoKey> {
+    const mergedCfg = config.merge(this.cfg, cfg)
+    const maybeKey = this.store[keyName]
+    if(maybeKey !== null) {
+      return maybeKey
+    }
+    const key = await aes.makeKey(config.symmKeyOpts(mergedCfg))
+    this.store[keyName] = key
+    return key
+  }
```
matheus23, line 128: I duplicate some code here [from keystore-idb](https://github.com/fission-suite/keystore-idb/blob/74d2a96429dde2b86d3daa3597b4ef35334db937/src/rsa/keystore.ts).

We should probably eventually make it possible to use our dependencies in node. However, doing that for keystore-idb could be questionable, given it has "idb" in its name :smile: 

Any way, this feels kind of quick-and-dirty right now. Let me know if you have any other ideas.

src/index.ts
```diff
@@ -256,6 +256,7 @@ export * as ipfs from './ipfs'
 export * as keystore from './keystore'
 export * as machinery from './common'
 export * as crypto from './crypto'
+export * as cbor from 'cborg'
```
matheus23, line 259: I'm not happy about exposing this, but I need this to test the browser build in `encoding.test.ts`

package.json
```diff
@@ -36,7 +36,7 @@
     "prebuild": "rimraf dist && node scripts/gen-version.js",
     "build": "tsc && rollup -c rollup.config.ts",
     "start": "tsc -w",
-    "test": "jest",
+    "test": "jest --forceExit",
```
walkah, line 39: Hrm. This smells like there's something wrong with the tests... async operations not resolving?

src/fs/protocol/private/mmpt.test.ts
```diff
@@ -1,8 +1,9 @@
 import crypto from "crypto"
-import Ipfs, { IPFS } from "ipfs"
+import { IPFS } from "ipfs-core"
```
walkah, line 2: Is this a size thing? (full IPFS includes more code than we use, etc?)

walkah (approved):
This is great! thanks for your continued efforts getting this working!

I Appreciate You - KD22dF8apIKLSapKSC

FYI: `bcrypto` isn't currently building for me locally in the nix shell on macOS. I'm going to poke around, but it seems to work in CI so may be my environment.

src/fs/protocol/private/mmpt.test.ts
matheus23: Yep, exactly. It's just much less dependencies.
I think the difference is that ipfs also contains the CLI and an http-server, which ipfs-core does not.

package.json
matheus23: Yep. I tried to track it down for a while, but gave up. From my testing it seems like in some edge-cases it's possible that ipfs leaves a child process and some sockets hanging after `await ipfs.stop()`. Not sure what's up with that, but after fighting with it for a while I just gave up.
The IPFS we're using in the tests is basically in-memory only. It doesn't write to disk. So killing it shouldn't be an issue.

walkah: Cool. yeah, something we can track down and fix as a follow up later, imo.

matheus23:
> FYI: `bcrypto` isn't currently building for me locally in the nix shell on macOS. I'm going to poke around, but it seems to work in CI so may be my environment.

Yeah. Weird. `bcrypto` is building fine on my end (including in the nix shell).

bgins (approved):
Looks great! Thanks for implementing these Jest DI crypto functions :tada: :pray:

tests/encoding.test.ts
```diff
@@ -0,0 +1,43 @@
...
+        expect(decoded).toEqual(message)
+    })
+
+    it("works in the browser with encryption in between", async () => {
+        await loadWebnativePage()
+
+        async function runRoundTrip(message) {
+            const keyStr = await webnative.crypto.aes.genKeyStr()
+
+            const encoded = webnative.cbor.encode(message)
```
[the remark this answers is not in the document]
matheus23: Did you `yarn build` before you `yarn test`ed?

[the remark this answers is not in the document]
bgins: I did not :facepalm: and that was it. After building, this test and all the others are passing for me :100:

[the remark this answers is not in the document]
matheus23: Admittedly, it's really confusing. The puppeteer html file fixture uses the index.umd.min.js file, so unless one `yarn build`s before running the tests, it'll use an outdated webnative, but only for the puppeteer stuff.

PR #236, opened 2021-05-19
Fix tsc output

dholms:
## Problem
tsc is outputting built files to `dist/src/` instead of `dist`. This breaks type references such as `types` in `package.json`

## Solution
This is because some code in `src` (specifically `setup/jest.ts`) references some code in `tests/`

So we exclude `src/setup/jest.ts` in `tsconfig.ts`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-### v0.25.1
+### v0.25.2
 
 Typescript output to `dist/` instead of `dist/src/`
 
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Changelog
 
+### v0.25.1
```
walkah, line 3: this should be v0.25.2, yeah?

walkah (approved):
Thanks for cleaning this up! 🙏 🚢

CHANGELOG.md
dholms: yup! thanks!

PR #243, opened 2021-05-27
Use native ipfs types and remove hand-written ones

matheus23:
## Summary

This will make webnative use the typescript types that were now developed by ipfs (https://github.com/ipfs/js-ipfs/issues/2945).
This in turn makes the types more accurate, as it seems they've become somewhat outdated. E.g. see [this line](https://github.com/fission-suite/dashboard/blob/8919141129a209b2cd3151a49d66d4b6fa0a3a80/src/Javascript/index.ts#L490) in the dashboard for one of the cases we've gotten it slightly wrong.

## Test plan (required)

Unfortunately it's not a pure refactor: On the public side, I've had to transform the `FileContent` type into `ImportCandidate` (from ipfs). This makes sense, because some of the `FileContent` types wouldn't be supported to be added on the public filesystem side.
This changes behavior, but I decided that in this case it makes sense to fix the implementation and adjust it to the types instead of adjusting the types. This is something I'm not 100% sure about, so please discuss if you have some ideas.

Since I've changed behavior, we should test some apps, including drive, the dashboard & maybe flatmate (or something else using webnative-elm).


matheus23:
I've updated dependencies in this PR. I think merging this will also close most of the dependabot PRs.

package.json
```diff
@@ -58,64 +58,65 @@
...
   "dependencies": {
     "base58-universal": "https://github.com/digitalbazaar/base58-universal#de970560f005de0f7054723c35ef6e0ff4b328b7",
     "borc": "^3.0.0",
     "buffer": "^6.0.3",
-    "cids": "^1.1.5",
-    "fission-bloom-filters": "1.6.0",
-    "ipfs-core": "^0.6.1",
+    "cids": "^1.1.6",
+    "fission-bloom-filters": "1.7.0",
+    "ipfs-core": "^0.7.0",
```
icidasset, line 111: Shouldn't this be in `devDependencies`?

src/ipfs/util.ts
```diff
@@ -1,19 +1,19 @@
-import dagPB from 'ipld-dag-pb'
-import { DAGNode, RawDAGNode, DAGLink, RawDAGLink } from './types'
+import dagPB, { DAGLink, DAGNode } from 'ipld-dag-pb'
+import type { GetResult } from 'ipfs-core-types/src/dag'
+import { CID } from 'ipfs-message-port-client/src/block'
```
icidasset, line 3: Is this the same `CID` type as from the `cids` library?

icidasset (approved):
Few questions, looks great otherwise!

package.json
matheus23: YES! Fuuuuu...

src/ipfs/util.ts
matheus23: Nope. Super annoying that it isn't. Not sure what's up with that in the js-ipfs ecosystems right now :thinking:

PR #247, opened 2021-06-01
fix #241

### LINKED ISSUE #241: Ucan.build should fail if missing parameters
# Summary

## Problem

Creating a UCAN without an `aud` field works fine, but fails to generate a valid UCAN.

Reported by @fuck-capitalism

### Impact

It's confusing for users!

## Solution

Fail with a helpful message when required fields are missing

fuck-capitalism:


icidasset:
Thanks for the PR @fuck-capitalism !

We've decided that we want to be explicit about what the `audience` and `issuer` are. We made a mistake somewhere along the road in which we made `issuer` optional, both `audience` and `issuer` should be required. 

I think here we want to throw an error if either `audience` or `issuer` is undefined, as opposed to setting a default value.

icidasset:
UCAN creation has moved to https://github.com/fission-suite/ucan

PR #251, opened 2021-06-11
Add Filesystem API tests

bgins:
## Summary

This PR adds tests of filesystem operations on the private and public filesystem.

* [x] Add public filesystem tests
* [x] Add private filesystem tests

Our goal is to test external filesystem API calls with prop checks that:
- Test `write`, `read`, `mkdir`, `exists`, `mv`, `rm`, `ls`
- Use random (but biased) directory paths in the `public` and `private` directories
- Use random (but biased) `FileContent` for `write` and `read` operations

**Note:** `fast-check` biases towards a small number of simple examples, but generates random samples beyond that.

Our `FileContent` type is `Record<string, unknown> | FileContentRaw | Blob | string | number | boolean` and these tests cover:
- All types except `Blob` and `Record<string, unknown>` in the public filesystem
- All types except `Blob` in the public filesystem

The current number of runs per test is set at `10` because the time to test gets very long with a larger number of runs. The plan is to add an environment variable to change the number of runs for different environments (like the GitHub action for example).

## Test plan (required)

Run `npm run test` to run the test suites. The new test suites should pass.


tests/fs/api.private.test.ts
```diff
@@ -0,0 +1,284 @@
...
+
+          expect(await fs.exists(dirpath)).toEqual(false)
+        })
+    )
+  })
+
+  it('writes files to a directory', async () => {
+    const fs = await emptyFilesystem()
+    const dirpath = path.directory('private', 'testDir')
+    await fs.mkdir(dirpath)
```
icidasset, line 155: You could even make the `mkdir` optional here, so that it also tests if `fs.write` creates the parent dir.

icidasset (approved):

tests/fs/api.private.test.ts
```diff
@@ -0,0 +1,284 @@
...
+import { createInMemoryIPFS } from '../helpers/in-memory-ipfs'
+import FileSystem from '../../src/fs/filesystem'
+
+import * as ipfsConfig from '../../src/ipfs'
+import * as path from '../../src/path'
+import * as crypto from '../../src/crypto'
+
+
+let ipfs: IPFS | null = null
+let rootKey: string
```
matheus23, line 13: `rootKey` doesn't have to be global here, right? It's only used in `emptyFilesystem`, but never read.

tests/fs/api.private.test.ts
```diff
@@ -0,0 +1,284 @@
...
+
+fc.configureGlobal({ numRuns: 10 })
+
+describe('the filesystem api', () => {
+  it('writes files', async () => {
+    const fs = await emptyFilesystem()
+
+    await fc.assert(
+      fc.asyncProperty(
+        fc.tuple(pathSegment(), fileContent()), async data => {
```
matheus23, line 33: `data` seems like a really generic function name. Also, let's say (hypothetically) you'd at some point change an `fc.tuple` call from 3 arguments to 2 arguments. It might be pretty easy to miss an undefined `data[2]` access.

What do you think about using pattern matching here?
```ts
 fs.tuple(pathSegment(), fileContent()), async ([pathSegment, [val, type]]) => { // probably needs a type signature
    // ...
}
```

I don't really know what the types are, but it should be something like `[pathSegment, [val, type]]: [string[], [<sth>, <sth>]]`.

tests/fs/api.private.test.ts
```diff
@@ -0,0 +1,284 @@
...
+      }
+    },
+    rootKey
+  })
+}
+
+/* Paths */
+
+const pathSegment = () => {
+  return fc.hexaString({ minLength: 1, maxLength: 20 })
```
matheus23, line 233: This is good, but I'd like it if we had a specified set of "what characters are good" for paths.
That needs some discussion outside of this PR probably. I.e. do we want to allow basically all utf-8 strings, except containing `/`? Do we want to interoperate with existing filesystems, e.g. look valid path strings are for windows/unix/mac? (Probably, not, given `AUX.txt` is not a valid file in windows).

Just noting this here. (Also, what would happen if we were generating arbitrary utf-8 strings here that don't contain slashes? So something like `fc.fullUnicodeString({ minLength: 1, maxLength: 20 }).filter(s => !s.includes("/"))`.)

tests/fs/api.private.test.ts
```diff
@@ -0,0 +1,284 @@
...
+    { arbitrary: simpleContent(), weight: 10 },
+    { arbitrary: rawFileContent(), weight: 1 },
+    { arbitrary: recordContent(), weight: 1 }
+  )
+}
+
+const simpleContent = () => {
+  return fc.frequency(
+    { arbitrary: fc.tuple(fc.json(), fc.constant('string')), weight: 6 },
+    { arbitrary: fc.tuple(fc.string({ minLength: 1 }), fc.constant('string')), weight: 4 },
```
matheus23, line 256: Empty strings should definitely be tested, too, I think.

tests/fs/api.public.test.ts
```diff
@@ -0,0 +1,292 @@
...
+          expect(toExists && !fromExists).toEqual(true)
+        })
+    )
+  })
+})
+
+
+/* Filesystem */
+
+const emptyFilesystem = async () => {
```
matheus23, line 219: Can we move these utilities (they're great!), to a file in `tests/helpers/`?

Same thing with the `pathSegment` and `pathSegmentPair`. Maybe we'll make `privateFileContent` and `publicFileContent`.

(Maybe this is premature abstraction, though, so consider this just an idea)

matheus23 (changes requested):
First off: `fast-check` is *amazing*. I'm so glad that you're introducing fast-check tests to webnative, brian! :)

I'd generally like more usage of `fc.record` instead of `fc.tuple`. With `fc.tuple` you often end up with a very generic-sounding variable name `data` and lots of index-access, which seems a little brittle and harder to read compared to something using records.

tests/fs/api.private.test.ts
bgins: Good call. Removing the `mkdir` from this totally makes sense.

tests/fs/api.private.test.ts
bgins: Yep, that's correct. Cleaning that up.

tests/fs/api.private.test.ts
bgins: Yeah, that's a great point. One other consideration here is should there be some bias towards a set of characters? Hex strings or ASCII don't seem like the right bias, but some limitations might be good because the tests are slow and we don't want to use up our runs with characters that aren't likely to be used.

tests/fs/api.private.test.ts
bgins: I've been playing with this idea and `fc.record`. Both suggestions are great, thanks for calling them out. :pray: Will clean up these tests nicely.

I'm leaning towards tuples with pattern matching over records. The records are nice because they are more constrained, but I'm not quite sure what to with the tuples in `fileContent` and `pathSegmentPair`. It seems like we would need to destructure or index in the test body for these. 

Tuples with pattern matching seems a bit more readable to me. I'll change over to using that, but I'm open to discussing records more.

matheus23: Can we perhaps make `fileContent` (and possible `pathSegmentPair`) defined using `fc.record`, too? (I hope I'm understanding you correctly here)

tests/fs/api.private.test.ts
bgins: Yep, and early iteration of the tests was failing, but likely there was something faulty with the test itself. Changing this to include empty strings.

tests/fs/api.private.test.ts
matheus23: I'm not 100% sure whether matching the frequency of use with the frequency of generation is a good idea. I'd rather try to choose the tested values from a space that's most likely to uncover bugs, right? If we try to match the frequency of use with the frequency of generation, it'll (I think) result in missing the less frequently used edge-cases.

To resolve this issue, we'd need to converge on what we think are the actually allowed values in file paths, though. So again, I don't think we'll move on this issue, yet. Maybe an internal talk post in the forum might be the best next step. :thinking:

tests/fs/api.private.test.ts
bgins: Yes, you are. And it's a good suggestion. 

`pathSegmentPair` is a bit trickier than `fileContent` because it's a set, not a tuple:

```javascript
const pathSegmentPair = () => {
  return fc.set(
    fc.hexaString({ minLength: 4, maxLength: 20 }),
    { minLength: 2, maxLength: 2 }
  )
}
```

but we might be able to get the same effect using `filter` on records. Also, the minimum length of 4 keeps the members of the pair unique in most cases. :thinking:

tests/fs/api.private.test.ts
bgins: Yep, talk post for sure. There's a tradeoff between catching edge cases and being sure things work for the common cases. Biasing towards common cases could definitely mean we miss edge cases, but with the number of runs we can do, we might miss them anyways.

Maybe this is another thing that we could  do is configure per environment. Like sanity checks in the local environment and more expansive tests in a GH action.

Anyways, yes, talk post. :bookmark: :bookmark:

tests/fs/api.private.test.ts
matheus23: Ah right. Yeah let's keep that an `fc.set`.

> It seems like we would need to destructure or index in the test body for these.

Keep in mind that you can also destructure record fields in javascript (unlike in e.g. Elm):
```js
const { pathPair: [first, second] } = { pathPair: ["abc", "def"] }
```

---

I'm fine either way. I see it might be more involved. I'll leave it to your judgement :)

matheus23: Oh, one more option: There's a `.map` for `Arbitrary`s in fast-check. So this would work:

```js
const pathSegmentPair = () => {
  return fc.set(
    fc.hexaString({ minLength: 4, maxLength: 20 }),
    { minLength: 2, maxLength: 2 }
  ).map(([first, second]) => ({ first, second }))
}
```

In case that'd make it any more readable :sweat_smile:

bgins: It's great! Fun to explore the possibilities here and learn more `fast-check` along the way. :smiley:

bgins: I went ahead with records instead of tuples. With a bit of pattern matching, it came out fairly readable.

Thanks again for all the great suggestions on this. :pray:

tests/fs/api.public.test.ts
bgins: I went ahead and did it! Can totally see that these will see more use than in just these tests.

When I was typing the public decoder, I used some type assertions. Like this one for example:

https://github.com/fission-suite/webnative/blob/3105c2ee8a97dcea9a5c5d52f0e8f1578cda372e/tests/helpers/fileContent.ts#L81

Does this seem alright? I wasn't able to use `BufferSource` in the type signature of the decoder, because we end up with a type mismatch in the tests. The problem is that `fs.read` returns `FileContent` that we decode, but the `TextDecoder` expects a `BufferSource`.

matheus23 (approved):
:clap: :tada: :pray:

tests/fs/api.public.test.ts
matheus23: Hmm. Yeah, technically the public side will only ever return `Uint8Array` (which is one of the types in the `FileContent` union), so at the moment it's not precise enough. I want to change that for a webnative API redesign in the future, but casting is just fine for now :+1:

PR #252, opened 2021-06-14
Build system changes

matheus23:


package.json
```diff
@@ -84,44 +86,33 @@
...
     "rimraf": "^3.0.2",
-    "rollup": "^2.50.3",
-    "rollup-plugin-gzip": "^2.5.0",
-    "rollup-plugin-node-polyfills": "^0.2.1",
-    "rollup-plugin-sourcemaps": "^0.6.3",
-    "rollup-plugin-terser": "^7.0.2",
-    "rollup-plugin-typescript2": "^0.30.0",
     "tslib": "^2.2.0",
-    "typedoc": "^0.20.36",
+    "typedoc": "^0.21.0-beta.2",
```
icidasset, line 93: Any significant changes in the beta?

src/package.json
```diff
@@ -0,0 +1 @@
+{ "type": "module" }
```
icidasset, line 1: What do we get by putting this here?

tests/helpers/page.ts
```diff
@@ -1,7 +1,35 @@
...
+  page.on("pageerror", async error => {
+    console.error("Error in puppeteer: " + error.name, error.message, error.stack)
+  })
+  await page.goto(`file://${htmlPath}`, { waitUntil: 'domcontentloaded' })
+  const { isWebnativeLoaded } = await page.evaluate(async function () {
+    return {
+      isWebnativeLoaded: window.webnative != null,
+    }
+  })
+  if (!isWebnativeLoaded) {
```
icidasset, line 27: 👌

package.json
```diff
@@ -28,53 +45,38 @@
...
+    "node": ">=15"
   },
   "scripts": {
     "docs": "rimraf docs && typedoc && typedoc --entryPoints ./src/fs --out docs/fs",
     "lint": "yarn eslint src/**/*.ts",
-    "prebuild": "rimraf dist && node scripts/gen-version.js",
-    "build": "tsc && rollup -c rollup.config.ts",
+    "prebuild": "rimraf lib dist && node scripts/gen-version.js",
+    "build": "tsc && yarn run build:minified && cp src/package.json lib/package.json",
+    "build:minified": "esbuild src/index.ts --outfile=dist/index.min.js --bundle --minify --sourcemap --platform=browser --format=iife --global-name=globalThis.webnative --target=es2020 && gzip -k9 dist/index.min.js",
```
icidasset, line 55: Does this generate both `index.min.js` and `index.min.js.gz`?

icidasset (approved):
Looks great! Added a few questions.

tsconfig.json
```diff
@@ -1,21 +1,18 @@
 {
   "compilerOptions": {
     "moduleResolution": "node",
-    "target": "es5",
-    "module":"es6",
-    "lib": ["es2015", "es2016", "es2017", "dom"],
+    "target": "ES2020",
+    "module":"ESNext",
```
icidasset, line 5: What's the reason to prefer this over the previously used values?

package.json
```diff
@@ -8,18 +8,35 @@
...
     "files"
   ],
-  "main": "dist/index.cjs.js",
-  "module": "dist/index.es5.js",
-  "browser": "dist/index.umd.min.js",
+  "main": "lib/index.js",
+  "exports": {
+    ".": "./lib/index.js",
+    "./*": "./lib/*.js",
+    "./package.json": "./package.json"
```
icidasset, line 15: Is this referring to the `package.json` in the `src`? If so why doesn't this have `src` or `lib` in the path?

package.json
```diff
@@ -8,18 +8,35 @@
...
+      "*": [
+        "lib/*"
+      ]
+    }
+  },
+  "browser": {
+    "crypto": false
+  },
   "files": [
-    "dist",
```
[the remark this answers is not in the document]
matheus23: Yup. That's an oversight

package.json
matheus23: Yes. Typescript 4.3 support. I forgot why we are using typescript 4.3 now, though :sweat_smile:

matheus23: Oh and unrelated: There's a "library mode" now. We might be able to make use of that instead of running typedoc twice

src/package.json
matheus23: This gets copied into `lib/package.json`. When you're writing nodejs/using a bundler, this will inform node that the files in this directory will use ESNext imports.

I didn't want to put this into the root directory's `package.json`, because that'd mean that we'd have to define config files such as `babel.config.js` or `jest.config.js` using ESNext modules, otherwise node complains. But if we do, then the tools sometimes break.

package.json
icidasset: I thought we were already using that mode, but I'll reinvestigate to make sure 👀

tsconfig.json
matheus23: I want the files in `lib/` to not be transpiled for older versions of node as little as possilbe.
The reasoning for that is: (1) Because of webcrypto usage, we require node >=15 (node >= 16 in terms of LTS) anyway, and that supports ES2020 and ESNext modules, and (2) because applications will either use our pre-bundled `index.min.js` file, or they'll bundle the code (and therefore transpile it to older versions of js) in a way they need for their use-case.

Practically this means that for applications that don't need to support, let's say "pretty old firefox", we get reduced bundle sizes and slightly faster performance.

package.json
matheus23: No this is actually referring to the root `package.json` file. The reason for this line is such that you can do this:
```js
import pkg from "webnative/package.json`
```

If we wouldn't have this line, it would try to load `webnative/lib/package.json.js`.

It's not *really* necessary to expose the package.json, it's just kinda useful, sometimes.

matheus23:
There's some issues left to solve.

1. In webpack, the relative imports between our files don't work (e.g. when `index.js` requires `util`, it `import "./util"`s. But webpack only understands `import "./util.js"` with the `.js` suffix) https://stackoverflow.com/questions/62619058/appending-js-extension-on-relative-import-statements-during-typescript-compilat
2. In snowpack, it doesn't get that when in `webnative`, we've got `browser: { crypto: false }` set, that this should also affect the `noble-25519` library. When I manually change that, it works. I've created a PR in that project: https://github.com/paulmillr/noble-ed25519/pull/24

So, this PR is waiting on those things.

matheus23:
Ignore point (2) above. Not sure why there was an issue, I must have done something wrong. It works flawlessly now.

I still need to fix browser tests. Not sure what's going on. :/

matheus23:
I've reverted changes from the last comment. I've followed another strategy instead.

So, to recap, the issue was: Different environments use different semantics for resolving imports.

In `typescript` it's valid to import a file `util.ts` in a relative import via `import "./util"`. This is also true for some bundlers, e.g. `esbuild` and `vite`. However, if you've got a file `utils.js`, nodejs will not let you `import "./util"`. In nodejs you need to add an extension.

According to [the stackoverflow post I linked above](https://stackoverflow.com/questions/62619058/appending-js-extension-on-relative-import-statements-during-typescript-compilat), most people just use `import "./util.js"` *in typescript* to refer to both `util.ts` (in the source files) and `util.js` in the compiled files in the end. Typescript allows you to use `.js` extensions to refer to `.ts` files (which is a result of [this typescript issue](https://github.com/microsoft/TypeScript/issues/16577)).

However, unfortunately that won't work with `jest`. It uses `babel` and some typescript plugin to let you write your tests in typescript and it doesn't 100% match the typescript import semantics. It doesn't allow relative `.js` imports to refer to `.ts` files. Thus, changing all imports to `.js` doesn't work. I briefly tried to use `ts-jest` instead, but that didn't work out either.

So, I followed the same strategy as [this PR](https://github.com/vasco-santos/ipfs-car/pull/34). They basically have 100% the same problems as we had.
We can transform the `tsc` output by using "transform typescript", which is the `ttypescript` package and `ttsc` CLI. It allows us to install a plugin that adds `.js` to imports in the transpiled .js output.

One thing that it doesn't do: It doesn't correctly resolve `/index.js` imports. So if you have, say `util/index.ts` and you try to `import "./util"`, it won't work. Instead you have to `import "util/index"` and it'll get transformed to `import "util/index.js"`, which resolves fine.

So much for the explanation. Tests are now running and it's working with various environments: vite, snowpack, esbuild, webpack 5, and - as soon as we'll expose the dependency injection stuff - it'll work in nodejs.

package.json
```diff
@@ -28,53 +43,39 @@
...
     "docs": "rimraf docs && typedoc && typedoc --entryPoints ./src/fs --out docs/fs",
     "lint": "yarn eslint src/**/*.ts",
-    "prebuild": "rimraf dist && node scripts/gen-version.js",
-    "build": "tsc && rollup -c rollup.config.ts",
-    "start": "tsc -w",
+    "prebuild": "rimraf lib dist && node scripts/gen-version.js",
+    "build": "ttsc && yarn run build:minified && cp src/package.json lib/package.json",
+    "build:minified": "esbuild src/index.ts --outfile=dist/index.min.js --bundle --minify --sourcemap --platform=browser --format=iife --global-name=globalThis.webnative --target=es2020 && gzip -k9 dist/index.min.js",
+    "start": "ttsc -w",
     "test": "jest --forceExit",
```
walkah, line 55: Not totally relevant, but do we want to switch stuff around so `yarn test` works "out of the box"? I would suggest make it the same as `test:unit` and drop `test:unit` ... maybe `test:prod` should be `test:integration` to be more accurate?

walkah (approved):
3ov9jK5HPcBvwsNucE

Thanks for cleaning all this stuff up!

icidasset:
Took a look at the latest changes, looks good 👍  Bit of a shame we have to move away from the `dir/index.ts` pattern, but I guess we can find a better alternative when do the next big refactor.

matheus23:
>  Bit of a shame we have to move away from the `dir/index.ts` pattern

Yes. :confused: 
We could change all `x/index.ts` files to be `x.ts` instead. :thinking:

PR #254, opened 2021-06-23
Also clear the keystore when it wasn't loaded before

matheus23:
Fixes a bug where calling `webnative.crypto.keystore.clear()` wouldn't actually clear the keystore unless you ran some other keystore operation before which would've internally loaded it.


icidasset (approved):
Nice catch 👌

PR #255, opened 2021-06-23
Add testing environment variable for GitHub Action

bgins:
## Summary

* [x] Add testing environment variable for the GH Action
* [x] Add an npm script that uses it

The immediate goal is to increase the number of runs of the pubic and private API tests in the GitHub action, but we could use the environment variable in other/future tests.

## Test plan (required)

Tests should pass in the GitHub action. It should take longer than last time.


matheus23 (approved):
Nice! :+1:

PR #258, opened 2021-07-06
Set ES2015 as target in tsconfig

### LINKED ISSUE #257: Parse failure with webpack 4.x / create-react-app
# Summary

## Problem

Adding `webnative` to a create-react-app built app results in a parse error:

```
./node_modules/webnative/lib/fs/v1/PublicTree.js 157:49
Module parse failed: Unexpected token (157:49)
File was processed with these loaders:
 * ./node_modules/react-scripts/node_modules/babel-loader/lib/index.js
You may need an additional loader to handle the result of these loaders.
|       return { ...acc,
|         [cur.name]: { ...cur,
>           isFile: this.header.skeleton[cur.name]?.isFile
|         }
|       };
```

### Impact

It's too hard to get webpack integrated with a simple react app.

### Versions
- webnative v0.26.0
- create-react-app 4.0.3 (webpack 4.44)

matheus23:
Fixes #257

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,7 +1,7 @@
 {
   "compilerOptions": {
     "moduleResolution": "node",
-    "target": "ES2020",
+    "target": "ES2015",
     "module":"ESNext",
     "lib": ["dom"],
     "strict": true,


matheus23:
Size comparison:

The `dist/index.min.js` file size is unaffected (548346 bytes).

All `*.js` files in `lib/`:
Before: 161056 bytes
After: 214605 bytes

walkah (approved):
Seth Meyers Reaction - 26gsjCZpPolPr3sBy

PR #261, opened 2021-07-16
Bring back the umd build

matheus23:
## Summary

When we switched to `esbuild`, we dropped UMD support, because it doesn't support that natively. This PR brings it back using esbuild's `banner` and `footer` options.

The motivation for this was that TiddlyWiki uses and expects UMD in the browser.

## Test plan (required)

There's a [test](https://matheus23.files.fission.name/p/webnative-umd/) running webnative in the browser using the UMD build, it works.

I didn't test TiddlyWiki, yet. I've sent @saqimtiaz a message on discord, so he can give me some feedback.

## After Merge
* Make a release

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+### Next
+
+- Bring back the [UMD](https://github.com/umdjs/umd) build (#261)
+
+
 ### v0.26.0
 
 - Export maps, you can now import `webnative/path` instead of `webnative/dist/path`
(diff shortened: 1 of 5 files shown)


icidasset:
What's the reason TiddlyWiki needs UMD specifically? So UMD works in commonjs/amd/browser contexts right? And IIFE is just for the browser (ie. what we had before)?

matheus23:
TiddlyWiki's plugin system apparently expects CommonJS-style objects. So I guess something that has `module.exports = ` set. I honestly don't really know.

@saqimtiaz mentioned that the update broke him. Maybe he can elaborate?

However, adding UMD wrappers around the module really shouldn't do any harm. It should be a write once, forget about it thing. (Don't quote me on that :smile: )

icidasset (approved):
Yeah the UMD pattern is fine, has been around for quite a while. Although, I don't know if anyone uses the AMD module system anymore 😅 

Launch it!

Issue #260, opened 2021-07-14
NPM install issue missing dependency


icidasset:
## Problem

```
│ sh: rimraf: command not found
│ npm ERR! code ELIFECYCLE
│ npm ERR! syscall spawn
│ npm ERR! file sh
│ npm ERR! errno ENOENT
│ npm ERR! base58-universal@1.0.1-0 clear: `rimraf dist && mkdir dist`
│ npm ERR! spawn ENOENT
│ npm ERR! 
│ npm ERR! Failed at the base58-universal@1.0.1-0 clear script.
```
Because we're currently using a git dependency for `base58-universal`, it tries to run it's build task, resulting in a missing dependency.

## Solution

1. Don't use a git dependency
2. Add `rimraf` as a dependency along side your `webnative` dependency

matheus23:
I've looked into this a little. We *might* not need the `base58-universal` dependency.
We're using [uint8arrays](https://github.com/achingbrain/uint8arrays) and they support all bases that [multibase](https://github.com/multiformats/js-multiformats#multibase-encoders--decoders--codecs) support (since version 2.1.6 as far as I understand), and `base58btc` is among them.

PR #262, opened 2021-07-16
Replace base58-universal dependency with uint8arrays (& add changelog checker)

matheus23:
## Summary

This PR replaces usage of `base58-universal` with `uint8arrays`

We depend on a *branch* in the `base58-universal` github repository, instead of on the npm package. This means it has to be built by anyone using webnative (if they use webnative as an npm package). This process can be flaky (#260).

Instead, we can depend on `uint8arrays`, which is a dependency of many of *our* dependencies. E.g. [js-ipfs-core depends on it](https://github.com/ipfs/js-ipfs/blob/dc041aa3de789290bfc5c1c0b8c2878d8c79186a/packages/ipfs-core/package.json#L123).
It makes sense to go with ecosystem momentum here.

## Test plan (required)

The tests run :white_check_mark: . Only DID parsing/generation are affected by this change, and they're fairly well covered by our tests. The behavior is the same as before.

## After Merge
* Release!

### DIFF
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -16,7 +16,7 @@ jobs:
         uses: Zomzog/changelog-checker@v1.2.0
         with:
           fileName: CHANGELOG.md
-          noChangelogLabel: no changelog
+          noChangelogLabel: "no changelog"
           checksNotification: Simple
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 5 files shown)


matheus23:
Please excuse that I'm mixing PR concerns here. I'm testing out the changelog checker github action. It seems to work fine :)

If you want to disable the checker, you can opt-out by putting the `no changelog` label on a PR.

This action only runs for PRs that merge into main.

---

I don't think we need a changelog item for this PR, as it's purely internals that we're changing.

icidasset (approved):

PR #263, opened 2021-07-16
Typedoc Improvements

matheus23:
## Summary

Some typedoc improvements: There's now a root 'readme' for the typedocs, and we're only running typedoc once and configured it slightly differently.

Our typedocs previously weren't linked to each other correctly. So e.g. I wouldn't be able to take a look at the `FileContent` type when I was inside the `FileSystem` class. Also there were no links to `FileSystem` form the root of the docs, you'd have to know about `webnative.fission.app/fs`. 

## Test plan (required)

Hm. Ideally I want the docs to be rendered in github actions, but for that I'd need access to the fission account that belongs to `webnative.fission.app` (I think you have that @icidasset?).

### DIFF
--- a/scripts/TYPEDOC.md
+++ b/scripts/TYPEDOC.md
@@ -1,0 +1,30 @@
+# Webnative API Reference
+
+[version](/modules/common_version.html#VERSION)
+
+This is source-derived, auto-generated API reference for [webnative](https://github.com/fission-suite/webnative).
+
+If you need an introduction to webnative, we recommend reading the [fission guide](https://guide.fission.codes/developers/webnative) instead. This document is meant to be used as API reference.
+
+## Which Docs to Look At
+
+When you import webnative like this:
+```js
+import * as webnative from "webnative"
+```
+
+or like this:
+
+```js
+const webnative = require("webnative")
+```
+
+or using a script tag:
+```html
+<script src="https://unpkg.com/webnative@latest/dist/index.umd.min.js">
+```
+
+Then you'll have a `webnative` variable at disposal which contains the [`index` module](/modules/index.html).
+
+Once you've loaded a filesystem, you'll have an instance of the [`FileSystem` class](/classes/fs_filesystem.FileSystem.html). Read its docs to see what filesystem operations are supported.
+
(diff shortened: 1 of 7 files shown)


scripts/TYPEDOC.md
```diff
@@ -0,0 +1,43 @@
+# Webnative API Reference
+
+[version](/modules/common_version.html#VERSION)
+
+This is source-derived, auto-generated API reference for [webnative](https://github.com/fission-suite/webnative).
```
icidasset, line 5: I would say "the source-derived"

scripts/TYPEDOC.md
```diff
@@ -0,0 +1,43 @@
...
+
+If you need an introduction to webnative, we recommend reading the [fission guide](https://guide.fission.codes/developers/webnative) instead. This document is meant to be used as API reference.
```
icidasset, line 7: Would make this "the API reference" or "an API reference".

scripts/TYPEDOC.md
```diff
@@ -0,0 +1,43 @@
...
+Then you'll have a `webnative` variable at disposal which contains the [`index` module](/modules/index.html).
+
+Once you've loaded a filesystem, you'll have an instance of the [`FileSystem` class](/classes/fs_filesystem.FileSystem.html). Read its docs to see what filesystem operations are supported.
+
+
+## These Docs for another Webnative Version
+
+The docs at [webnative.fission.app](https://webnative.fission.app) will always be the docs for the most recent webnative version.
+
+If you want to look at the docs for an older webnative version, keep in mind that webnative's npm package ships with these docs at`node_modules/webnative/docs/`. So if you want to take a look at them, you can start your own local http server:
```
icidasset, line 37: You can also just open the HTML file in your browser, which is easier. We only would need to make the links in this README relative instead of absolute.

icidasset (approved):
Few suggestions regarding the README, looks good otherwise 👌

scripts/TYPEDOC.md
matheus23: Good idea!

matheus23:
I think auto-publishing to `webnative.fission.app` should be another PR, along with auto-publishing the npm module. Those two things should happen at the same time, after all.

I guess that's also a release-workflow kinda thing.

PR #266, opened 2021-07-20
Fix #259

### LINKED ISSUE #259: Account signup user name validation issues
We're not catching some invalid username on signup. This was caught investigating fission-suite/auth-lobby#81 

Usernames like this:

* `jeff--`

...get an error like 

`Sorry, jeff-- is not a valid username. You can use letters, numbers and hyphens in between.`

After some testing, it looks like the code involved doesn't catch the '_' and '@' characters, leading to the creation of an invalid  personal address for the account.

matheus23:


matheus23:
Closing this because of https://github.com/fission-suite/webnative/issues/259#issuecomment-887750886

Issue #265, opened 2021-07-19
Webpack 5 module resolution errors


bgins:
# Summary

## Problem

Building webnative 0.26.0 with webpack 5 fails with module resolution errors.

### Impact

It may be challenging to build with some webpack configs.

# Detail

**Describe the bug**

When building with webpack 5, two types of build errors occur
- Can't import from default-exporting module
- Module failed to resolve only because it was resolved as fully specified. 
 
Here are example outputs for each:

```
ERROR in ./node_modules/webnative/lib/common/base64.js 3:11-22
Can't import the named export 'Buffer'.'from' (imported as 'Buffer') from default-exporting module (only default export is available)
 @ ./node_modules/webnative/lib/auth.js 3:0-45 59:271-287
 @ ./node_modules/webnative/lib/index.js 130:0-26 130:0-26
 @ ./src/index.js 1:0-39 17:0-20 19:9-41 20:9-40 23:36-55 33:9-41 34:9-41 41:4-29 50:34-53 64:39-63 86:32-51 93:42-61 113:36-55 125:34-53 126:36-55

ERROR in ./node_modules/webnative/lib/crypto/browser.js 2:0-51
Module not found: Error: Can't resolve 'keystore-idb/rsa/index' in '/home/thuselem/development/work/fission/fission-stored-wasm-example/node_modules/webnative/lib/crypto'
Did you mean 'index.js'?
BREAKING CHANGE: The request 'keystore-idb/rsa/index' failed to resolve only because it was resolved as fully specified
(probably because the origin is a '*.mjs' file or a '*.js' file where the package.json contains '"type": "module"').
The extension in the request is mandatory for it to be fully specified.
Add the extension to the request.
 @ ./node_modules/webnative/lib/setup/dependencies.js 1:0-54 5:16-36 8:17-38 9:17-38 10:19-42 11:20-44 14:16-39 17:16-43 20:23-52 21:24-54 22:17-40 23:14-34 24:23-52 25:23-52 26:19-44 27:16-38 28:15-36
 @ ./node_modules/webnative/lib/setup.js 28:0-58 28:0-58
 @ ./node_modules/webnative/lib/index.js 140:0-36 140:0-36
 @ ./src/index.js 1:0-39 17:0-20 19:9-41 20:9-40 23:36-55 33:9-41 34:9-41 41:4-29 50:34-53 64:39-63 86:32-51 93:42-61 113:36-55 125:34-53 126:36-55
```

**To Reproduce**
[...]

Issue #264, opened 2021-07-16
TypeError with webpack 4.x and create-react-app


bgins:
## Problem

Using webnative 0.26 with the `react-scripts` local dev server and build results in TypeErrors.

### Impact

It is difficult to use webnative 0.26 with an app generated by create-react-app.

# Detail

When running `react-scripts start`, the local dev server fails with an error in the browser:

![cra-wn-26](https://user-images.githubusercontent.com/7957636/125972345-e299d82d-b0e2-47de-ad41-8bb191bf6856.png)

Running `react-scripts build` succeeds, but the built app does not load and fails with a `Uncaught TypeError: can't convert BigInt to number` error.

**To Reproduce**
1. Clone https://github.com/fission-suite/react-todomvc
2. Upgrade it to webnative 0.26.0
3. Test the local dev server with `npm start`
4. Test the build with `npm run build`

matheus23:
Can you try it with current `main` (not that [this PR](https://github.com/fission-suite/webnative/pull/258) is merged with a similar webpack 4.x issue)?

bgins:
Tried this out on `main` at https://github.com/fission-suite/webnative/commit/01360d92ec4a182828d7086397dc938a8ed4e8e8 and I'm seeing the same errors.

To make sure I am testing this correctly, these were my steps:

1. Pull the latest webnative on `main`
2. `nix-shell`, `yarn install`, `yarn build` to build webnative
3. In my project, `npm install ../webnative` (that's where webnative is relatively)
4. Run my project as above in the original bug report

I think that's the right way to test this, but just want thorough to make sure I didn't overlook something!

matheus23:
I figured out the source of the problem: The stacktrace in the image somehow shows the wrong code for line 36. It's actually a line containing `Buffer.concat(...)`. That function won't work if its arguments aren't buffers, which for some reason is the case with your build (but not with lots of other builds???).

I'll just remove the last remaining usage of `Buffer` from webnative, I think it makes sense to switch to `Uint8Array`s and the `uint8arrays` library anyway (because that's what supported in *both* nodejs and the browser natively).

bgins:
Nice, good find! Yeah, that makes sense to switch over to `Uint8Array`.

matheus23:
Thanks!
[...]

PR #268, opened 2021-07-22
Fix keystore-idb imports: add .js extensions and remove buffer usage

matheus23:
Esbuild and typescript (only typechecking) handle extensionless imports just fine, but the compiled output js modules are invalid, and some bundlers will fail (e.g. webpack 5).

Also, we still had some conversions/usage of nodejs `Buffer` *helper functions*, e.g. `Buffer.concat` or `Buffer.from` (to convert to base64 and vice versa). These will fail if js-ipfs decides it'll give use `Uint8Arrays` which happen to not be `Buffer`s in some cases (:man_shrugging:).
(Remember: Every `Buffer` is a `Uint8Array`, but js-ipfs only guarantees we get `Uint8Array`s.)

## Closing issues

* #264  
* #265


tests/fs/tree.test.ts
```diff
@@ -24,10 +24,10 @@ describe("FS", () => {
 
       return [
         await fs.read(privatePath),
-        await fs.read(publicPath)
+        new TextDecoder().decode(await fs.read(publicPath))
```
matheus23, line 27: This didn't fail before, because `Buffer`s would return their content interpreted as a utf8 string if forced to be turned to a string (e.g. via `[<buffer>].join("")`).
But we really shouldn't depend on that behavior.

At some point we should really add an encoding parameter to webnative's filesystem operations similar to how it works in many other APIs (i.e. `fs.read(path, { encoding: 'utf8' })`, and `fs.write(path, content, { encoding: 'utf8' })`).

src/common/api.test.ts
```diff
@@ -1,32 +0,0 @@
-import * as dns from '../dns/index'
-import { did } from './api'
-
-jest.mock('../dns')
```
matheus23, line 4: This might be the most controversial thing in this PR: I'm deleting this test, because I didn't figure out mocking yet. It's [supposedly possible](https://sinonjs.org/how-to/stub-dependency/), but I just haven't looked into it much.

I'm not sure whether introducing a mocking library to keep this test alive would be worth it, though.

I'm open to other's opinions, though!

icidasset (approved):

PR #270, opened 2021-08-04
Add namefilter tests

bgins:
## Summary

This PR adds an initial set of tests of the namefilter implementation used in the private filesystem.

* [x] Add hex to bloom filter conversion tests
* [x] Add bare filter bit count tests
* [x] Add revision filter test

We run each test 10,000 times because they are fast and they finish before some of our other longer running tests.

## Test plan (required)

The new tests should pass in the GH action and can also be run in isolation with
 
```
yarn test --grep "filter"
```

All tests should pass.

### DIFF
--- a/src/fs/protocol/private/namefilter.test.ts
+++ b/src/fs/protocol/private/namefilter.test.ts
@@ -1,0 +1,125 @@
+import * as fc from "fast-check"
+import expect from "expect"
+
+import { BloomFilter } from "fission-bloom-filters"
+
+import * as namefilter from "./namefilter.js"
+
+
+describe("hex bloom filter conversion", () => {
+
+  before(() => {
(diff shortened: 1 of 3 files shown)


matheus23 (approved):
These tests are great!
Super nitpicky, but can you change the single quotes `'` to double quotes `"`? :pray:

bgins:
> Super nitpicky, but can you change the single quotes `'` to double quotes `"`?

Yeah, totally happy to change that!

> it seems like the unsigned commits are keeping this PR from getting merged. Can you see whether there's an easy way to fix this?

Yep, I think there is a way to retroactively sign commits that's not too much effort. Lemme give that a try and that way we can do it properly.

bgins:
I went ahead and squashed everything down locally, that way it would be all signed as one commit.

PR #273, opened 2021-08-06
Add extra lobby params

### LINKED ISSUE #271: Add theme param to redirectToLobby function
Or something more generic, like a `extraLobbyParams` object.
https://github.com/fission-suite/auth-lobby#theming

bgins:
## Summary

This PR adds an `extraLobbyParams` parameter to `redirectToLobby`. The properties on `extraLobbyParams` get converted into query params and sent to the auth lobby.

* [x] Add `extraLobbyParams` to `redirectToLobby`

Initially, `extraLobbyParams` is intended for use with theming (https://github.com/fission-suite/auth-lobby#theming), but it could be used for other purposes in the future.

**Note:** In this implementation, a user must pass something to `redirectToLobby` for `permissions` and `redirectTo` in order to use `extraLobbyParams`. For example:

```
wn.redirectToLobby(state.permissions, window.location.href, { theme: <CID> })
``` 
In a future release, we could pass a single object to `redirectToLobby` with `permissions`, `redirectTo`, and any additional parameters.

## Test plan (required)

Calling `redirectToLobby` with `extraLobbyParams` should add query params to the redirect URL. For example:
```
wn.redirectToLobby(state.permissions, window.location.href, { dogs: "yes", cats: "yes"})
```
Would append `&dogs=yes&cats=yes` to the URL.

An empty `extraLobbyParams` object should not add additional query params.

Calling `redirectToLobby` with only `permissions` or `permissions` and `redirectTo` should work like normal and not add additional query params.

## Closing issues

Closes #271 

## After Merge

* [x] Does this change invalidate any docs or tutorials? It does not invalidate, but we can add a theming section to the guide! :paintbrush: :art: 

### DIFF
--- a/src/auth.ts
+++ b/src/auth.ts
@@ -50,10 +50,14 @@ export async function leave({ withoutRedirect }: { withoutRedirect?: boolean } =
  *                    Pass `null` if working without permissions.
  * @param redirectTo Specify the URL you want users to return to.
  *                   Uses the current url by default.
+ * @param extraLobbyParams Extra parameters object. Properties on the
+ *                         object are converted to query params and sent
+ *                         to the auth lobby.
  */
 export async function redirectToLobby(
   permissions: Maybe<Permissions>,
-  redirectTo?: string
+  redirectTo?: string,
+  extraLobbyParams?: Record<string, string>
 ): Promise<void> {
   const app = permissions?.app
   const fs = permissions?.fs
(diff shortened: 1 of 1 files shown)


src/auth.ts
```diff
@@ -50,10 +50,14 @@ export async function leave({ withoutRedirect }: { withoutRedirect?: boolean } =
...
  *                   Uses the current url by default.
+ * @param extraLobbyParams Extra parameters object. Properties on the
+ *                         object are converted to query params and sent
+ *                         to the auth lobby.
  */
 export async function redirectToLobby(
   permissions: Maybe<Permissions>,
-  redirectTo?: string
+  redirectTo?: string,
+  extraLobbyParams?: Record<string, string>
```
bgins, line 60: Wondering if we might want to make this type more flexible? For example, would `Record<string, string | number>` make sense and we can convert `number`s to strings for the user? Maybe some other types?

matheus23: I personally prefer `Record<string,string>`, because that's what it'll be in the end in the URL. There's an infinite amount of types that *could* be converted to strings and no good way to choose which to support, I think.
Might be confusing if we choose to support converting `boolean` and `number`, but not `bigint` or `{ a: boolean, b: boolean }`.

matheus23 (approved):

src/auth.ts
bgins: True, aside from `string`, it's hard to predict what other types would make sense and we'd just convert them to strings anyways. 

Let's keep it at `Record<string,string>` for now. If there is eventually a compelling reason to add more types, we can revisit this.

PR #275, opened 2021-08-09
Fix regression: Transpile down to ES2018

matheus23:


bgins (approved):
Looks good! :tada: :taco:

matheus23:
Ironically, I'm not even sure we want to merge this anymore :sweat_smile: turns out we need ES2020 in our environments anyway.

matheus23:
Closing this, because we need our environment to have ES2020 support anyway, because we're relying on the `noble-ed25519` library, which uses native `BigInt`s, and those are an ES2020 feature.
Thus, we're going to document ways to configure some popular environments to support ES2020 instead.

PR #277, opened 2021-08-17
Replace noble-ed25519 with tweetnacl

### LINKED ISSUE #276: Add more build targets
_Note: This issue may not be resolved with additional build targets. The goal is to document and consider the options._

# Summary

Webnative builds currently targets `es2020`. We would like to support a wider range of targets, as old as possible.

## Problem

Webnative will not work in some browsers that lack full `es2020` support. Mostly, this means Internet Explorer and less used mobile browsers, though notably `export * as namespace` is not available in Safari on iOS or Android Browser. See [es2020 features summary](https://caniuse.com/?feats=mdn-javascript_operators_optional_chaining,mdn-javascript_operators_nullish_coalescing,mdn-javascript_builtins_globalthis,es6-module-dynamic-import,bigint,mdn-javascript_builtins_promise_allsettled,mdn-javascript_builtins_string_matchall,mdn-javascript_statements_export_namespace,mdn-javascript_statements_import_meta).

In addition to browsers, some bundlers and build tools may not support `es2020` and may require users to target some other environment. Notably, SvelteKit only supports `es2019` builds.

### Impact

Webnative will not work in some older browsers, mostly notably Internet Explorer. We have not encountered build environment issues in Safari for iOS or Android Brower. (Please add a comment below if you have noticed otherwise!)

Webnative will not build in SvelteKit due to the lack of support for big integer literals. SvelteKit does not currently offer a way to configure the Vite build target, which is set at `es2019`. See https://github.com/sveltejs/kit/issues/859#issuecomment-894556133 for a summary.

There may be additional bundlers and build environments with similar restrictions. (Please add them below :pray:)

## Solution

We have explored a couple of solutions. None stands out yet as the right way to go. See the details below.

# Detail

### Is your feature request related to a problem? Please describe.

Webnative does not build in SvelteKit.

### Describe the solution you'd like

Unclear. No solution stands out as the best option yet.

### Describe alternatives you've considered

**Transpile or polyfill to older targets.** Unfortunately, this does not seem like a viable solution. Webnative uses [noble-ed25519](https://github.com/paulmillr/noble-ed25519) which uses big integer literals. The general consensus is that big integer literals are "impossible to polyfill and infeasible to transpile because the performance would be terrible".

Google wrote the [JSBI](https://github.com/GoogleChromeLabs/jsbi) with this in mind. In their README:

> To use BigInts in code that you want to run everywhere, you need a library. But there’s a difficulty: the BigInt proposal changes the behavior of operators (like +, >=, etc.) to work on BigInts. These changes are impossible to polyfill directly; and they are also making it infeasible (in most cases) to transpile BigInt code to fallback code using Babel or similar tools. The reason is that such a transpilation would have to replace every single operator in the program with a call to some function that performs type checks on its inputs, which would incur an unacceptable performance penalty.

Esbuild cites this as their reason for not transforming big integer literals: https://github.com/evanw/esbuild/issues/732#issuecomment-770519567

**Find replacement libraries.** We may be able to find alternative implementations that do not rely on big integer literals or other features that cannot be transpiled. 

Open questions:

1. Do other libraries use big integer literals or other `es2020`-only features?

2. Does webnative use any of these features directly?

One related is alternative is asking library authors about alternatives that use JSBI or similar alternatives that provide compatibility with older build targets. This request may be declined because, in some cases, it would essentially mean a re-write.

**Do nothing. Wait for the ecosystem to catch up.** This would mean that webnative would not be available in some environments for a time.

**Something else?** Please suggest other alternatives below. This is all based on initial research that should not be considered exhaustive or complete.

bgins:
## Summary

* [x] Replace `noble-ed25519` with `tweetnacl`
* [x] Update `tsconfig` target to `ES2018`

Explain the **motivation** for making this change. What existing problem does the pull request solve?

The `noble-ed25519` library uses big integer literals which only work in `es2020` or newer environments. This PR replaces `noble-ed25519` with `tweetnacl` to support older environments. See #276 for more details.

We use the [verify](https://github.com/paulmillr/noble-ed25519#verifysignature-hash-publickey) method from `noble-ed25519`. This PR replaces it with [sign.detached.verify](https://github.com/dchest/tweetnacl-js#naclsigndetachedverifymessage-signature-publickey) from `tweetnacl`. 

The function signatures are nearly the same. `noble-ed25519` permits more types and `tweetnacl` requires `Uint8Array`s. This is enforced here:

https://github.com/dchest/tweetnacl-js/blob/3389b7c9f00545e516a0fdafb324b7857cf1527d/nacl.js#L949-L954

In any case, our wrapper function enforces `Uint8Array`s so we should be fine for types. 

`noble-ed25519` returns `Promise<bool>` and `tweetnacl` returns `bool`. The `tweetnacl` result is wrapped in a promise here to keep the original behavior. 

## Test plan (required)

The two environments that lack `es2020` support that we are aware of are Create React App and SvelteKit. See https://github.com/sveltejs/kit/issues/2220 for a discussion regarding SvelteKit. 

There are bundler issues with SvelteKit that make this difficult to test. Create React App can be tested as follows:

```bash
create-react-app my-app
cd my-app
npm install webnative
```

Import `webnative` in `src/App.js`
```javascript
import * as wn from 'webnative';
```

Build and run the app locally (with `http-server` or something else)
```bash
npm run build
http-server build/
```
This should fail with a `Uncaught TypeError: can't convert BigInt to number` runtime error.

Now replace webnative with the version from this PR, build again, and run it. The app should run without an error.

⚠️ We should test this more somehow! Getting it to work with SvelteKit would be nice, but there are likely other things we can do to check that the libraries behave in the same way.

*Edit: These changes pass in our test suites, and beyond that we can rely on these libraries to work equivalently. See https://github.com/fission-suite/webnative/pull/277#issuecomment-920145665 for a summary of where this PR stands.*

## Closing issues

Closes #276 

## After Merge
* [x] Does this change require a release to be made? Is so please create and deploy the release

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,7 +1,7 @@
 {
   "compilerOptions": {
     "moduleResolution": "node",
-    "target": "ES2020",
+    "target": "ES2018",
     "module":"ESNext",
     "lib": ["dom", "es2021"],
     "strict": true,
--- a/yarn.lock
+++ b/yarn.lock
@@ -3,115 +3,115 @@
 
 
 "@assemblyscript/loader@^0.9.4":
-  version "0.9.4"
-  resolved "https://registry.yarnpkg.com/@assemblyscript/loader/-/loader-0.9.4.tgz#a483c54c1253656bb33babd464e3154a173e1577"
-  integrity sha512-HazVq9zwTVwGmqdwYzu7WyQ6FQVZ7SwET0KKQuKm55jD0IfUpZgN0OPIiZG3zV1iSrVYcN0bdwLRXI/VNCYsUA==
-
-"@babel/code-frame@7.12.11":
-  version "7.12.11"
-  resolved "https://registry.yarnpkg.com/@babel/code-frame/-/code-frame-7.12.11.tgz#f4ad435aa263db935b8f10f2c552d23fb716a63f"
-  integrity sha512-Zt1yodBx1UcyiePMSkWnU4hPqhwq7hGi2nFL1LeA3EUl+q2LQx16MISgJ0+z7dnmgvP9QtIleuETGOiOH1RcIw==
-  dependencies:
-    "@babel/highlight" "^7.10.4"
+  "integrity" "sha512-HazVq9zwTVwGmqdwYzu7WyQ6FQVZ7SwET0KKQuKm55jD0IfUpZgN0OPIiZG3zV1iSrVYcN0bdwLRXI/VNCYsUA=="
+  "resolved" "https://registry.npmjs.org/@assemblyscript/loader/-/loader-0.9.4.tgz"
+  "version" "0.9.4"
 
 "@babel/code-frame@^7.12.13":
-  version "7.12.13"
-  resolved "https://registry.yarnpkg.com/@babel/code-frame/-/code-frame-7.12.13.tgz#dcfc826beef65e75c50e21d3837d7d95798dd658"
-  integrity sha512-HV1Cm0Q3ZrpCR93tkWOYiuYIgLxZXZFVG2VgK+MBWjUqZTundupbfx2aXarXuw5Ko5aMcjtJgbSs4vUGBS5v6g==
+  "integrity" "sha512-HV1Cm0Q3ZrpCR93tkWOYiuYIgLxZXZFVG2VgK+MBWjUqZTundupbfx2aXarXuw5Ko5aMcjtJgbSs4vUGBS5v6g=="
+  "resolved" "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.12.13.tgz"
+  "version" "7.12.13"
   dependencies:
     "@babel/highlight" "^7.12.13"
 
+"@babel/code-frame@7.12.11":
+  "integrity" "sha512-Zt1yodBx1UcyiePMSkWnU4hPqhwq7hGi2nFL1LeA3EUl+q2LQx16MISgJ0+z7dnmgvP9QtIleuETGOiOH1RcIw=="
+  "resolved" "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.12.11.tgz"
+  "version" "7.12.11"
+  dependencies:
+    "@babel/highlight" "^7.10.4"
+
 "@babel/helper-validator-identifier@^7.14.0":
-  version "7.14.0"
-  resolved "https://registry.yarnpkg.com/@babel/helper-validator-identifier/-/helper-validator-identifier-7.14.0.tgz#d26cad8a47c65286b15df1547319a5d0bcf27288"
-  integrity sha512-V3ts7zMSu5lfiwWDVWzRDGIN+lnCEUdaXgtVHJgLb1rGaA6jMrtB9EmE7L18foXJIE8Un/A/h6NJfGQp/e1J4A==
+  "integrity" "sha512-V3ts7zMSu5lfiwWDVWzRDGIN+lnCEUdaXgtVHJgLb1rGaA6jMrtB9EmE7L18foXJIE8Un/A/h6NJfGQp/e1J4A=="
+  "resolved" "https://registry.npmjs.org/@babel/helper-validator-identifier/-/helper-validator-identifier-7.14.0.tgz"
+  "version" "7.14.0"
 
 "@babel/highlight@^7.10.4", "@babel/highlight@^7.12.13":
-  version "7.14.0"
-  resolved "https://registry.yarnpkg.com/@babel/highlight/-/highlight-7.14.0.tgz#3197e375711ef6bf834e67d0daec88e4f46113cf"
-  integrity sha512-YSCOwxvTYEIMSGaBQb5kDDsCopDdiUGsqpatp3fOlI4+2HQSkTmEVWnVuySdAC5EWCqSWWTv0ib63RjR7dTBdg==
+  "integrity" "sha512-YSCOwxvTYEIMSGaBQb5kDDsCopDdiUGsqpatp3fOlI4+2HQSkTmEVWnVuySdAC5EWCqSWWTv0ib63RjR7dTBdg=="
+  "resolved" "https://registry.npmjs.org/@babel/highlight/-/highlight-7.14.0.tgz"
+  "version" "7.14.0"
   dependencies:
     "@babel/helper-validator-identifier" "^7.14.0"
-    chalk "^2.0.0"
-    js-tokens "^4.0.0"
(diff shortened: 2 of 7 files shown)


matheus23:
> There are bundler issues with SvelteKit that make this difficult to test

What exactly are these issues?

matheus23:
I was just wondering about performance differences. This is from the noble-ed25519 readme:

```sh
# noble-ed25519
[...]
sign x 3,247 ops/sec @ 307μs/op
[...]
# tweetnacl-fast@1.0.3
[...]
sign x 519 ops/sec @ 2ms/op
```

No comparison for `verify`, though. There's some mentions about 

> verify 100 messages per second on a laptop

In tweetnacl's readme, but the numbers on noble's readme are from an Apple M1:

> verify x 726 ops/sec @ 1ms/op

bgins:
> > There are bundler issues with SvelteKit that make this difficult to test
> 
> What exactly are these issues?

Thinking about it some more, these problems might be the result of how I am testing this locally. Let me explain a bit about how I am testing this with SvelteKit.

SvelteKit has this notion of [adapters](https://kit.svelte.dev/docs#adapters) that are meant to target different environments like `node`, `netlify`, `vercel`, or a `static` build. For our purposes, I have been using [adapter-static](https://github.com/sveltejs/kit/tree/master/packages/adapter-static) with with server-side rendering disabled by setting the `ssr` to false in the Svelte config. The `svelte.config.js` I am using looks like:

```javascript
import adapter from '@sveltejs/adapter-static';

/** @type {import('@sveltejs/kit').Config} */
const config = {
	kit: {
		adapter: adapter({
			// default options are shown
			pages: 'build',
			assets: 'build',
			fallback: null
		}),
		ssr: false,
		// hydrate the <div id="svelte"> element in src/app.html
		target: '#svelte'
	}
};

export default config;
```
To get a similar setup
```bash
npm init svelte@next my-app
cd my-app
npm install
npm i -D @sveltejs/adapter-static@next
```
select the "SvelteKit demo app", "no" to the remaining questions, and the replace `svelte.config.js` with the one above.

Import `webnative` in `src/routes/__layout.svelte`

```javascript
<script>
	import * as wn from 'webnative'
	
	import Header from '$lib/header/Header.svelte';
	import '../app.css';
</script>
```

When I install my local build of webnative with `npm install ../webnative` (i.e. the relative path to it in my filesystem) and then build the project, I get this error:

```
> Using @sveltejs/adapter-static
> Cannot find package 'ipfs-message-port-protocol' imported from /home/brian/code/fission/my-app/.svelte-kit/output/server/app.js
Error [ERR_MODULE_NOT_FOUND]: Cannot find package 'ipfs-message-port-protocol' imported from /home/brian/code/fission/my-app/.svelte-kit/output/server/app.js
    at new NodeError (node:internal/errors:371:5)
    at packageResolve (node:internal/modules/esm/resolve:712:9)
    at moduleResolve (node:internal/modules/esm/resolve:753:18)
    at Loader.defaultResolve [as _resolve] (node:internal/modules/esm/resolve:867:11)
    at Loader.resolve (node:internal/modules/esm/loader:89:40)
    at Loader.getModuleJob (node:internal/modules/esm/loader:242:28)
    at ModuleWrap.<anonymous> (node:internal/modules/esm/module_job:76:40)
    at link (node:internal/modules/esm/module_job:75:36)
```

I see `ipfs-message-port-protocol` down in `node_modules/webnative/node_modules` but not at the top level. Maybe this error is a quirk of testing locally?

bgins:
> I was just wondering about performance differences.

Yeah, `noble-ed25519` is way faster at signing! 

Do you have a sense for how many verify operations we might be doing? Would this lead to a performance issue for us?

bgins:
Hmm, looks like this didn't work out in our GH action test

``` 
import { sign } from "tweetnacl";
         ^^^^
SyntaxError: Named export 'sign' not found. The requested module 'tweetnacl' is a CommonJS module, which may not support all module.exports as named exports.
CommonJS modules can always be imported via the default export, for example using:

import pkg from 'tweetnacl';
const { sign } = pkg;
```

bgins:
Adding a comparison of webntative bundle sizes using `noble-ed25519` and `tweetnacl`:

```
noble-ed25519
- index.umd.min.js: 569.3 kB
- index.umd.min.js.gz: 167.9 kB
- index.umd.min.js.map: 2.4 MB

tweetnacl
- index.umd.min.js: 589.8 kB
- index.umd.min.js.gz: 176.1 kB
- index.umd.min.js.map: 2.5 MB
```
`noble-ed25519` is smaller by ~20.5 kB and ~8.2 kB when gzipped.

bgins:
Fixed a few things on this PR to get it ready.

Merging this PR does not fulfill the original motivation of supporting Create React App because we have other bundler issues that prevent it. But this change is still worthwhile because:

- `tweetnacl` has been audited and has seen much more real-world use than `noble-ed25519`
- It supports a wider range of environments because `tweetnacl` does not use big integer literals

The tradeoffs are basically larger bundle size as noted in a comment above.

Note that `noble-ed25519` was our only dependency with big integer literals, and we can set our tsconfig target to `ES2018` with that removed.

Our `ed25519.browser.test.ts` test suite it passing, and I think that beyond that, we can trust that these libraries are verifying signatures equivalently.

yarn.lock
```diff
@@ -4,31 +4,31 @@
 
 "@assemblyscript/loader@^0.9.4":
   version "0.9.4"
-  resolved "https://registry.yarnpkg.com/@assemblyscript/loader/-/loader-0.9.4.tgz#a483c54c1253656bb33babd464e3154a173e1577"
+  resolved "https://registry.npmjs.org/@assemblyscript/loader/-/loader-0.9.4.tgz"
```
matheus23, line 7: I wonder whats up with these changes. Most likely nothing to worry about, though! (And definitely doesn't prevent merging :)

tsconfig.json
```diff
@@ -1,7 +1,7 @@
 {
   "compilerOptions": {
     "moduleResolution": "node",
-    "target": "ES2020",
+    "target": "ES2018",
```
matheus23, line 4: :tada:

matheus23 (approved):
Looking good! Quite a simple change for what we get out of it. Ship it :) :ship: :shipit:

walkah (approved):
🚢  I agree with @matheus23 nice wins for a relatively minor change 👍

yarn.lock
bgins: Yeah, that's odd. I sometimes use `npm` instead of `yarn` out of habit, but doesn't seem like that would update `yarn.lock`. :thinking:

PR #280, opened 2021-08-24
Secure math for spiral ratchet

expede:
A similar PR may already be submitted!
Please search among the [Pull request](../) before creating one.

Thanks for submitting a pull request! Please provide enough information so that others can review your pull request:

For more information, see the `CONTRIBUTING` guide.

## Summary

* [ ] Bug 1
* [ ] Bug 2
* [ ] Feature 1
* [ ] Feature 2
* [ ] Breaking changes

Explain the **motivation** for making this change. What existing problem does the pull request solve?

## Test plan (required)

Demonstrate the code is solid. Example: The exact commands you ran and their output, screenshots / videos if the pull request changes UI.

## Closing issues

Fixes #

## After Merge

### DIFF
--- a/src/fs/data/private/spiralratchet.test.ts
+++ b/src/fs/data/private/spiralratchet.test.ts
@@ -61,6 +61,7 @@ describe("the spiral ratchet module", () => {
       context("small change", async () => test(8, 0, 0))
       context("medium change", async () => test(450, 0 ,0))
       context("large change", async () => test(70000, 0, 0))
+      context("huge change", async () => test(999999, 0, 0))
     })
 
     context("near rollover point", async () => {
@@ -68,6 +69,7 @@ describe("the spiral ratchet module", () => {
       context("small change c", async () => test(8, 255, 255))
       context("medium change", async () => test(450, 255, 255))
       context("large change", async () => test(70000, 255, 255))
+      context("huge change", async () => test(999999, 255, 255))
     })
 
     context("prop change", async () => {
--- a/src/fs/data/private/spiralratchet.test.ts
+++ b/src/fs/data/private/spiralratchet.test.ts
@@ -4,56 +4,63 @@ import * as fc from "fast-check"
 import * as ratchet from "./spiralratchet.js"
 import { getCrypto } from "./context.js"
 
-
 describe("the spiral ratchet module", () => {
 
   const ctx = getCrypto()
 
- //  it("has the property that complementing twice will always yield the same", () => {
- //    fc.assert(
- //      fc.property(fc.uint8Array(), arr => {
- //        expect(new Uint8Array(ratchet.complement(ratchet.complement(arr.buffer))))
- //          .toEqual(arr)
- //      })
- //    )
- //  })
+  describe('complement', async () => {
+    it("has the property that complementing twice will always yield the same", () => {
+      fc.assert(
+        fc.property(fc.uint8Array(), arr => {
+          expect(new Uint8Array(ratchet.complement(ratchet.complement(arr.buffer))))
+            .toEqual(arr)
+        })
+      )
+    })
 
- //  it("has the property that complementing once will always yield another value", () => {
- //    fc.assert(
- //      fc.property(fc.uint8Array({ minLength: 1 }), arr => {
- //        expect(new Uint8Array(ratchet.complement(arr.buffer)))
- //          .not.toEqual(arr)
- //      })
- //    )
- //  })
-
-//  it("has the property next65536Epoch = 256 * next256Epoch", async () => {
-//    const spiral = await ratchet.setup(ctx)
-//    expect(await ratchet.next65536Epoch(spiral, ctx))
-//      .toEqual(await iterateAsync(spiral, s => ratchet.next256Epoch(s, ctx), 256))
-//  })
-//
-//  it("has the property next256Epoch = 256 * inc", async () => {
-//    const spiral = await ratchet.setup(ctx)
-//    expect(await ratchet.next256Epoch(spiral, ctx))
-//      .toEqual(await iterateAsync(spiral, s => ratchet.inc(s, ctx), 256))
-//  })
-//
-//  it("has the property next65536Epoch = 65536 * inc", async () => {
-//    const spiral = await ratchet.setup(ctx)
-//    expect(await ratchet.next65536Epoch(spiral, ctx))
-//      .toEqual(await iterateAsync(spiral, s => ratchet.inc(s, ctx), 65536))
-//  })
-
-  const iters = 123456
-  it(`has the property incBy ${iters} = ${iters} * inc`, async () => {
-    const spiral = await ratchet.setup(ctx)
-    const positional = await ratchet.incBy(spiral, ctx, iters)
-    const unary = await iterateAsync(spiral, s => ratchet.inc(s, ctx), iters)
-
-    expect(positional).toEqual(unary)
+    it("has the property that complementing once will always yield another value", () => {
+      fc.assert(
+        fc.property(fc.uint8Array({ minLength: 1 }), arr => {
+          expect(new Uint8Array(ratchet.complement(arr.buffer)))
+            .not.toEqual(arr)
+        })
+      )
+    })
   })
 
+  describe('next65536Epoch', async () => {
+    it("has the property that next65536Epoch rounds up to the next large zero", async () => {
+      const spiral = await ratchet.setup(ctx)
+      const slow = await iterateAsync(spiral, s => ratchet.next256Epoch(s, ctx), 256 - spiral.mediumCounter)
+      const fast = await ratchet.next65536Epoch(spiral, ctx)
+      check(slow, fast)
+    })
+  })
+
+  describe('next256Epoch', async () => {
+    it("rounds up to the next medium zero", async () => {
+      const spiral = await ratchet.setup(ctx)
+      const fast = await ratchet.next256Epoch(spiral, ctx)
+      const slow = await iterateAsync(spiral, s => ratchet.inc(s, ctx), 256 - spiral.smallCounter)
+      check(fast, slow)
+    })
+  })
+
+  describe("incBy", async () => {
+    const test = (iters: number) =>  {
+      it(`has the property incBy ${iters} = ${iters} * inc`, async () => {
+        const spiral = await ratchet.setup(ctx) // , str2ab("hello world"), 255, 255)
+        const positional = await ratchet.incBy(spiral, ctx, iters)
+        const unary = await iterateAsync(spiral, s => ratchet.inc(s, ctx), iters)
+        check(positional, unary)
+      })
+    }
+
+    context("no change", async () => test(0))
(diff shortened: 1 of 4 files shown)


src/fs/data/private/spiralratchet.ts
```diff
@@ -11,13 +11,13 @@ interface SpiralRatchet {
 }
 
 
-export async function setup(ctx: EncryptionContext): Promise<SpiralRatchet> {
+export async function setup(ctx: EncryptionContext, seed?: ArrayBuffer, ffSmall?: number, ffMedium?: number): Promise<SpiralRatchet> {
```
expede, line 14: This is here to make it easier to tests specific values

src/fs/data/private/spiralratchet.ts
```diff
@@ -43,46 +44,91 @@ async function zero({ webcrypto, crypto }: EncryptionContext): Promise<SpiralRat
...
   return Object.freeze({
     ...ratchet,
-    small: await sha(webcrypto, ratchet.small),
+    small: await sha(webcrypto, ratchet.small), // NOTE: this uses the input ratchet
     smallCounter: ratchet.smallCounter + 1,
   })
 }
 
-export async function add256(ratchet: SpiralRatchet, ctx: EncryptionContext): Promise<SpiralRatchet> {
+export async function next256Epoch(ratchet: SpiralRatchet, ctx: EncryptionContext): Promise<SpiralRatchet> {
```
expede, line 61: This was the big change. `add256` didn't change the smaller values, even though they have to change (based on the _input_ ratchet, not the updated / zeroed out one)

src/fs/data/private/spiralratchet.ts
```diff
@@ -43,46 +44,91 @@ async function zero({ webcrypto, crypto }: EncryptionContext): Promise<SpiralRat
...
+    smallCounter: 0
   })
 }
 
-export async function add65536(ratchet: SpiralRatchet, { webcrypto }: EncryptionContext): Promise<SpiralRatchet> {
+export async function next65536Epoch(ratchet: SpiralRatchet, ctx: EncryptionContext): Promise<SpiralRatchet> {
+  return await zero(ctx, ratchet.large)
+}
+
+export async function incBy(ratchet: SpiralRatchet, ctx: EncryptionContext, n: number, smCarry?: number): Promise<SpiralRatchet> {
```
expede, line 83: The main event. Increment by an arbitrary number.

src/fs/data/private/spiralratchet.ts
matheus23: Ok, so I think I need to understand this a little better.
I was thinking of the ratchet like an at-least 3-digit number. Let's simplify this and say it's base-10 instead of base-256.
So in base-10, add256 becomes add10.
So, e.g. 123 + 10 = 133.
Why should the least significant digit change when I add 256?
I assume my mental model of the ratchet is wrong. It's not like the ratchet I have in mind wouldn't work as a counter, so I assume it's not (as) secure. Is that what it is, and if so, why?
Trying to learn & understand.

matheus23: Is there a test case that would've failed with the old version? That might illustrate the problem for me very clearly (and maybe make a good test case generally :) )

src/fs/data/private/spiralratchet.ts
matheus23, line 14:
```suggestion
export async function setup(options: EncryptionContext & { seed?: ArrayBuffer; ffSmall?: number; ffMedium?: number }): Promise<SpiralRatchet> {
```

This is the pattern I go for usually. Originally All the `*Context` types were called `*Options`, but since they were often mandatory at that point, I called them `Context` instead. But `seed`, `ffSmall` and `ffMedium` is exactly what I'd put in the options object :)
Also has the bonus of naming the arguments. So e.g. you wouldn't mix up `ffSmall` and `ffMedium` on the call site.

src/fs/data/private/spiralratchet.ts
```diff
@@ -11,13 +11,13 @@ interface SpiralRatchet {
...
 
 
-export async function setup(ctx: EncryptionContext): Promise<SpiralRatchet> {
+export async function setup(ctx: EncryptionContext, seed?: ArrayBuffer, ffSmall?: number, ffMedium?: number): Promise<SpiralRatchet> {
   const { webcrypto, crypto } = ctx
   let [mediumSkip, smallSkip] = crypto.getRandomValues(new Uint32Array(2))
-  mediumSkip = mediumSkip % 255
-  smallSkip = smallSkip % 255
+  mediumSkip = ffMedium === undefined ? mediumSkip % 256 : ffMedium
+  smallSkip = ffSmall === undefined ? smallSkip % 256 : ffSmall
```
matheus23, line 18: This doesn't adhere to the spec here. The spec says:
```hs
setup :: SpiralRatchet
setup = do
  largePre   <- random -- ratchet seed
  mediumSkip <- randomBetween 0 254
  smallSkip  <- randomBetween 0 254
```
So, `randomBetween 0 254`. Is the spec wrong? (To be fair, I should've mentioned that I was confused by that in the spec.

If so, then we can simplify this code:
```ts
let mediumSkip = crypto.getRandomValues(new Uint8Array(2))
```
No need to `% 256` if they're bytes then :+1: 

Very minor nitpick: I like to make all defined-ness checks `something == null`. It checks both whether it's `undefined` or `null`.

src/fs/data/private/spiralratchet.ts
```diff
@@ -28,10 +28,11 @@ export async function setup(ctx: EncryptionContext): Promise<SpiralRatchet> {
   })
 }
 
-async function zero({ webcrypto, crypto }: EncryptionContext): Promise<SpiralRatchet> {
-  const largePre = crypto.getRandomValues(new Uint8Array(32)).buffer
-  const medium = await sha(webcrypto, complement(largePre))
-  const small = await sha(webcrypto, complement(medium))
+async function zero({ webcrypto, crypto }: EncryptionContext, seed?: ArrayBuffer): Promise<SpiralRatchet> {
```
matheus23, line 31: Same thing here
```suggestion
async function zero({ webcrypto, crypto, seed }: EncryptionContext & { seed?: ArrayBuffer }): Promise<SpiralRatchet> {
```

src/fs/data/private/spiralratchet.test.ts
```diff
@@ -54,3 +92,31 @@ async function iterateAsync<T>(initial: T, f: (obj: T) => Promise<T>, n: number)
   }
   return obj
 }
+
+function ab2str(buf: ArrayBuffer): string {
+    var dataView = new DataView(buf);
+    var decoder = new TextDecoder('utf-8');
+    return decoder.decode(dataView);
+}
```
matheus23, line 100:
```suggestion
function ab2str(buf: ArrayBuffer): string {
    return new TextDecoder("utf-8").decode(buf); // but utf-8 is also the default
}
```

src/fs/data/private/spiralratchet.test.ts
```diff
@@ -54,3 +92,31 @@ async function iterateAsync<T>(initial: T, f: (obj: T) => Promise<T>, n: number)
...
+
+function str2ab(str: string): ArrayBuffer {
+  var buf = new ArrayBuffer(str.length*2); // 2 bytes for each char
+  var bufView = new Uint16Array(buf);
+  for (var i=0, strLen=str.length; i < strLen; i++) {
+    bufView[i] = str.charCodeAt(i);
+  }
+  return buf;
+}
```
matheus23, line 109:
```suggestion
function str2ab(str: string): ArrayBuffer {
  return new TextEncoder().encode(str).buffer
}
```
:grimacing: 
I'm not 100% sure. Was this function encoding into utf-16? Then my suggesting will do something different.
I'd propose to get rid of this anyway, by simply generating uint8arrays in fast-check and calling `.buffer` on those.
Leaving this here just FYI though :D

src/fs/data/private/spiralratchet.test.ts
```diff
@@ -4,47 +4,85 @@ import * as fc from "fast-check"
...
+      context("small change c", async () => test(8, 255, 255))
+      context("medium change", async () => test(450, 255, 255))
+      context("large change", async () => test(70000, 255, 255))
+      context("huge change", async () => test(999999, 255, 255))
+    })
 
+    context("prop change", async () => {
+      it("works with any number of iterations", async () => {
+        fc.assert(fc.asyncProperty(fc.nat(), fc.string(), async (iters: number, seed: string): Promise<void> => {
+          const spiral = await ratchet.setup(ctx, str2ab(seed))
```
matheus23, line 78:
```suggestion
        fc.assert(fc.asyncProperty(fc.nat(), fc.uint8Array(), async (iters: number, seed: Uint8Array): Promise<void> => {
          const spiral = await ratchet.setup(ctx, seed.buffer)
```

matheus23 (approved):
Thanks for working on this brooke!

I just commented some stuff just for knowledge sharing purposes.

src/fs/data/private/spiralratchet.ts
matheus23: Ok I think I get it now :)

The problem is, we can't have 123 + 10 = 133, because then it'd not be 100% backwards secret! If someone knows 133, then they know 1/3 of the ratchet 10 iterations ago. They'll know it was **3.
Even worse, if we have the second digit do the same thing. So if they know all the second and third digits (i.e. they know *00-*99), then they can go back and forth between any n00 and n99.

We could make a test out of this, I think I'll try that. It'd just check that none of the previous ratchet values has occurred before, not even partially.

matheus23:
Ah and one more trick:

```ts
function canonicalize(spiral: SpiralRatchet) { return JSON.parse(JSON.stringify(spiral)) }

expect(canoncialize(spiral)).toEqual(canoncialize(spiral2))
```

matheus23:
Actually, that doesn't work:
```
> JSON.stringify(initial)
'{"large":{},"medium":{},"mediumCounter":155,"small":{},"smallCounter":18}'
```

matheus23:
This does work:
```ts
function canonicalize(spiral: ratchet.SpiralRatchet): unknown {
  return JSON.parse(JSON.stringify(spiral, (_, value) =>
    value instanceof ArrayBuffer
      ? uint8arrays.toString(new Uint8Array(value), "base64url")
      : value
  ))
}
```

matheus23:
Another BIG gotcha, if we have async tests using fast-check, we need to await them.
I.e.
```js
it("tests something async", async () => {
  fc.assert(fc.asyncProperty(...))
})
```
**WIll not test correctly**. It'll start the promise but then return immediately, making mocha think all is fine. Exceptions won't be caught.

We need to do this:
```js
it("tests something async", async () => {
  await fc.assert(fc.asyncProperty(...))
})
```

I'm adding [a linter rule for this](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-floating-promises.md). That's a serious foot gun :eyes:

PR #281, opened 2021-08-25
Add eslint rule @typescript-eslint/no-hanging-promises to prevent not awaiting fast check assertions

matheus23:
I recently wrote a test case in another branch that didn't await the fast-check assertion:

```ts
it("has some asynchronous property", async () => {
  fc.assert(fc.asyncProperty(...))
}
```
But that' won't actually fail if the property fails. In fact it returns almost immediately (synchronously). So it doesn't think there's anything going wrong in the test, it just runs.

The correct version is this:
```ts
it("has some asynchronous property", async () => {
  await fc.assert(fc.asyncProperty(...))
}
```

I was so shocked at how dangerous this is that I decided to look for a rule preventing this.

There *are* use cases for having hanging promises, e.g. if you want to intentionally run an asynchronous action. In these cases, prepend your hanging promise with `void` like you'd do it with `await`:

```ts
async function hugeComputationWithSideEffect(n: number): Promise<void>

// usage:
void hugeComputationWithSideEffect(0) // hanging promise
void hugeComputationWithSideEffect(1) // hanging promise
```

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -1,6 +1,9 @@
 module.exports = {
   root: true,
   parser: "@typescript-eslint/parser",
+  parserOptions: {
+    project: `./tsconfig.eslint.json`
+  },
   plugins: [
     "@typescript-eslint",
   ],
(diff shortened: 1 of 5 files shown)


bgins (approved):
Awesome, great to make sure we are handling these cases correctly! :100:

matheus23:
When awaiting one test, it seemed to never finish. Not sure what was going on.

I've been bothered by our own implementation of `Promise.any` anyway, so I removed it. Am anxious that I'm messing something up completely by doing this small change (once again :P ). Would appreciate another close look at this (@bgins ?).

I worry especially that (1) adding an "es2021" entry to `tsconfig.json` messes something up. It should just provide polyfills for `Promise.any` and (2) that `Promise.any` doesn't work like `race` used to/is not the thing we needed in that place.

PR #284, opened 2021-09-02
Improve login flow resilience against extensions

matheus23:
This happens to me from time to time when logging in with fission apps:
image

I identified the cause of this issue to be my "10ten" extension. It would `postMessage` from the `auth.fission.codes` origin with a message like `10ten(X)ping`.

Webnative would think that a message from the auth lobby must be login info and try to parse that as JSON.

I should've really thought of the whole `postMessage` thing like a distributed system. We just don't know anything about the messages we're getting.

So, the assumptions I'm making now is that we won't get more than 9 unrelated messages and we'll the message we care about within 10 seconds.

With the 10ten extension enabled, I don't have any problems logging in anymore.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -139,11 +139,11 @@ export async function initialise(
     const newUser = url.searchParams.get("newUser") === "t"
     const username = url.searchParams.get("username") || ""
 
-    await importClassifiedInfo(
+    await retry(async () => importClassifiedInfo(
       authorised === "via-postmessage"
         ? await getClassifiedViaPostMessage()
         : await ipfs.cat(authorised) // in any other case we expect it to be a CID
-    )
+    ), { tries: 10, timeout: 10000, timeoutMessage: "Trying to retrieve authentication secrets from the auth lobby timed out after 10 seconds." })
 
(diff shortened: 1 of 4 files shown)


src/index.ts
```diff
@@ -388,6 +388,10 @@ async function getClassifiedViaPostMessage(): Promise<string> {
       window.addEventListener("message", listen)
 
       function listen(event: MessageEvent<string>) {
+        if (new URL(event.origin).host !== new URL(setup.endpoints.lobby).host) {
+          console.log(`Got a message from ${event.origin} while waiting for login credentials. Ignoring.`)
+          return
+        }
         window.removeEventListener("message", listen)
         if (event.data) {
           resolve(event.data)
```
icidasset, line 397: In addition to this, we could also easily check if this is the correct data?
Check if it's JSON and if the object has a `sessionKey` property.

icidasset (approved):
One note, looks good otherwise.

src/index.ts
matheus23: Yeah. I like this. I've added a commit for that :+1:

icidasset (approved):

PR #285, opened 2021-09-07
Store exchange key in well known location

icidasset:
Adds
```ts
fs.addPublicExchangeKey()
```
which will add the public part of your exchange key as a DID to
```
/public/.well-known/exchange/YOUR_DID
```
with the contents
```
{}
```

Also adds the `fs.hasPublicExchangeKey()` function to check if it's there.


matheus23 (approved):
PR is great, just wondering about the design itself:
Why did we go for writing a file with dummy content instead of writing a directory?
As of right now I think a directory has more benefits. Especially when it comes to reconciliation in the future: Let's say two clients want to attach some information to some exchange key. They can do so just by writing a file in the respective exchange key's folder, concurrently. As long as the file names don't match, reconciliation will take care of that.

In contrast, if we store the exchange key as a file, when two clients change the exchange key file at the same time, then only one of the changes will get applied: The change with the bigger root CID.

icidasset:
A directory works too yeah, I'll change it 👀

matheus23 (approved):

PR #286, opened 2021-09-13
Update keystore-idb

icidasset:
Implements the 0.15 keystore-idb changes.
Still uses an alpha release, will replace before merging.


matheus23 (approved):
Nice!
I think we should add a changelog item for this change. Just in case anyone is wondering why `(await webnative.keystore.get()).publicReadKey()` isn't working anymore.
Let's get this into the alpha release

PR #288, opened 2021-09-17
Use newer js-ipfs proxy libraries

icidasset:
We can't seem to get our js-ipfs PRs merged in, so this is an attempt to use the proxy libraries without our fixes.

## PRs

https://github.com/ipfs/js-ipfs/pull/3575

Adds support for pinning to the proxies.
We can probably ignore this, since js-ipfs doesn't do pinning anyway (turns out 😆)

https://github.com/ipfs/js-ipfs/pull/3573

Fixes issue with buffers.
We most likely need this one.

## Status

Currently this code gives an error regarding CIDs.
I vaguely remember this was an issue back when I worked on those PRs.
It's definitely an error in one of the proxy libraries.

Screenshot 2021-09-17 at 21 58 43

But maybe this is some inconsistency with all the different CID objects/classes floating around. We mostly use strings, there's the CID class from the `cids` package, then the one from `multiformats` and possibly another one from the ipfs packages (depending on the version).


matheus23:
> But maybe this is some inconsistency with all the different CID objects/classes floating around. We mostly use strings, there's the CID class from the cids package, then the one from multiformats and possibly another one from the ipfs packages (depending on the version).

This might be due to js-ipfs's change to a new CID library. They've now moved on to the `multiformats/cid` as representation for CIDs everywhere and deprecated the `cids` package. We'd need up update js-ipfs generally for that though.
I've done that in the wnfs2 branch and I think I could backport that change for v1 as well. Should I do that?

matheus23:
> since js-ipfs doesn't do pinning anyway

I think you meant to say "doesn't garbage collect", right?

icidasset:
> I've done that in the wnfs2 branch and I think I could backport that change for v1 as well. Should I do that?

Nah it's fine, we'll power through until your `wnfs2` branch is merged 😜 

> I think you ment to say "doesn't garbage collect", right?

Yeah, for me pinning is synonymous to "don't garbage collect".

src/setup/internal.ts
```diff
@@ -17,5 +17,5 @@ export const setup = {
     user: "fission.name"
   },
 
-  shouldPin: true,
+  shouldPin: false,
```
matheus23, line 20: I think we can just remove this now, right?
I'm not 100% sure. There might be a use-case for pinning for users who're providing their own IPFS instance.
... But at the same time I think there's noone who's actually doing that. :smile:

icidasset: Can't hurt to keep it around I guess? In case js-ipfs does turn on GC, like how it's supposed to work 😅

src/ipfs/util.ts
```diff
@@ -9,7 +10,9 @@ type RawDAGLink = {
 }
 
 const rawToDAGLink = (raw: RawDAGLink): DAGLink => {
-  return new dagPb.DAGLink(raw.Name, raw.Tsize, raw.Hash)
+  const h = raw.Hash
+  const c = new OldCID(h.version, h.code, h.multihash.bytes)
+  return new dagPb.DAGLink(raw.Name, raw.Tsize, c)
```
icidasset, line 15: This is a bit dumb, but I didn't want to put too much work in here, since this is getting replaced soon anyway by WNFS v2.

tests/helpers/in-memory-ipfs.ts
```diff
@@ -1,8 +1,23 @@
...
+import tempDir from "ipfs-utils/src/temp-dir.js"
+import type { IPFS } from "ipfs-core"
+
 
 export async function createInMemoryIPFS(): Promise<IPFS> {
+  const dir = tempDir()
+  fs.mkdirSync(dir)
+
+  const memoryDs = new MemoryDatastore()
+  const memoryBs = new MemoryBlockstore()
```
icidasset, line 19: @matheus23 Bit of double effort here, but needed to replace this to get the tests working with the new `ipfs-core`.

icidasset:
Ready for review. Goes hand in hand with this lobby PR: https://github.com/fission-suite/auth-lobby/pull/92 Which I'll need to deploy to staging to get all the tests passing.

tests/helpers/in-memory-ipfs.ts
matheus23: Ah yeah damnit. Should've told you I've battled with that in the wnfs2 branch. There I had to update ipfs-core to work with the newer IPLD libs. In case you hit something else while updating ipfs-core, take a look at that branch. There might be something useful :)

tests/helpers/loadCAR.ts
```diff
@@ -13,7 +13,7 @@ export async function loadCAR(filepath: string, ipfs: IPFS): Promise<{ roots: CI
     const blockIterator = await CarBlockIterator.fromIterable(inStream)
     for await (const block of blockIterator) {
       cids.push(block.cid)
-      await ipfs.block.put(block.bytes, { cid: block.cid })
+      await ipfs.block.put(block.bytes, { version: 1, mhtype: "sha2-256" })
```
matheus23, line 16: I think this code has issues, because it'll strip away the IPLD codec from all CIDs and turn all blocks to "raw" blocks implicitly (this code is probably from the wnfs2 branch).

We should totally get rid of this whole thing and just use the newly implemented [`ipfs.dag.import` api](https://github.com/ipfs/js-ipfs/blob/master/docs/core-api/DAG.md#ipfsdagimportsource-options).

icidasset: Aight! I'll see if I can use that `import` function 👀

PR #290, opened 2021-09-28
Rewire keystore-idb

icidasset:
**We have some bundling issues on snowpack, parcel, create-react-app and possibly some others. All of those don't support the `exports` setting in package.json (or maybe not nested ones).**

I made a repo to test webnative in various bundlers:
https://github.com/fission-suite/webnative-bundler-test

This PR uses the old/real path to the keystore-idb files to import them, whilst still having keystore-idb support `exports`. With this webnative works in old and new bundlers. I added the same `"./lib/*": "./lib/*"` line to the exports map as we did with keystore-idb. That allows bundlers to use the exports map, but also the `lib` path.

Made an alpha release to test this in said repo: `0.28.1-alpha`


bgins (approved):
Tested these changes out with https://github.com/fission-suite/react-todomvc and they are working great! 💯 🎉

matheus23 (approved):

PR #307, opened 2021-10-26
Prepend information about the encryption algorithm to private filesystem blocks & switch to `AES-GCM`

### LINKED ISSUE #289: Use AES-GCM everywhere
Unless we have a specific reason to, we should be using AES-GCM across the board. Safari is giving me this warning:

Screen Shot 2021-09-27 at 15 08 11

There's some open questions about if we GCM is overkill when we have integrity hashes from IPFS, but it can't hurt in the meantime. We've discussed this internally previously, and I thought we had already switched over to GCM. We should double check.

Possible that it's just picking up on this with a naive scan of the source? https://github.com/fission-suite/webnative/blob/26dcae9e4a8f0145e62648984ae29dccb374a02b/src/lobby/blocklist.ts#L73-L77

matheus23:
Closes #289 

Depends on #296


src/ipfs/encoded.ts
```diff
@@ -15,12 +17,33 @@ export const add = async (content: FileContent, key: Maybe<string>): Promise<Add
...
   const encoded = cbor.encode(normalized)
-  const toAdd = isJust(key) ? await crypto.aes.encrypt(encoded, key) : encoded
+  
+  if (!isJust(key)) {
+    return basic.add(encoded)
+  }
+  
+  const alg = SymmAlg.AES_GCM
+  const cip = await crypto.aes.encrypt(encoded, key, alg)
+  const toAdd = cbor.encode({ alg, cip })
```
matheus23, line 27: The main attraction:
Instead of storing the encrypted blocks directly, we store a cbor block that contains the cipher in `cip` and the algorithm in `alg`.

matheus23:
> This PR is on hold until a sufficient amount of webnative apps have moved on to at least version 0.29.1 (version checking).

Actually, I take that back. We should do the following:
* Release this (possibly together with other breaking changes, e.g. symlinks) as 0.30 alpha
* Test this on staging
* <ensure apps have moved on to 0.29 meanwhile>
* Release this as 0.30 non-alpha.

icidasset (approved):
Release plan sounds good as well.
I'll merge symlinks after this.

Issue #318, opened 2021-10-29
Webcrypto import errors in 0.29.1


bgins:
# Summary

## Problem

`webcrypto` cannot be imported or is `undefined` with some bundlers/build tools with webnative `0.29.1`.

### Impact

Some applications will not be able to use webnative at `0.29.1`, depending on the build system or bundler they use if any.

## Solution

Unknown. The issue appears to be related to the change over to `one-webcrypto` and there may be something we can change there.

# Detail

**Describe the bug**

Parcel 2, Webpack 5, and Create React App report errors in the console. (See the screenshots below.)

**To Reproduce**
1. Clone https://github.com/fission-suite/react-todomvc
2. Upgrade webnative to 0.29.1 on each branch
3. `npm start` for local dev and `npm run build` for a build in each
3. Vite should work, but the `parcel`, `webpack`, and `main` branches should fail (main is CRA)

**Expected behavior**

No errors in the console. The app works and can do TodoMVC things.

**Screenshots**

CRA local development

<img width="960" alt="cra-dev" src="https://user-images.githubusercontent.com/7957636/139485300-b0d12df3-9698-4c69-a50c-6f22955c4599.png">

CRA build (fails at build time)

<img width="960" alt="cra-build" src="https://user-images.githubusercontent.com/7957636/139485374-0ba45219-e5ca-484c-b523-4d72f6acbc7b.png">

Parcel 2 local development

<img width="960" alt="parcel-dev" src="https://user-images.githubusercontent.com/7957636/139485405-e74ae28b-c17b-4250-94fc-43b0491f6521.png">

Parcel 2 build

<img width="960" alt="parcel-build" src="https://user-images.githubusercontent.com/7957636/139485426-64d06dfc-4395-477b-ae2d-3b1bed27ca51.png">

Webpack 5 local development and build

<img width="937" alt="webpack-dev-and-build" src="https://user-images.githubusercontent.com/7957636/139485497-a3700267-cc45-4694-9fa5-4f13f632162c.png">

matheus23:
I'm able to fix issues with parcel and webpack with this commit: https://github.com/fission-suite/one-webcrypto/commit/a95b889887ffa557cdaddd2f19cc14a4bec79e56

CRA remains a problem. I'm not sure why it can't import webcrypto at all :zany_face:

matheus23:
"Progress" on CRA:

[...]

PR #321, opened 2021-11-02
Bump one-webcrypto to 1.0.3 and keystore-idb to 0.15.4

matheus23:
Allows webnative to be used in more environments.

Main action is a version bump of one-webcrypto to 1.0.3 and keystore-idb to 0.15.4 (which again is only due to a version bump of one-webcrypto).

Discussed in #318

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+### v0.29.2
+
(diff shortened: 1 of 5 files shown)


bgins (approved):
Looks good! Let's bring it in.

matheus23:
Fixes #318

PR #326, opened 2021-11-17
Remove debug console.log

matheus23:
Just a console.log that slipped through.

It prints the variable "classifiedInfo", but that's not dangerous. The contents of that variable are encrypted with a non-extractable key.

Still, not necessary to have that printing.


matheus23:
I'll merge this if nobody minds :eyes:

PR #330, opened 2021-11-30
App owned logins

### LINKED ISSUE #302: Move account creation to webnative
* [x] Move from auth lobby to webnative
* [x] Provide clean high-level API
  * [ ] Configurable registration endpoint (not just the hostname, default to Fission)
  * [ ] Ability to customize username blocklist
* [ ] Include downloading backup keys & UCAN (simple version, not BLS yet)
* [ ] Tests

### LINKED ISSUE #316: Generalize Username Resolution
Make the REST endpoint configurable (I believe we still get this via REST today?)

walkah:
This PR introduces dependency injection for initialization and account registration (along with the various username validation features). This allows apps to override these and potentially use alternative authentication backends / mechanisms.

It also introduces a "local" version that allows an app to init/register accounts in a self-contained manner like this:

```js
import { LOCAL_IMPLEMENTATION } from "webnative/auth/local";
setDependencies(LOCAL_IMPLEMENTATION);
wn.initialise({ loadFileSystem: false })
  .then((result) => {
    const { authenticated, username } = result;
    if (authenticated) {
      console.log(`Logged in as: ${username}`)
    }
  })
  .catch((e) => console.error(e));
```

Closes #316 
Closes #302

### DIFF
--- a/src/auth/local.ts
+++ b/src/auth/local.ts
@@ -1,0 +1,37 @@
+import { State } from "../index.js"
+import { createAccount } from "../lobby/index.js"
+import * as user from "../lobby/username.js"
+import * as storage from "../storage/index.js"
+import { USERNAME_STORAGE_KEY } from "../common/index.js"
+
+export const init = async (): Promise<State | null> => {
+  console.log("initialize local auth")
+  return new Promise((resolve) => resolve(null))
+}
+
+export const register = async (options: { username: string; email: string}): Promise<{success: boolean}> => {
+  const { success } = await createAccount(options)
+
+  if (success) {
+    await storage.setItem(USERNAME_STORAGE_KEY, options.username)
+    return { success: true }
+  }
+  return { success: false }
+}
+
+export const isUsernameValid = async (username: string): Promise<boolean> => {
+  return user.isUsernameValid(username)
+}
+
+export const isUsernameAvailable = async (username: string): Promise<boolean> => {
+  return user.isUsernameAvailable(username)
+}
+
+export const LOCAL_IMPLEMENTATION = {
+  auth: {
+    init,
+    register,
+    isUsernameValid,
+    isUsernameAvailable
+  }
+}
--- a/src/setup/dependencies.ts
+++ b/src/setup/dependencies.ts
@@ -1,5 +1,7 @@
 import * as browserCrypto from "../crypto/browser.js"
 import * as browserStorage from "../storage/browser.js"
+import * as authLobby from "../auth/lobby.js"
+import { InitOptions, State } from "../index.js"
 
 
 export const DEFAULT_IMPLEMENTATION: Dependencies = {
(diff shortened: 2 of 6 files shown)


src/auth/local.ts
```diff
@@ -0,0 +1,37 @@
+import { State } from "../index.js"
+import { createAccount } from "../lobby/index.js"
+import * as user from "../lobby/username.js"
+import * as storage from "../storage/index.js"
+import { USERNAME_STORAGE_KEY } from "../common/index.js"
+
+export const init = async (): Promise<State | null> => {
+  console.log("initialize local auth")
```
matheus23, line 8: We could switch out `console.log` with the debug logging functions, so they only print with `setup.debug({ enabled: true })`

src/setup/dependencies.ts
```diff
@@ -71,4 +80,10 @@ export interface Dependencies {
     removeItem: (key: string) => Promise<void>
     clear: () => Promise<void>
   }
+  auth: {
+    init: (options: InitOptions) => Promise<State | null>
+    register: (options: { email: string; username: string }) => Promise<{ success: boolean }>
+    isUsernameValid: (username: string) => Promise<boolean>
+    isUsernameAvailable: (username: string) => Promise<boolean>
```
matheus23, line 87: Just a Q for clarification: Why do we abstract these functions? We're using the same implementation for both local auth as well as auth-lobby based auth it seems.
Is this so that someone would in principle be able to implement their own "username database"? If so, is that possible today? I think we'd need email-like identifiers for that, i.e. something like `matheus23@fission.codes` usernames instead of `matheus23`. E.g. in the DNSLink fetching code.
Maybe it doesn't hurt having these abstractions in here today already, but I'm not sure. :thinking:

matheus23 (approved):
Minor stuff only & a small question. Otherwise looking good to me :)
tumbsup

src/setup/dependencies.ts
expede: Yes, this is so that you can plug in any arbitrary lookup method, rather than forcing everyone to make it look like Fission's. For example, web3.storage has this in a big database table that's at a different endpoint.

expede: RE validation: services may have special validation rules for their service (e.g. no curse words in Spanish). I believe that the plan is to provide defaults and let users override them if needed.

> something like matheus23@fission.codes usernames instead of matheus23

That's more of an interoperability question. Blaine calls this style the "Jesus of Nazareth" format, and I agree that it reads very well (in English at least)

expede: (But per service, you can imply the host, in the same way that Discord hides the nonce in your username unless there's a collision on that server)

src/auth/local.ts
expede: > setup.debug({ enabled: true })

This is also generally togglable in the console directly — no need for managing it ourselves

Screen Shot 2021-12-09 at 10 01 27

expede (approved):
LGTM! Bump the version and...

PR #331, opened 2021-12-02
Add private sharing

icidasset:
**Enables sharing of private data.**

_Flow:_
1. Create a share
2. Create a link
3. User opens link and goes to auth lobby
4. User accepts share in lobby
5. Content is available at `private/Shared with me/Sender's username/`

## Top Level API

```ts
/********************
 * Creating a share *
 ********************/
sharePrivate(paths: DistinctivePath[], { sharedBy, shareWith }: { sharedBy?: SharedBy, shareWith: string | string[] }): Promise<ShareDetails>

fs.sharePrivate(
  [ wn.path.file("private", "Photos", "🏞.jpg") ],
  { shareWith: "receiver-username-or-exchange-did" }
)
// { shareId: string, sharedBy: { rootDid: string, username: string } }


/*************************
 * Creating a share link *
 *************************/
shareLink(details: ShareDetails): string

wn.lobby.shareLink(shareDetailsFromTheAbove)
// "https://auth.fission.codes/#/share/sender-username/1/"


/*********************
 * Accepting a share *
 *********************/
acceptShare({ shareId, sharedBy }: { shareId: string, sharedBy: string }): Promise<this>

fs.acceptShare({ shareId: "1", sharedBy: "sender-username" })
// fs


/* Internally calls these methods, which are also public: */

fs.loadShare({ shareId, sharedBy }: { shareId: string, sharedBy: string }): Promise<PrivateTree>
fs.add(path: DistinctivePath, content: FileContent | SoftLink | SoftLink[] | Record<string, SoftLink>, options: MutationOptions = {}): Promise<this>
```

## Other API Additions

```ts
// Utility function to resolve soft links (in addition to PrivateTree.resolveSoftLink)
resolveSymlink(link: SoftLink): Promise<File | Tree | null>

// Utility function to update nested trees (as opposed to writing specialised recursive functions)
async runOnChildTree(node: Tree, relPath: Path, fn: (tree: Tree) => Promise<Tree>): Promise<Tree>
```

## Other Changes

Sends a `sharing=t` query param to the lobby to request access to the `/shared` section of the filesystem.

## Notes

Implementation notes and entire technical flow can be found here:
https://talk.fission.codes/t/initial-private-sharing-implementation/2140/3

This does not break backwards compatibility.


src/fs/root/tree.ts
```diff
@@ -246,6 +275,50 @@ export default class RootTree implements Puttable {
...
+  // SHARING
+  // -------
+
+  async addShares(links: SimpleLink[]): Promise<this> {
+    this.sharedLinks = links.reduce(
+      (acc, link) => ({ ...acc, [link.name]: link }),
+      this.sharedLinks
+    )
+
+    const cid = await ipfs.add(
```
icidasset, line 287: This doesn't actually work, because these links will not be recursively pinned.
This should use `ipfs.dag.put` instead.

src/fs/share.ts
```diff
@@ -0,0 +1,134 @@
...
+    await crypto.aes.encrypt(
+      new TextEncoder().encode( JSON.stringify(indexNode) ),
+      index.key,
+      symmKeyAlgo
+    )
+  )
+
+  // Create share payload
+  const payload = cbor.encode(shareKey.payload({
+    entryIndexCid: indexResult.cid,
```
icidasset, line 93: This entry index is not being pinned by the server because it can't reach it.
We need to find another place to put this CID so the server pins it.

src/fs/protocol/private/namefilter.ts
```diff
@@ -61,7 +61,7 @@ export const toHash = async (filter: BloomFilter): Promise<PrivateName> => {
 }
 
 // saturate a filter (string) to 320 bits
-export const saturate = async (filter: RevisionNameFilter, threshold = SATURATION_THRESHOLD): Promise<SaturatedNameFilter> => {
+export const saturate = async (filter: RevisionNameFilter | BareNameFilter, threshold = SATURATION_THRESHOLD): Promise<SaturatedNameFilter> => {
```
matheus23, line 64: I didn't quite figure out where this was needed from reading the code. Can this be changed back?

matheus23 (commented):
Excited for this :)
Apart from needing a fix for pinning and a couple of small details mostly typescript-related, this looks good to me :+1:

src/fs/share.ts
icidasset: I thought so too but that doesn't seem like a good idea to put them there in their current form, since the `index` contains the AES keys with the pointers to that data (soft links). I guess one could then go through one's MMPT and extract that content?

src/fs/filesystem.ts
```diff
@@ -227,13 +228,49 @@ export class FileSystem {
...
+        || typeCheck.isSoftLinkList(content)
+
+      if (contentIsSoftLinks && destinationIsFile) {
+        throw new Error("Can't add soft links to a file")
+      } else if (!contentIsSoftLinks && pathing.isDirectory(path)) {
+        throw new Error("`add` only accepts file paths when working with regular files")
+      }
+
+      if (destinationIsFile && !contentIsSoftLinks) {
+        return (node as File).updateContent(content as FileContent)
```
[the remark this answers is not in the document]
icidasset: Typescript only seems to infer the types correctly when the `if` conditional surrounding it contains the type check. So if I would say `if (typeCheck.isFile(node)) return node.updateContent()` that would work just fine. But if I remove the type casts here, it won't work.

src/fs/filesystem.ts
```diff
@@ -227,13 +228,49 @@ export class FileSystem {
...
+
+      } else if (contentIsSoftLinks) {
+        const links = Array.isArray(content)
+          ? content
+          : typeChecks.isObject(content)
+            ? Object.values(content) as Array<SoftLink>
+            : [ content ] as Array<SoftLink>
```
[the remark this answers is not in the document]
icidasset: Doesn't work 😞 Error is `Type 'unknown[]' is not assignable to type 'SoftLink[]'`

src/fs/protocol/private/namefilter.ts
icidasset: Will change back, ended up not using this.

src/fs/share.ts
icidasset: Oh no wait, I forgot the index was also encrypted with AES 🤦 
I'll put it in the MMPT 👍

icidasset: Added and now working fine across different accounts ✅

src/fs/filesystem.ts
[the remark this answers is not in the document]
icidasset: We do confirm these have the correct types in `contentIsSoftLinks` a few lines above this. This is more of that weird Typescript type inference.

src/fs/share.ts
expede: @icidasset

Correct, it goes in the private partition 👍 Yes we _definitely_ need the index encrypted!

Also down the road it will make sense to put _unencrypted_ data in the same structure ("unlisted" as opposed to public/private), but that's down the road.

---

@matheus23 

Sure — I have no strong feelings one way or the other. You end up with a CID either way (namefilters point to CIDs), and you may get some speedup _when the link is missing_ by using namefilters... but that's not the typical case when you've had something shared with you. You can think of the shared link as being an assertion that the content is there, but the one layer of indirection can prevent you from spinning forever waiting to resolve a CID that doens't exist.

PR #332, opened 2021-12-09
Don't `put` multiple times and add MMPT warning

icidasset:
This was causing the "Shared with me" folder not to load on my other devices. Because `.put()` was called multiple times on that folder, causing it to be added multiple times to the MMPT with different CIDs. `.put()` produces different CIDs every time because the AES encryption is done with a random `iv` value, and the MMPT doesn't allow addition of the same key/namefilter multiple times (it will keep the first, original, value).

- Removes multiple `put` calls
- Adds MMPT warning in case this happens elsewhere

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "webnative",
-  "version": "0.30.0-alpha7",
+  "version": "0.30.0-alpha8",
   "description": "Fission Webnative SDK",
   "keywords": [
     "WebCrypto",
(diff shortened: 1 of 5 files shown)


matheus23 (approved):
:+1: 
If I understand it correctly, then `updateDirectChild` doesn't need `parent` to update anymore, because `chain` is essentially an array of nodes which are related to each other, i.e index 0 is parent of index 1, which is parent of index2, etc., right? So, we're running `updateDirectChild` on all of them anyway?

To me it isn't immediately clear why it's fine passing `null`. Maybe we can put a short comment there?

PR #335, opened 2021-12-10
App owned: account linking

### LINKED ISSUE #305: Move account linking to webnative
Move from https://github.com/fission-suite/auth-lobby/blob/develop/src/Javascript/Main.js (some more logic lives in the Elm code as well)

* [ ] Open unencrypted channel
  * [ ] Split “open channel” from “bootstrap secure session” if not already like that
  * [ ] Make URL endpoint configurable
* [ ] Migrate device linking flow to webnative
  * [ ] Open channel, bootstrap secure session
  * [ ] Already broken down by stage in the auth lobby code
    * [ ] Make each message into its own function
    * [ ] Stateful flow (it's a state machine — track how far through, reject wrong steps, &c)

walkah:
_@bgins here 👋. I took over this feature and submit the PR described below._

---

## Summary

This PR brings the account linking functionality currently in the Fission Auth Lobby into webnative. 

* [x] Adds account linking
* [x] Adds an event emitter implementation
* [x] Adds `console.warn` to `debug` module
* [x] Adds a `Result` type to our collection of common types

This PR aims to make account linking available to any app that uses webnative. Before this PR, account linking was only available in the Fission Auth Lobby.

### API surface

The API exposes account linking providers and requestors. A provider is an authenticated device with the capability to delegate account access. A requestor is an unauthenticated device that would like access.

_Note: Internally, we refer to `providers` as `producers` and `requestors` as `consumers`. We rename these in the API with more intuitive names._

_Follow-up note: We decided to use the names `producer` and `consumer` everywhere. See discussion below._

Providers are event emitters with a `cancel` function:

```typescript
import { auth } from 'webnative'

const provider = await auth.createProvider({ username })

provider.on('challenge', ({ pin, confirmPin, rejectPin }) => {
  // Request pin confirmation, then call confirmPin or rejectPin
});

provider.on('link', ({ approved, username }) => {
  // Indicate to user whether linking was approved or not
});

provider.on("done", () => {
  // Clean up after account linking
});

// Cancel linking
provider.cancel();
```

A provider immediately listens for account linking requests when created and continues to listen until `cancel` is called. Providers can link multiple requestors without needing to be recreated.

Requestors are also event emitters with a `cancel` function:

```typescript
import { auth } from 'webnative'

const requestor = await auth.createRequestor({ username })
  
requestor.on('challenge', ({ pin }) => {
  // Display pin
});

requestor.on('link', ({ approved, username }) => {
  // Indicate to user whether linking was approved or not
});

requestor.on("done", () => {
  // Clean up after account linking
});

// Cancel linking
requestor.cancel();
```

A requestor attempts to link on creation. Unlike providers, requestors are automatically terminated when they complete linking or receive a linking declined message.

### Custom account semantics

By default, account linking sets a UCAN that delegates `SUPER_USER` access to the requestor. It expects that the provider is the root user (whose public key matches the DID registered with the server) or has been delegated `SUPER_USER` access by another device with sufficient capabilities.

Developers may want to use custom account semantics and they can by dependency injecting three functions:

https://github.com/fission-suite/webnative/blob/98ceb8f5264101481ba44e0bc85f27ed29d7823c/src/auth/implementation/types.ts#L12-L14

`checkCapability` checks that a provider can delegate an account for a username when `createProvider` is called. A provider calls `delegateAccount` to delegate credentials. A consumer calls `linkDevice` on receiving a delegation to process and store credentials. These last two functions are tightly coupled -- `linkDevice` receives whatever `delegateAccount` returns.

The [default implementation](https://github.com/fission-suite/webnative/blob/98ceb8f5264101481ba44e0bc85f27ed29d7823c/src/auth/local.ts#L41-L84) and [auth lobby implementation](https://github.com/fission-suite/webnative/blob/98ceb8f5264101481ba44e0bc85f27ed29d7823c/src/auth/lobby.ts#L111-L157) contain examples of these functions.

Account linking always sets the username in storage at `webnative.auth_username`. We rely on a username at this location in other parts of webnative, so we set this internally.

### Custom channels

By default, account linking uses websockets to send account linking messages between the provider and requestor. Developers may want to use another transport and they can by dependency injecting `createChannel`:

https://github.com/fission-suite/webnative/blob/98ceb8f5264101481ba44e0bc85f27ed29d7823c/src/auth/implementation/types.ts#L11

A channel is responsible for sending messages and calling `handleMessage` when messages are received. A channel must be named for the DID associated with the username.

https://github.com/fission-suite/webnative/blob/f75f4d22cca84c4437cdf60957971aa37a0f93b6/src/auth/channel.ts#L7-L17

## Test plan (required)

Unit tests and prop checks for the provider and requestor are in `producer.test.ts` and `consumer.test.ts`. Individual linking functions are tested, but the interaction between provider and requestor is not.

Integration tests to check the interaction between provider and requestor are in `linking.node.test.ts`. This test suite uses an event emitter for its account linking channel. It tracks when providers delegate and requestors link in separate objects. If account linking was successful, the objects should be the same at the end of the test.

## Closing issues

Closes #305 

## After Merge
* [ ] Does this change invalidate any docs or tutorials? No, but we will need new docs.
* [ ] Does this change require a release to be made? Yes, we will do an alpha release first

## Task list

There were several key pieces:

- [x] Add mechanisms for Opening / Closing and publishing data across channels
- [x] Make that system dependency injectable
- [x] Add a "state" machine for tracking progress in both consumer & producer
- [x] Implement each step as separate functions
- [x] Finalize API for interaction
- [x] Ensure that each step follows the current whitepaper
- [x] Handle errors / cancellations effectively


src/auth/linking.ts
```diff
@@ -0,0 +1,242 @@
...
+import utils from "keystore-idb/lib/utils.js"
+import { KeyUse, SymmAlg, HashAlg, CharSize } from "keystore-idb/lib/types.js"
+import * as did from "../did/index.js"
+import * as ucan from "../ucan/index.js"
+import { publishOnChannel } from "./index.js"
+import * as auth from "./index.js"
+
+type LinkingRole = "CONSUMER" | "PRODUCER"
+
+type LinkingStep = "BROADCAST" | "NEGOTIATION" | "DELEGATION"
```
walkah, line 13: These are the step names used in the auth-lobby... there might be better ones?

src/auth/linking.ts
```diff
@@ -0,0 +1,242 @@
...
+  role: null,
+  step: null
+}
+
+const resetLinkingState = () => {
+  ls.sessionKey = null
+  ls.temporaryRsaPair = null
+}
+
+const nextStep = () => {
```
walkah, line 36: we're keeping a very simple state machine here - one for each "role" in the exchange (producer and consumer) and tracking both in our `LinkingState`

src/auth/linking.ts
```diff
@@ -0,0 +1,242 @@
...
+  await sendTemporaryExchangeKey()
+  return null
+}
+
+export const startLinkingProducer = async (username: string): Promise<null> => {
+  ls.role = "PRODUCER"
+  ls.step = "BROADCAST"
+  await auth.openChannel(username)
+  return null
+}
```
walkah, line 62: I'm 100% open to suggestions here, I'm not sure I love this. I have these as separate, explicit functions because the get called separately (Producer on whatever linking UI page when authenticated, and Consumer when the user has provided user info and explicitly started the process). I suppose these two should be injectable- although both get called directly from implementer code and it might be easier to just re-implement with any modification (I.e. if you're not using usernames, etc ... and you probably have already swapped in different channel stuff).

TODO: Need to figure out how to handle the challenge PINs (since that all happens over the channel) .. I'm leaning towards having these functions have a callback for `handleChallengePin` (where both sides display the PIN, and producer would have an accept/reject mechanism).

src/auth/linking.ts
```diff
@@ -0,0 +1,242 @@
...
+  const message = new TextDecoder().decode(data.arrayBuffer ? await data.arrayBuffer() : data)
+  console.debug("Linking Status", ls)
+  console.debug("message (raw)", message)
+  switch (ls.role) {
+    case "CONSUMER":
+      await handleConsumerMessage(message)
+      break
+    case "PRODUCER":
+      await handleProducerMessage(message)
+      break
```
walkah, line 75: I don't know if this makes this more readable or not...

src/auth/linking.ts
```diff
@@ -0,0 +1,299 @@
...
+ * 
+ * This step is user initiated, create the UCAN with delegate rights for the CONSUMER (and a WNFS read key).
+ * 
+ * @param data 
+ * @returns 
+ */
+export const sendReadKeyAndUcan = async (data: any): Promise<null> => {
+  return null
+}
+
```
walkah, line 285: This is what should get called when the user accepts the PIN on the producer side. The auth-lobby code lives here: https://github.com/fission-suite/auth-lobby/blob/develop/src/Javascript/Main.js#L783

Because this is "user facing" (i.e. the implementor needs to trigger this in response to user action)... I'm not sure the best way to surface this. Maybe a friendlier / clearer name is all it needs?

VERY open to thoughts here.

src/auth/linking.ts
```diff
@@ -0,0 +1,299 @@
...
+ * Receive the encrypted UCAN & readKey and store them.
+ * 
+ * @param readKey 
+ * @param ucan 
+ * @param username 
+ * @returns 
+ */
+export const handleLinkedDevice = async (readKey: string, ucan: string, username: string): Promise<null> => {
+  //
+  return null
```
walkah, line 298: same as above.... not sure how best to surface this to the implementor. this data will actually come across the channel, so *not* via user action, but an implementor will probably want to have control over what happens to the readKey & UCan (and could potentially not even care about a readKey?)

I am leaning towards just having this be a DI impelemtation.

src/auth/linking.ts
expede: If it were me, I'd break these into separate modules. e.g. `Producer.handleMessage`

expede: i.e. I'm also not sold on this being the right abstraction

walkah: I'm certainly not opposed to that... would need to sort through the impact on the channel abstraction, but it's probably safe to assume that opening a channel will always require some level of callback for received messages (although the signature will change, which I was trying to keep consolidated in WSS implementation).

Anyway, this is 100% the layer where I was still noodling on what makes the most sense for potential future usage.

src/auth/linking.ts
```diff
@@ -0,0 +1,299 @@
...
+
+  const iv = utils.randomBuf(16)
+  const msg = await aes.encrypt(ucan.encode(u), ls.sessionKey, { iv, alg: SymmAlg.AES_GCM })
+
+  await publishOnChannel(
+    JSON.stringify({
+      iv: utils.arrBufToBase64(iv),
+      msg,
+      sessionKey: utils.arrBufToBase64(encryptedSessionKey)
+    })
```
walkah, line 156: Reminder for @bgins - this is where the auth-lobby and whitepaper have diverged (sending the UCAN encrypted by the AES key and the AES key separately encrypted by the RSA key).

bgins: I discussed this with @expede this morning, and this implementation is correct. While it would be nice to encrypt with the temporary RSA key, we can't fit enough data in with RSA. 

Instead, we encrypt the AES key with the temporary RSA key and the UCAN with the AES key. On the other end, the AES key is decrypted and then used to decrypt the UCAN.

matheus23:
Writing down some thoughts that Brooke and I had late last year, not because I think that should be implemented *in this PR*, but because I think it could inform the design for the protocol state:

When there's more than 2 participants in a websocket channel for device linking, we should track the protocol state for all of them (maybe until a maximum of 10 or 100 or 1000) and simultaneously execute the AWAKE protocol with all of them.
When the step for "comparing the numbers on both devices" comes, instead of showing the user numbers, we show the user a number input and they have to enter the number that they see on the other device themselves.
This both serves as a check "are you connecting with the device you think you're connecting?" as well as a way to *identify* the device you want to connect with, if there's multiple devices that successfully went through every previous step of the AWAKE protocol.

(I thought I had written up an issue for that, but I can't find it :/ )

Again, just to inform the design code-wise. I personally find it helpful seeing ways that code might evolve in the future.

bgins:
> When there's more than 2 participants in a websocket channel for device linking, we should track the protocol state for all of them (maybe until a maximum of 10 or 100 or 1000) and simultaneously execute the AWAKE protocol with all of them. When the step for "comparing the numbers on both devices" comes, instead of showing the user numbers, we show the user a number input and they have to enter the number that they see on the other device themselves. This both serves as a check "are you connecting with the device you think you're connecting?" as well as a way to _identify_ the device you want to connect with, if there's multiple devices that successfully went through every previous step of the AWAKE protocol.

Thanks for bringing it up! Good to think about this now while we are working on this for the app owned version.

Do we think we should replace this original pin interaction with this new flow? Or allow for both?

I see that @expede opened an issue to discuss AWAKE improvements. I have some questions about how the new flow would work that I'll ask over there.

matheus23:
> Do we think we should replace this original pin interaction with this new flow? Or allow for both?

I (personally) think we should replace it at some point. But not this PR anyway, so that's a discussion for another day :smile:

tests/auth/linking.node.test.ts
```diff
@@ -0,0 +1,186 @@
...
+    producer.cancel()
+
+    expect(consumerAccounts["elm-owl"]).not.toBeDefined()
+    expect(producerAccounts["elm-owl"]).not.toBeDefined()
+    expect(consumerAccounts).toEqual(producerAccounts)
+  })
+
+
+  // TODO: Run this test when we have implemented a message queue
+  it.skip("links with one producer and multiple consumers", async () => {
```
bgins, line 156: We are skipping this test for now because one producer and multiple consumers does not work reliably. We should bring this test back when we implement a queue to track producer and consumer messages.

src/auth/index.ts
```diff
@@ -0,0 +1,7 @@
+import { impl } from "./implementation.js"
+
+const {  isUsernameAvailable, isUsernameValid, register } = impl
+
+export { isUsernameAvailable, isUsernameValid, register }
+export { AccountLinkingRequestor, createConsumer as createRequestor } from "./linking/consumer.js"
+export { AccountLinkingProvider, createProducer as createProvider } from "./linking/producer.js"
```
bgins, line 7: All app owned auth functions are collected up and re-exported here. We also rename consumer and producer to requestor and provider.

src/index.ts
```diff
@@ -99,6 +99,7 @@ export { Scenario, State } from "./auth/state.js"
 export { AuthCancelled, AuthSucceeded, Continuation, NotAuthorised } from "./auth/state.js"
 export { InitialisationError, InitOptions } from "./init/types.js"
 
+export * as auth from "./auth/index.js"
```
bgins, line 102: App owned auth functions are exported from the top level as `auth`.

tests/auth/linking.node.test.ts
```diff
@@ -0,0 +1,186 @@
...
+
+    setImplementations({
+      auth: {
+        ...LOCAL_IMPLEMENTATION.auth,
+        createChannel,
+        checkCapability,
+        delegateAccount,
+        linkDevice
+      }
+    })
```
bgins, line 70: This is roughly what dependency injection will look like from a developer perspective. (With different function implementations, these ones are only good for the tests.)

src/auth/linking/producer.ts
```diff
@@ -0,0 +1,294 @@
...
+        if (userChallengeResult.ok) {
+          const { pin, audience } = userChallengeResult.value
+
+          const challengeOnce = () => {
+            let called = false
+
+            return {
+              confirmPin: async () => {
+                if (!called) {
+                  called = true
```
matheus23, line 88: It's good that we have this check in here! A user might accidentally e.g. call *both* `confirmPin()` and `rejectPin()`.

Because this is a user error, can we throw an error in the case `called` is true?

src/auth/channel.ts
```diff
@@ -0,0 +1,73 @@
...
+  username: string
+  handleMessage: (event: MessageEvent) => void
+}
+
+type ChannelData = string | ArrayBufferLike | Blob | ArrayBufferView
+
+export const createChannel = async (options: ChannelOptions): Promise<Channel> => {
+  const { username, handleMessage } = options
+
+  const rootDid = await await did.root(username).catch(_ => null)
```
matheus23, line 22: I know it's annoying that javascript (or rather eslint) doesn't recognize `_` as unused variable, but you can silence this warning by using `() => null` instead.

src/auth/channel.ts
```diff
@@ -0,0 +1,73 @@
...
+}
+
+export type ChannelOptions = {
+  username: string
+  handleMessage: (event: MessageEvent) => void
+}
+
+type ChannelData = string | ArrayBufferLike | Blob | ArrayBufferView
+
+export const createChannel = async (options: ChannelOptions): Promise<Channel> => {
```
matheus23, line 19: What do you think of calling this `createWssChannel`? I think it might be more accurate, and I liked the naming of `closeWssChannel` below, although I know it doesn't match the dependency-injected function's naming.

src/auth/index.ts
matheus23: Actually, why do we rename them? Should we perhaps use `requestor` and `provider` everywhere?

src/auth/linking/consumer.test.ts
```diff
@@ -0,0 +1,442 @@
...
+    })
+    if (!temporaryRsaPair.privateKey) throw new Error("Temporary RSA private key missing")
+
+    const sessionKeyResult = await consumer.handleSessionKey(temporaryRsaPair.privateKey, message)
+
+    let val
+    if (sessionKeyResult.ok) { val = sessionKeyResult.value }
+
+    expect(sessionKeyResult.ok).toBe(true)
+    expect(val).toEqual(sessionKey)
```
matheus23, line 79: I usually go with this pattern:

```suggestion
    expect(sessionKeyResult.ok).toBe(true) // will already throw in case ok == false
    if (!sessionKeyResult.ok) throw "Invalid result" // but this is to make typescript happy
    
    expect(sessionKeyResult.value).toEqual(sessionKey)
```

But that's just to share my approach to it, I don't know which one is better :D

src/common/event-emitter.ts
```diff
@@ -0,0 +1,23 @@
+export type EventListener = (...args: any[]) => void
```
matheus23, line 1: Can you use `unknown[]` here?

tests/auth/linking.node.test.ts
matheus23: Sounds fine. We can create an issue for that once this is merged I think :+1:

matheus23 (approved):
Super happy with this & the tests! :tada: Great work.

matheus23:
CI note: we should either increase the timeout or decrease the `numRuns` for one of the symlink tests that keeps timing out occasionally.

src/auth/index.ts
bgins: We could! The renaming is in part an artifact starting with `consumer` and `producer` and arriving at the other names later. 

But there are a couple of other considerations:
- The whitepaper uses the terms `consumer` and `producer`. Keeping these terms makes it easier to compare the implementation with the spec.
- We may re-purpose parts of the `consumer` and `producer` modules for other AWAKE purposes (account recovery, chat, etc) where the names `requestor` and `provider` may not make sense. `consumer` and `producer` are generic, architectural terms that apply broadly.

src/auth/linking/producer.ts
```diff
@@ -0,0 +1,294 @@
...
+    sessionKey: null,
+    temporaryRsaPair: null,
+    step: "BROADCAST"
+  }
+
+  const handleMessage = async (event: MessageEvent): Promise<void> => {
+    const { data } = event
+    const message = data.arrayBuffer ? new TextDecoder().decode(await data.arrayBuffer()) : data
+
+    if (ls.step === "BROADCAST") {
```
icidasset, line 69: Minor note, might be useful to make these steps (eg. "BROADCAST") constants, so the compiler will detect typos? Also, is there a reason the `Steps` type is redefined in this module, rather than defining it one place. Is it so you expose as little as possible of the internals?

src/auth/lobby.ts
```diff
@@ -100,6 +104,58 @@ export const isUsernameAvailable = async (username: string): Promise<boolean> =>
...
+      proof,
+  
+      // TODO: UCAN v0.7.0
+      // proofs: [ await localforage.getItem("ucan") ]
+    })
+  
+    return { readKey, token: token.encode(u) }
+}
+
+export const linkDevice = async (data: Record<string, unknown>): Promise<void> => {
```
icidasset, line 149: Could we require the `token` and `readKey` params specifically here, so it's clear from looking at the docs what the function expects?

src/auth/linking/producer.ts
matheus23: To be clear: I absolutely agree on having the `if (!called)` in there.
I was just thinking, maybe having an `else throw new Error(...)` in there might be nice as well :)

matheus23: Maybe I'm misunderstanding you - is there a case where throwing would be bad?

bgins: Yes, there is. I may not be explaining it well.

If a provider links for more than one requestor, all `confirmPin` or `rejectPin` functions it has emitted will be called. This would lead to exceptions in the functions that had been marked as called.

I'd be happy to discuss this synchronously and demonstrate where I think this would go wrong. There may be a simple solution that I am overlooking here.

src/auth/channel.ts
bgins: Good point. I've updated this to `()`.

src/auth/channel.ts
bgins: Agreed. Renaming this is also great because it makes it more explicit that the default and lobby implementations are using websockets.

src/auth/linking/consumer.test.ts
bgins: Yeah, that is a nice approach because you don't need the `val` variable, with the tradeoff of the extra line for TypeScript. 🤔 

I'm not sure which approach is better either. Will keep these as is for now since they about equal to me.

src/auth/index.ts
matheus23: This sounds like `consumer` and `producer` are better names. I'd vouch for using them consistently everywhere then :+1:

bgins: The goal of renaming them is to make them more intuitive in the API. In my opinion, consumer and producer don't carry enough meaning in the context of account linking unless you are familiar with the implementation or have read the whitepaper. Provider and requestor carry more implication of their purpose.

There is absolutely a tradeoff here. Folks who are interested in reading the whitepaper to understand the implementation will need to take a first step translating the terms when reading the code. We also have a bit of cognitive overhead maintaining the code. On the other hand, folks who just want to use the API have clearer, more meaningful terms.

bmann: I get this take but in general there isn’t a good way around this for ALL auth / identity related terms. eG IDP =
Identity Provider. 

If people are using this as a library where they aren’t needing to write code, maybe alignment with the spec version is useful. 

I’d like to see an external user say “this is confusing” first. I don’t at first read see requestor / provider being that much clearer?

the two “p” words of producer / provider seem too close especially?

bgins: As an eyeball comparison, let's look at both options:

```typescript
const provider = await auth.createProvider({ username })
const requestor = await auth.createRequestor({ username })
```
A provider provides auth, and a requestor requests auth.

```typescript
const producer = await auth.createProducer({ username })
const consumer = await auth.createConsumer({ username })
```
A produces produces auth, and a consumer consumes auth.

Providing and requesting auth make sense intuitively, where, at first read, I question what it means to produce and consume auth. Again, this is just my intuition on it! This is highly subjective and my view isn't necessarily right.

> If people are using this as a library where they aren’t needing to write code, maybe alignment with the spec version is useful.

That's a fair point! If someone only looked at the API docs and the whitepaper, this might not be clear.

> I get this take but in general there isn’t a good way around this for ALL auth / identity related terms. eG IDP =
Identity Provider.

Even if there aren't good standard terms, it's worth thinking about what terms we can draw intuitions from with the goal of making the API self-evident to developers.

bgins: Another thing to consider is if we do re-use portions of the consumer and producer for other use cases, we may want names that are more specific. For example, some parts of the consumer and producer will be usable in account recovery (and maybe more!).

One other way to approach this would be to differentiate consumers and producers by top level export.

bmann: To me neither of these are clearer.

> Providing and requesting auth make sense intuitively

I don't think it does, especially as you look across language barriers and such. Needing a provider/requester or producer/consumer just doesn't feel different enough to me to warrant renaming.

And that's the challenge. We both have "correct" hypotheses, and won't know until we get feedback in the wild. Mostly, I think people will cut and paste code examples to start with.

Not a hugely important issue to me either way.

src/common/event-emitter.ts
bgins: If we use `unknown[]` here, we end up with type errors here:

https://github.com/fission-suite/webnative/blob/6464460024c91e47b3cfba12681f77e7c24b2acc/src/auth/linking/consumer.ts#L119

For example, `Type 'unknown' is not assignable to type '{ pin: number[]; }'` which comes from the `OnChallenge` type.

I tried experimenting with generics like this:

```typescript
export type EventListener = <T>(...args: T[]) => void
```
```typescript
export type AccountLinkingRequestor = {
  on: OnChallenge<ChallengeArgs> & OnLink<LinkArgs> & OnDone
  cancel: () => void
}

type ChallengeArgs = {
  pin: number[]
}

type OnChallenge<ChallengeArgs> = (event: "challenge", listener: (args: ChallengeArgs) => void) => void
```
but that results in a similar type error: `Type 'T' is not assignable to type 'ChallengeArgs'`

Any other ideas on how to type this?

One thing to consider is the hints displayed in VSCode. For example, it would be better for them to see `{ pin: number[]; }` instead of `ChallengeArgs`.

src/auth/index.ts
bgins: > And that's the challenge. We both have "correct" hypotheses, and won't know until we get feedback in the wild.

Totally agreed. It's hard to gauge without hearing from users. 💯 

> Mostly, I think people will cut and paste code examples to start with.

True! Most folks won't care all that much and will directly use what we have in docs or a demo app.

bgins:
> CI note: we should either increase the timeout or decrease the `numRuns` for one of the symlink tests that keeps timing out occasionally.

Sounds good. I decreased the `numRuns` from 50 down to 25 and that gets them running through in time.

src/auth/index.ts
bgins: After discussing this with @therealjeffg, I think we will avoid renaming and use consumer and producer. The names provider and requestor aren't distinct enough to warrant renaming.

We did discuss contextualizing consumer and producer at the top level export by exposing them from `account` instead of `auth`. From a developer perspective, they would look like:

```typescript
import { account } from 'webnative'

const acctProducer = await account.createProducer({ username })
```

In the future, when we find other uses for AWAKE, we can export from other top level exports. For example, we might export account recovery from `recovery`.

src/auth/channel.ts
```diff
@@ -0,0 +1,73 @@
...
+  const { username, handleMessage } = options
+
+  const rootDid = await await did.root(username).catch(_ => null)
+  if (!rootDid) {
+    throw new LinkingError(`Failed to lookup DID for ${username}`)
+  }
+
+  const apiEndpoint = setup.getApiEndpoint()
+  const endpoint = apiEndpoint.replace(/^https?:\/\//, "wss://")
```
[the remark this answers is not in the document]
bgins: That's a good question! My hunch is that should work fine.

I'd say let's take that on as a separate PR. We may have a few places that are hardcoded like this, and we will probably want to add a setup option to select secure or insecure.

src/auth/linking/producer.ts
bgins: > Minor note, might be useful to make these steps (eg. "BROADCAST") constants, so the compiler will detect typos? 

I'm not sure if I understand. Do you mean making them enums? We have the `LinkingStep` type

https://github.com/fission-suite/webnative/blob/a2cb14a2b9c493329bdb207fa5b5cdb524597763/src/auth/linking/consumer.ts#L27

which gives us some checks on what gets assigned to `ls.step`.

> Also, is there a reason the `Steps` type is redefined in this module, rather than defining it one place. Is it so you expose as little as possible of the internals?

No good reason! They turned out the same in the producer and consumer. I'll move them out to `linking.ts` one level up.

src/auth/lobby.ts
bgins: I think we need to type to match the type in the implementation

https://github.com/fission-suite/webnative/blob/787db781ff4061736fc8c1f6225e68be7de06984/src/auth/implementation/types.ts#L14

and we want to keep that type general to suit different account semantics. But maybe we could use generics somehow to form an explicit contract between `delegateAccount` and `linkDevice`? 🤔 

I'm not sure how we would do this, but will ponder it. Any ideas or thoughts on other ways we could approach this?

src/auth/linking/producer.ts
icidasset: > which gives us some checks on what gets assigned to `ls.step`

Yeah true, not sure it matters that much regarding type checking.
Enums are indeed one way to approach this. I was thinking of this:
```ts
enum Steps { Broadcast = "BROADCAST",  Negotiation = "NEGOTIATION" }
type LinkingSteps = Broadcast | Negotiation
```

src/auth/lobby.ts
icidasset: Ah gotcha 👍 I would just add some documentation then. Not worried about the type checking or anything here.

src/index.ts
bgins: Renamed this to export as `account` instead of `auth` to contextualize the consumer and producer.

src/auth/linking/producer.ts
bgins: Good idea, I've converted this to use an `enum`. The `enum` version reads cleaner and safer than the string version. Thanks for the suggestion. 🙏

matheus23:
This is looking good!
Last thing I'd love to see is backwards compatibility with the auth lobby. My impression was that this PR shouldn't change the protocol, so it seems weird to me that the auth lobby wouldn't link against the demo app :thinking: My feeling is this *might* expose an issue. We should check what it is just to be sure.

matheus23 (approved):
We can go ahead and merge this. We're not 100% sure yet whether we'll need/want backwards-compat on the AWAKE protocol, so we wouldn't want to cut a release with this PR immediately.
However, we can decide that still after merging, so let's go ahead! :tada: 

yes

src/common/event-emitter.ts
bgins: We converted this interface to use an event maps to type the event emitter in https://github.com/fission-suite/webnative/pull/335/commits/2470765a1387e917bf1845ee228c49c9b4e19a17.

PR #339, opened 2022-01-11
Some soft links improvements

icidasset:
1. Improves soft link resolving when using "nested" paths
  ```js
  fs.read({ file: [ "private", "Symlink", "file" ] })
  ```
2. Doesn't do the dnslink lookup when resolving our own "username.files.fission.name" dnslink


tests/fs/api.private.node.test.ts
```diff
@@ -1,9 +1,14 @@
 import expect from "expect"
 import * as fc from "fast-check"
 
+// @ts-ignore
+import mockery from "mockery"
```
matheus23, line 5: I can't find a usage of this. Can it be deleted?

matheus23 (approved):
Ok, so this mainly moves the "resolving soft links" logic into `getRecurse` instead of `get`, which allows it to resolve also when the soft link isn't the last element of the path, right?

Makes sense! :) :shipit:

tests/fs/api.private.node.test.ts
icidasset: Oh whoops, yeah was part of the "proxying" experiments.

icidasset:
@matheus23 Yes exactly 👍

PR #343, opened 2022-01-21
Fix CID and DAG-PB inconsistencies

icidasset:
- Removes usage of older DAG-PB library
- Removes usage of older CID library (cids)
- Removes usage of other CBOR library (cborg)
- Uses CID class instances as much as possible (explained in changelog as well) instead of just strings
- Updates all dependencies to the latest version
- Introduces `decodeCID()` and `encodeCID()` functions, see `common/cid.ts`

CIDs coming from IPFS can either be a string, a CID class instance or a CID object, hence the introduction of the `decodeCID()` function.

I've published this as version `0.30.0-alpha14`, in case you want to test it. I already did a publish because `0.30.0-alpha12` was broken 🙈


icidasset:
Oh also, the webnative generated docs have a dark mode now, thanks to the typedoc upgrade.

src/common/cid.ts
```diff
@@ -0,0 +1,26 @@
+import { CID } from "multiformats/cid"
+
+
+/**
+ * Decode a possibly string-encoded CID.
+ * Passing an already decoded CID instance works too.
+ * NOTE: Throws an error if a CID cannot be decoded!
```
matheus23, line 7: Nitpick: There's `@throws` you could also use. Not that it really matters that much for the tooling we use - just seems to be the convention.

src/common/cid.ts
```diff
@@ -0,0 +1,26 @@
...
+ */
+export function decodeCID(val: CID | object | string): CID {
+  const cid = CID.asCID(val)
+  if (cid) return cid
+
+  if (typeof val === "string") return CID.parse(val as string)
+  if (typeof val === "object" && "version" in val && "code" in val && "multihash" in val) {
+    return CID.create(val.version, val.code, val.multihash)
+  }
```
matheus23, line 16: What do you think about adding these helper functions I've been using a lot in various projects so far?
```ts
export function hasProp<K extends PropertyKey>(data: unknown, prop: K): data is Record<K, unknown> {
  return typeof data === "object" && data != null && prop in data
}
export function isRecord(data: unknown): data is Record<PropertyKey, unknown> {
  return typeof data === "object" && data != null
}
```
Then you'd use `isRecord(val) && hasProp(val, "version") && hasProp(val, "code") && hasProp(val, "multihash")`.
I think typescript doesn't have support for `"string" in x` to actually do anything to the types as far as I know (correct me if I'm wrong).

Also there's a couple of gotchas with `typeof val === "object"`, because e.g. `typeof null === "object"`, so I like getting used to using `isRecord`.

src/fs/protocol/public/index.ts
```diff
@@ -86,12 +88,14 @@ export const getValue = async (
...
-    const links = await basic.getSimpleLinks(linksOrCID)
+  const cid = CID.asCID(linksOrCID)
+
+  if (cid) {
+    const links = await basic.getSimpleLinks(cid)
     return getValueFromLinks(links, name)
   }
 
-  return getValueFromLinks(linksOrCID, name)
+  return getValueFromLinks(linksOrCID as SimpleLinks, name)
```
matheus23, line 98: Goddamnit. This whole `CID.asCID` pattern that PL has come up with breaks typescript type inference.
We *could* write a `(obj: unkown): obj is CID => notNull(CID.asCID(obj))` function, but I think that defeats the purpose, because you should still run `CID.asCID` on whatever you got & then typescript will complain that that returns something that can be null.
Hm. Let me know if I'm missing something we could do to help this situation.

src/fs/protocol/public/types.ts
```diff
@@ -11,9 +11,9 @@ export type PutDetails = AddResult & {
 }
 
 export type SkeletonInfo = {
-  cid: CID
-  userland: CID
-  metadata: CID
+  cid: CID | string
+  userland: CID | string
+  metadata: CID | string
```
matheus23, line 16: Why did you make these all `CID | string` instead of just `CID`?

src/ipfs/types.ts
```diff
@@ -1,4 +1,6 @@
-export type CID = string
+import type { CID as MultiformatsCID } from "multiformats/cid"
+
+export type CID = MultiformatsCID
```
matheus23, line 3: Shouldn't we just remove all of this, and use `import { CID } from "multiformats/cid"` everywhere?

tests/helpers/in-memory-ipfs.ts
```diff
@@ -18,6 +18,7 @@ export async function createInMemoryIPFS(): Promise<IPFS> {
   const memoryDs = new MemoryDatastore()
   const memoryBs = new MemoryBlockstore()
 
+  // @ts-ignore
```
matheus23, line 21: Why is that needed? Do we have to fix something in the config?

matheus23 (commented):
Glad that this has been done :smiley: 
A couple of comments, but mostly exactly what I'd expect :)

tests/helpers/in-memory-ipfs.ts
icidasset: I'm getting a type error for a weird multiformats inconsistency between the dag-pb and ipfs-core libraries 🤔 
```
✖ ERROR: tests/helpers/in-memory-ipfs.ts:21:3 - error TS2322: Type 'IPFS' is not assignable to type 'IPFS<{}>'.
  The types of 'object.put' are incompatible between these types.
    Type '(obj: import("/Users/icid/Work/webnative/node_modules/ipfs-core/node_modules/@ipld/dag-pb/types/src/interface").PBNode, options?: import("/Users/icid/Work/webnative/node_modules/ipfs-core-types/types/src/object/index").PutOptions | undefined) => Promise<...>' is not assignable to type '(obj: import("/Users/icid/Work/webnative/node_modules/@ipld/dag-pb/types/src/interface").PBNode, options?: import("/Users/icid/Work/webnative/node_modules/ipfs-core-types/types/src/object/index").PutOptions | undefined) => Promise<...>'.
      Types of parameters 'obj' and 'obj' are incompatible.
        Type 'import("/Users/icid/Work/webnative/node_modules/@ipld/dag-pb/types/src/interface").PBNode' is not assignable to type 'import("/Users/icid/Work/webnative/node_modules/ipfs-core/node_modules/@ipld/dag-pb/types/src/interface").PBNode'.
          Types of property 'Links' are incompatible.
            Type 'import("/Users/icid/Work/webnative/node_modules/@ipld/dag-pb/types/src/interface").PBLink[]' is not assignable to type 'import("/Users/icid/Work/webnative/node_modules/ipfs-core/node_modules/@ipld/dag-pb/types/src/interface").PBLink[]'.
              Type 'import("/Users/icid/Work/webnative/node_modules/@ipld/dag-pb/types/src/interface").PBLink' is not assignable to type 'import("/Users/icid/Work/webnative/node_modules/ipfs-core/node_modules/@ipld/dag-pb/types/src/interface").PBLink'.
                Types of property 'Hash' are incompatible.
                  Type 'import("/Users/icid/Work/webnative/node_modules/multiformats/types/src/cid").CID' is not assignable to type 'import("/Users/icid/Work/webnative/node_modules/peer-id/node_modules/multiformats/types/cid").CID'.
                    Types have separate declarations of a private property 'asCID'.
```

src/ipfs/types.ts
icidasset: Don't know 🤷‍♂️ What if the `CID` type changes again?! 😂

src/fs/protocol/public/types.ts
icidasset: Because these'll be coming from DAG structures directly which can have a CID of either types.

src/common/cid.ts
icidasset: We already have the `hasProp` and `isObject` functions yeah. That said, Typescript does support the `in` operator: https://www.typescriptlang.org/docs/handbook/2/narrowing.html#the-in-operator-narrowing

src/common/cid.ts
icidasset: Ah cool, will change.

src/fs/protocol/public/index.ts
icidasset: Yeah I'm not sure what Typescript is doing here.

This does not work:
Screenshot 2022-01-24 at 23 02 47
_If I remove the `&& cid` here, it gives me a "could null" error in the first `if` branch._

This does:
Screenshot 2022-01-24 at 23 04 44
_Should I do this instead?_

matheus23: I've given this another thought. Let's make the `CID.asCID` calls mostly happen on the surface-level API boundary. This function is pretty internal. So let's change it to something like this:
```ts

  if (check.isCID(linksOrCID)) {
    const links = await basic.getSimpleLinks(linksOrCID)
    return getValueFromLinks(links, name)
  }
```

src/ipfs/types.ts
matheus23: I think we're using `import { CID } from "multiformats/cid"` in some places already and we should go for consistency? If it happens to change again, find-replacing that import should do the same thing. It's just a *tiny bit* less indirections, but not a big deal to be honest.

tests/helpers/in-memory-ipfs.ts
matheus23: I added a commit that fixes this.
Behind the scenes: I took peek into the `yarn.lock` file. Among all installed libraries we see multiformats twice:
```
multiformats@^9.0.0, multiformats@^9.0.2, multiformats@^9.0.4, multiformats@^9.1.0, multiformats@^9.1.2, multiformats@^9.4.1, multiformats@^9.4.2, multiformats@^9.4.5, multiformats@^9.4.7:
  version "9.4.8"
  resolved "https://registry.yarnpkg.com/multiformats/-/multiformats-9.4.8.tgz#46f74ec116f7871f2a334cc6d4ad3d810dbac341"
  integrity sha512-EOJL02/kv+FD5hoItMhKgkYUUruJYMYFq4NQ6YkCh3jVQ5CuHo+OKdHeR50hAxEQmXQ9yvrM9BxLIk42xtfwnQ==

multiformats@^9.4.13, multiformats@^9.5.4:
  version "9.5.9"
  resolved "https://registry.yarnpkg.com/multiformats/-/multiformats-9.5.9.tgz#99ac8f4b997ec3dd2a2f563bdf8fdebdca437b99"
  integrity sha512-G3lKuMEV3H/414CnhiQr/BYPPacno3modE7ory2VFngxpMNAhg3GAYUeKGhdbllHPxX2Fw7mRjYB9mHO7yUwzQ==
```
It shows it's resolved to version `9.4.8` once and then `9.5.9` another time.
But all the version constraints are above: Everything that caused version `9.4.8` to install, would've worked fine with version `9.5.9` as well in terms of version constraints.
The reason I suspect this happens is because previously we used to not have the `^9.5.4` constraints before (at that time, that version probably didn't exist yet), because they're coming from the new `@ipld/*` libraries. The lockfile tries to prevent installing newer libraries, so instead it just forces two different versions.

The way I fixed this was I just set `resolutions: { multiformats: "9.6.2" }` in the `package.json` file, ran `yarn` and the removed the resolutions thing from the `package.json` again.
That forces yarn to use a single version for all dependencies. The dependencies are fine with that, because the new version is in their version bounds, I just have to force yarn to *not* lock the dependencies once and now they're locked at this compatible version again.

Running `yarn` (and not having `resolutions` in the `package.json`) totally works now & we don't need `ts-ignore` :)

src/fs/protocol/public/types.ts
matheus23: Hm. This forces us to handle `cid`, `userland` etc. being both string or CID at everywhere we're using these types.
I think it would be better if at the point where we're parsing something external we'd parse it into a canonical internal format, i.e. one with only `CID`s.
This *should* result in less usage of `decodeCID` overall, because it should only be used in the code that takes the "real-world" data. Everything internal wouldn't have to know about decodeCID anymore.

tests/helpers/in-memory-ipfs.ts
icidasset: Huh ok, thanks! I would've assumed `yarn` would override the lock when you update a dependency. Or maybe `pnpm` behaves differently in this regard, and I'm just confusing the two 🤔

matheus23 (approved):
THAT
:shipit:

src/fs/protocol/public/types.ts
icidasset: Hmm I now remember why I did this exactly. If I don't add the `CID | string` to these types here, I have to loop through all the skeletons and switch out the CIDs. I attempted to do this, but that seemed to involve much more refactoring than just adding a few `decodeCID` calls (10 of them to be exact, I just counted). I'm just gonna leave it for now, since we're going to be refactoring this later anyway.

matheus23: I think the other good news is that the skeleton will be gone with wnfs2 anyway :sweat_smile:

PR #347, opened 2022-01-28
Fixes #344

### LINKED ISSUE #344: Circular dependencies
Dependency injection for auth causes multiple circular dependencies.
Such as: setup/dependencies → auth/lobby → common/identifiers → crypto → setup/dependencies

icidasset:
Fixes the circular dependencies issue.

Also:
- Adds a test script to check for circular dependencies and runs this during `test:prod` too (CI)
- Improves the docs, thanks to https://github.com/Gerrit0/typedoc-plugin-missing-exports

Available as alpha version `0.31.0-alpha`


matheus23 (approved):
:tada: 

Should we release this as `v0.31.0` instead? Technically, it's a breaking change of the surface level API, because `setDependencies` is not `setImplementation`.

icidasset:
Ah good call 👍 Will change.

icidasset:
@bgins If you could take a look too that'd be great, since you wrote the initial dependency implementation I think? Also, we don't have any documentation on `setDependencies` (now `setImplementations`), right?

bgins (commented):
First, all credit is due to Daniel for implementing dependency injection. 😀 

One request. It looks like this doesn't have merge functionality to let developers pass in partial implementations. Could we add something like the `merge` used that was used in `setDependencies`?

Merging will be useful in App owned account linking, where a developer dependency injects in their own implementations of `delegateAccount` and `linkDevice` functions. We want to let them pass in just those functions and get the defaults for the rest of the implementation.

We don't have any docs on this at the moment, so nothing to change there.

icidasset:
@bgins Good point, added!

bgins (approved):
Thanks for adding the partial implementations. Looks good! 🎉

PR #348, opened 2022-01-31
Move some dependencies into devDependencies

matheus23:


icidasset (approved):
Thanks, I keep forgetting the `--save-dev` flag doesn't work with `yarn` 😅

PR #360, opened 2022-03-18
Make AWAKE protocol backwards compatible

matheus23:
Changes in this PR:
* Replace keystore-idb `aes.encode` and `aes.decode` calls with direct WebCrypto calls, because keystore-idb adds the IV (initialization vector) in front of messages, whereas the current AWAKE protocol in production expects the IV as JSON next to the message.
* Upon linking, also transfer the `readKey` in the message (to get compatible with what the auth lobby & CLI expect) and to enable reading the WNFS to which write access was granted through linking. Also rename the `token` message to `ucan`.
* When accepting linking, we don't send a `linkStatus: ACCEPT` key. However, I kept the `linkStatus: DENIED` message. That's a backwards-compatible addition, and now that it's in the API for AOL, it was easier to keep than to remove (and totally should be part of the AWAKE protocol anyway!).

As discussed with @bgins, this is essentially aligning code changes from #335 to the auth lobby's code for the AWAKE protocol, so that this will be compatible with the current auth lobby & the fission CLI, so they can all link to each other.


matheus23:
Tests are failing with `Cipher job failed`. Never seen that and can only fine a single unhelpful google result for what that is.

It's working for me when running the aol-demo app. I'm not sure what's going on. My guess is it must have something to do with WebCrypto in nodejs.

bgins:
> Upon linking, also transfer the readKey in the message (to get compatible with what the auth lobby & CLI expect) and to enable reading the WNFS to which write access was granted through linking.

How does this work in applications that do not use WNFS? Say an app uses the `local` implementation and does not use WNFS. Would it fail here:

https://github.com/fission-suite/webnative/blob/2b5364a5cc8d9232b5f31fe9a10bca362a9cdfd2/src/auth/local.ts#L76

There may be some tension here between providing backwards compatibility and decoupling account linking from the rest of the Fission and webnative stack. The local implementation was intended as a baseline that does not assume WNFS, but there is a loss of compatibility with the lobby (and the CLI). The loss of compatibility would not be ideal if an app wanted to add WNFS later. 

What would you think about making the `readKey` optional in `linkDevice` local implementation to provide compatibility, but removing the `readKey` expectation in `delegateAccount`? This would be one way to keep a balance between decoupling account linking and backwards compatibility. (We would lose some coherence between `delegateAccount` and `linkDevice`, would need to document the intent well if we did this.)

bgins:
> The local implementation was intended as a baseline that does not assume WNFS, but there is a loss of compatibility with the lobby (and the CLI).

Another possibility would be to rename `local.ts` to `base.ts` and export a `BASE_IMPLEMENTATION` from it without read key support. That way it serves as a building block without depending on WNFS.

If we did this (and maybe in general), we might want it into `src/auth/implementation/` and move `lobby.ts` there as well.

matheus23:
@bgins and I talked about this yesterday again.

We think that the "`local`" implementation shouldn't change to be auth-lobby compatible.

The only thing we *need* to make sure is that **the auth lobby can write an AWAKE dependency injection that loads UCANs & the read key that's and is compatible with the fission CLI**.

We don't strictly need to add changes to the AWAKE dependency injections that exist in webnative today, though. I personally think we can just revert them completely.

---

Apart from that, we should still try to fix the `Cipher job failed` errors in our tests.

matheus23:
@bgins I think this is ready for review again now :)

I've reverted the changes made to `auth/local.ts`, now that I actually understand what that file is for :grimacing: 
The implementation at `auth/lobby.ts` looks like it has the correct fields now (using `ucan` instead of `token` in the delegation message).

The "cipher job failed" messages were just caused by the tests still using keystore-idb for encryption and decryption, and thus messages & keys weren't lining up nicely. I've changed the tests to use different de & encryption methods now.

Another big thing I had to change was making the code not expect `linkingStatus` in the delegation message, and putting the delegation message on the top level, so to say (not behind the `delegation: {` key).

Please let me know what you think :)

src/auth/linking/producer.test.ts
```diff
@@ -79,19 +81,21 @@ describe("handle user challenge", async () => {
...
         }), async ({ pin, iv }) => {
-          const msg = await aes.encrypt(JSON.stringify({ did: DID, pin }), sessionKey, { iv, alg: SymmAlg.AES_GCM })
+          const msg = await aesEncrypt(JSON.stringify({ did: DID, pin }), sessionKey, iv)
           const challenge = JSON.stringify({ iv: utils.arrBufToBase64(iv), msg })
           const userChallengeResult = await producer.handleUserChallenge(sessionKey, challenge)
 
-          let val = null
-          if (userChallengeResult.ok) { val = userChallengeResult.value }
+          if (!userChallengeResult.ok) {
+            console.log(userChallengeResult)
```
bgins, line 92: Should this console log be removed?

bgins (approved):
Works great! Tried it out for backward compatibility with the auth lobby and CLI, going each direction with each.

> Another big thing I had to change was making the code not expect linkingStatus in the delegation message, and putting the delegation message on the top level, so to say (not behind the delegation: { key).

This change was part of what we needed for backward compatibility, yeah? I'd love to have richer messages in the future where approval isn't implicit, but this seems like the right change for now.

Thanks for updating this to make it backward compatible. Much appreciated. 🙏

matheus23:
> This change was part of what we needed for backward compatibility, yeah? I'd love to have richer messages in the future where approval isn't implicit, but this seems like the right change for now.

The `DENIED` message is still there! So we're still paving the road for a future where we can detect approval/denial!

It's just that we don't encode approval via the `linkStatus: APPROVED` field. That's all :) (It's a matter of encoding).

PR #364, opened 2022-03-29
Rejection Sampling in Pin Gen, AWAKE

### LINKED ISSUE #359: Use rejection sampling for AWAKE protocol
Currently we generate the pin to be sent by generating 6 bytes and interpreting each modulo 10:
https://github.com/fission-suite/webnative/blob/d69fce99533b312ef94544fbec3f6db7542c9fdb/src/auth/linking/consumer.ts#L224-L226

This is most likely good enough, but has a slight bias towards digits 0-5. Instead, we should use rejection sampling to get a uniform distribution.

appcypher:
## Summary

This PR fixes the following **issues**

* [x] Sampling bias towards low numbers

Currently pin sampling method has a slight bias towards low numbers (0-5). This PR gets rid of this bias with updated `randomBuf` implementation that takes a max param and uses rejection sampling. 

## Test plan (required)

Making sure existing tests around pin generation still pass. 

```bash
yarn test:prod -g "generates a pin and challenge message"
```

## Closing issues

Fixes #359

## Depends On

- [x] https://github.com/fission-suite/keystore-idb/pull/65
  - [x] Need to update keystore-idb version when it gets merged and published.

### DIFF
--- a/package.json
+++ b/package.json
@@ -99,7 +99,7 @@
     "ipfs-message-port-client": "^0.10.3",
     "ipfs-message-port-protocol": "^0.10.5",
     "ipld-dag-pb": "^0.22.3",
-    "keystore-idb": "^0.15.4",
+    "keystore-idb": "^0.15.5",
     "localforage": "^1.10.0",
     "multiformats": "^9.5.4",
     "one-webcrypto": "^1.0.3",
--- a/src/auth/linking/consumer.ts
+++ b/src/auth/linking/consumer.ts
@@ -221,7 +221,7 @@ export const handleSessionKey = async (temporaryRsaPrivateKey: CryptoKey, data: 
  * @returns pin and challenge message
  */
 export const generateUserChallenge = async (sessionKey: CryptoKey): Promise<{ pin: Uint8Array; challenge: string }> => {
-  const pin = new Uint8Array(utils.randomBuf(6, 9))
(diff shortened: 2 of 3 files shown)


matheus23 (changes requested):
Thanks for tackling this! :)

The method to get rid of the bias isn't familiar to me and it's not clear to me how this gets rid of the bias. Is there some resource I can read to understand it better? If so, we should probably also include a comment in the code linking to it.

I should've probably added this to the issue when I created it, but I was thinking of something like this: https://research.kudelskisecurity.com/2020/07/28/the-definitive-guide-to-modulo-bias-and-how-to-avoid-it/

This is not me saying "that's how it should be done". If there's a method that doesn't need to re-sample to reduce bias, then it seems superior!

expede:
I probably should have fleshed out that comment a lot more 😅 We don't want bias in the number; ideally they're fully evenly distributed in order to make attacking this code difficult. It's an edge case, but security is important. I was more gesturing at multiple options and that we're going to start more consistently phrasing things as stories rather than as specific solutions when we start scrum next week :)

appcypher:
Alright. Let me fix that

matheus23 (approved):
Perfect, this is good to go! :tada: 

(Just make sure to release the new keystore-idb version before merging this & make sure that the tests run :+1: )

PR #365, opened 2022-03-30
Type-Check CID

### LINKED ISSUE #345: CID conversion error
```
  index.js:73 TypeError: Object.defineProperty called on non-object
    at Function.defineProperty (<anonymous>)
    at decodeCID (auth.fission.codes/worker.min.js:2989:14)
    at DAGService.get (auth.fission.codes/worker.min.js:3055:64)
```
Problem with https://github.com/icidasset/nourish
Have not seen it in other apps yet.

appcypher:
## Summary

* [x] Type-checks CID for better error message

`cid`s have been updated to use the multiformat-based CID class but old code still use string encoded formats which leads to vague errors. This PR adds a check that returns a clear error message when a cid parameter isn't an object.

## Closing issues

Fixes #345

### DIFF
--- a/src/ipfs/basic.ts
+++ b/src/ipfs/basic.ts
@@ -132,5 +132,5 @@ const attemptPin = async (cid: CID): Promise<void> => {
 }
 
 const typeCheckCID = (cid: CID): void => {
-  if (!CID.asCID(cid)) throw new Error("Expected a CID object")
+  if (!CID.asCID(cid)) throw new Error(`Expected a CID object = ${cid}`)
 }
--- a/src/ipfs/basic.ts
+++ b/src/ipfs/basic.ts
@@ -28,6 +28,8 @@ export const add = async (content: ImportCandidate): Promise<AddResult> => {
 }
 
 export const catRaw = async (cid: CID): Promise<Uint8Array[]> => {
+  typeCheckCID(cid)
+
   const ipfs = await getIpfs()
   const chunks = []
   await attemptPin(cid)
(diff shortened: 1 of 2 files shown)


src/fs/filesystem.ts
```diff
@@ -164,6 +164,8 @@ export class FileSystem {
    * Loads an existing file system from a CID.
    */
   static async fromCID(cid: CID, opts: FileSystemOptions = {}): Promise<FileSystem | null> {
+    if (!(cid instanceof CID)) throw new Error("Expected a CID object")
```
matheus23, line 167: I'm sorry for throwing you into JS/TS land which is horrible :P

`instanceof` is dangerous in JS, because if for whatever reason you'll end up with a duplicated copy of the multiformats library in your bundle, their `CID` classes are not `instanceof`-compatible with each other.

Here's an issue comment that shows the desired pattern to use instead: https://github.com/multiformats/js-cid/issues/111#issuecomment-645003427

I know, this is an annoyingly small thing and you want to get working on something bigger, it's just for you to get familiar with different parts of our toolchain + code. (Almost done :tada:)

appcypher: No. It's fine. You have been really helpful. 😀

src/ipfs/basic.ts
```diff
@@ -105,6 +105,8 @@ export const size = async (cid: CID): Promise<number> => {
 }
 
 export const attemptPin = async (cid: CID): Promise<void> => {
+  if (CID.isCID(cid)) throw new Error("Expected a CID object")
```
icidasset, line 108: You forgot the negation, but otherwise all good! 😄 ie. you should check if it's **not** a CID.

appcypher: Oh. I was testing the commit signing here. I'm not done. I had a rethink of the implementation and I will be pushing that now. We might need to hop on a call and figure how the check should be done based on the discussion we had yesterday.

appcypher:
For context and as discussed with @icidasset, the [error we get](https://github.com/fission-suite/webnative/issues/345) comes from CID being sent via postMessage to a decoding piece that assumes CIDs are of CID class instance (not strings). So the ideal place to do a check is right before the CID is sent since there are other areas in the code that handle a CID multiform nature.

matheus23:
Can we make this `cid = typeCheckCID(cid)`, where `typeCheckCID` returns what `CID.asCID` returns if its non-falsy?
This way we're automatically converting between a couple of CID types in some edge cases.

appcypher:
Good to go? @icidasset @matheus23

icidasset (approved):

matheus23 (approved):
:tada:

PR #367, opened 2022-03-31
Root DID Lookup Retry

### LINKED ISSUE #338: Account linking: Retry DID lookup
# Summary

## Problem

The websockets account linking channel sometimes fails to lookup a new user's root DID.

### Impact

Account linking may fail for a new user. We ask users to link a new device immediately on sign up, so this might cause some failure cases.

## Solution

Wait for a moment on failure and try to open the channel again.

# Detail

**Describe the bug**

When a user creates a new account, a channel is opened to allow them to link other devices. In some cases, the server fails to find the user's root DID and the channel cannot be opened.

**To Reproduce**
1. Create a new account
2. Watch console for the error `failed to lookup root DID`

**Expected behavior**

The account linking channel should be opened on the first attempt or on a retry.

**Screenshots**

This screenshot is from app owned account linking. Some of the console messages are temporary, but this shows the error:

failed-lookup

 - OS: macOS
 - Browser: Firefox container
 - Version: commit #ff0814cefd8f887a3bce4337b36257c40c10be53

**Additional context**

This has only been observed in the AOL account linking flow. We should verify it in the regular auth lobby flow and see if this issue is accounted for there.

The DID lookup happens here:

https://github.com/fission-suite/webnative/blob/d69fce99533b312ef94544fbec3f6db7542c9fdb/src/auth/channel.ts#L22

appcypher:
## Summary

This PR fixes the following **issues**

* [x] Root DID lookup error

Sometimes the root DID lookup fails. The root cause is not known yet but this PR implements a fix for it. We simply retry when the first attempt fails. 

## Test plan (required)

```bash
yarn test
```

## Closing issues

Fixes #338

### DIFF
--- a/src/auth/channel.ts
+++ b/src/auth/channel.ts
@@ -19,7 +19,7 @@ export type ChannelData = string | ArrayBufferLike | Blob | ArrayBufferView
 export const createWssChannel = async (options: ChannelOptions): Promise<Channel> => {
   const { username, handleMessage } = options
 
-  const rootDid = await await did.root(username).catch(() => null)
+  const rootDid = await waitForRootDid(username)
   if (!rootDid) {
     throw new LinkingError(`Failed to lookup DID for ${username}`)
   }
@@ -40,6 +40,25 @@ export const createWssChannel = async (options: ChannelOptions): Promise<Channel
     send,
     close
   }
+}
+
+const waitForRootDid = async (username: string): Promise<string | null> => {
+  return new Promise((resolve) => {
+    const maxTries = 3
+    let tries = 0
+    let rootDid = null
+
+    const rootDidInterval = setInterval(async () => {
+      rootDid = await await did.root(username).catch(() => null)
+      if (!rootDid && tries < maxTries) {
+        tries++
+        return
+      }
+
+      clearInterval(rootDidInterval)
+      resolve(rootDid)
+    }, 2000)
+  })
 }
 
 const waitForOpenConnection = async (socket: WebSocket): Promise<void> => {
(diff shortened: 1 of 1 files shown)


src/auth/channel.ts
```diff
@@ -42,6 +42,25 @@ export const createWssChannel = async (options: ChannelOptions): Promise<Channel
...
 }
 
+const waitForRootDid = async (username: string): Promise<string | null> => {
+  return new Promise((resolve) => {
+    const maxTries = 3
+    let tries = 0
+    let rootDid = null
+
+    const rootDidInterval = setInterval(async () => {
+      rootDid = await await did.root(username).catch(() => null)
```
bgins, line 52: Can we remove one of the `await`s here? (This was there before your PR. Don't think we need it.)

bgins, line 52: Can we remove one of the `await`s here? (This was there before your PR. Don't think we need it.)

bgins (approved):
Looks great! 💯 

I see that you are getting a failing changelog check. You can get rid of that by adding the "no changelog" tag and pushing another commit to your branch. (Add the tag first, before pushing the commit.)

src/auth/channel.ts
```diff
@@ -19,7 +19,7 @@ export type ChannelData = string | ArrayBufferLike | Blob | ArrayBufferView
 export const createWssChannel = async (options: ChannelOptions): Promise<Channel> => {
   const { username, handleMessage } = options
 
-  const rootDid = await await did.root(username).catch(() => null)
+  const rootDid = await waitForRootDid(username)
```
icidasset, line 22: Should this not run `did.root()` immediately, instead of waiting 2 seconds the first time around?

appcypher: You are right. Creating a new PR to fix that.

PR #369, opened 2022-04-01
Fix Initial Delay in DID lookup

appcypher:
## Summary

This PR fixes the following **bug**

* [x] Initial delay in DID lookup

PR #367 implements root DID lookup retrying but does not address the initial 2 sec delay from `setInterval` as noted by @icidasset. This PR fixes that.

## Test plan (required)

```sh
yarn test
```


icidasset (approved):
Thanks! ☺️ Looks good!

PR #374, opened 2022-04-07
updated depedencies, some vulnerabilities

therealjeffg:
I noticed there were some vulnerabilities reported by dependabot, this should resolve them. r- @bgins

### DIFF
--- a/package.json
+++ b/package.json
@@ -78,7 +78,7 @@
     "eslint": "^8.7.0",
     "expect": "^27.4.6",
     "fast-check": "^2.21.0",
-    "ipfs-core": "^0.13.0",
+    "ipfs-core": "^0.14.2",
     "ipfs-core-types": "^0.9.0",
     "ipfs-repo": "^14.0.1",
     "madge": "^5.0.1",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -78,7 +78,7 @@
     "eslint": "^8.7.0",
     "expect": "^27.4.6",
     "fast-check": "^2.21.0",
-    "ipfs-core": "^0.13.0",
+    "ipfs-core": "^0.14.2",
```
matheus23, line 81: This can be scary, but from looking at the [CHANGELOG](https://github.com/ipfs/js-ipfs/blob/master/packages/ipfs-core/CHANGELOG.md#0140-2022-01-27) it should be fine :+1:

matheus23 (approved):

bgins (approved):
Looks good! Thanks for the PR. 🎉 

The changelog check is likely failing because the "no changelog" tag wasn't applied when the PR was opened. Unfortunately, the action will continue to fail without another commit pushed to branch. 

Could you push a meaningless commit (add some whitespace somewhere) to get that check to pass? We will squash the commits so no worries about adding nonsense to the commit history.

PR #375, opened 2022-04-11
App owned: Add local IPFS implementation

bgins:
## Summary

* [ ] Add app-local IPFS node

Apps that use app-owned auth should not rely on the Fission Auth Lobby to provide them with a `js-ipfs` node. Instead, they should be able to run their own node.

### Task list

* [x] Add `localIps` option
* [ ] Add `Worker` implementation
* [ ] Run IPFS from worker when `localIpfs` option is true
* [ ] Report connection status events from worker

## After Merge


bgins:
Discussed in: https://talk.fission.codes/t/aol-resource-creation-planning-and-discussion/2715

bgins:
The initial version of this feature will only support browser environments. In the future, we can find a way to do something similar in `node`.

bgins:
Closed in favor or #404.

PR #376, opened 2022-05-03
Fix issue with loading shares

icidasset:
The CIDs from the share links (ie. DAG objects) aren't always strings, as expected before. This PR removes the strict string type check and just checks for existence instead.


bgins (approved):
Looks good! 🎉

PR #377, opened 2022-05-03
Bump ipfs dependencies

icidasset:
Tested with current lobby + drive and lobby + drive with upgraded ipfs dependencies (message protocol).


bgins (approved):
Tested this out across a number of apps, and everything seems to be fine with these changes. Here's what I tested:


- React TodoMVC
	- CRA
		- ✅  local dev
		- ✅  built app
	- Webpack
		- ✅ local dev
		- ✅  built app
	- Vite
		- ✅  local dev
		- ✅  built app
- `webnative-bundler tests`
	- ✅ Vite
	- ✅ Webpack
	- ✅ Estrella
	- ✅ Snowpack
- ✅ Sharing public files example (uses UMD build)
- ✅ Storing WASM example (uses Webpack)

Looks like we're good. Let's ship it!

PR #379, opened 2022-05-09
App owned: Add use WNFS implementation

### LINKED ISSUE #368: Move WNFS creation to webnative
# Summary

## Problem

App-owned accounts cannot easily initialize webnative resources for a user.

### Impact

Resources such as WNFS or platform publishing APIs are not easily available to apps without replicating code from the auth lobby.

## Solution

Write an API that sets up WNFS and other webnative resources. All resources should be opt-in.

# Detail

Resource creation currently happens in the auth lobby when an app redirects to request capabilities. Apps should be able to create and manage resources on their own.

All resources should be optional. The primary goal is functions for setting up webnative resources and a proof-of-concept that they work.

bgins:
## Summary

This PR makes it possible for apps to create WNFS without visiting the auth lobby.

* [x] Adds a `useWnfs` option that creates or loads WNFS in apps with root authority (app-owned auth apps)
* [x] Adds a `USE_WNFS_IMPLEMENTATION` to provide account linking for apps that `useWnfs`
* [x] Rename `LOCAL_IMPLEMENTATION` to `BASE_IMPLEMENTATION`

The features in the PR will only make sense for apps with root authority. We can make these features coherent for linked apps when we implement app-owned app linking. For now, apps that initialize webnative with `permissions` will ignore the `useWnfs` option.

### `useWnfs` option

The `useWnfs` option is passed to `webnative.initialize`:

``` typescript
wn.initialise({ useWnfs: true })
  .then(async (state: State) => {
    switch (state.scenario) {
      case wn.Scenario.Continuation:
        // state.fs available
        
      case wn.Scenario.NotAuthorised:
        // state.fs not available
    })
```

We check if the user has a data root in their CID log, on the server, or in DNS. If any of these exist, we load their existing filesystem. Otherwise, we create a new filesystem with `public` and `private` roots and assign it to the user, which includes creating (or using) a read key, self-authorizing a filesystem UCAN, publishing the data root, and updating the CID log. Taken together, we think of these steps as "bootstrapping" the filesystem.

Developers will bootstrap or load the filesystem in two cases after webnative has already been initialized:

- A user registers an account, and the developer bootstraps a filesystem for them
- A user links a device, and the developer loads their existing filesystem

We provide the developer `bootstrapRootFileSystem` and `loadRootFileSystem` as convenience functions so they do not need to be aware of permissions. Their app has root authority and needs no permission.

### `USE_WNFS_IMPLEMENTATION`

Account linking when using WNFS requires the exchange of a read key. Linking works similar to how it works in the auth lobby, but we also need to configure the linked device with a filesystem UCAN. The `USE_WNFS_IMPLEMENTATION` implements these pieces and must be when using `useWnfs`:

``` typescript
import { USE_WNFS_IMPLEMENTATION } from 'webnative/auth/implementation/use-wnfs'

setImplementations({ auth: USE_WNFS_IMPLEMENTATION.auth })
```

We may ultimately decide to set this implementation for developers automatically. For the moment, this may be confusing for app-owned apps because they will need to use the `BASE_IMPLEMENTATION` when they don't want to use WNFS. In the long run, it may make sense to require the `useWnfs` option in *all* apps that wish to use WNFS. This approach would mean `BASE_IMPLEMENTATION` becomes the default, and `USE_WNFS_IMPLEMENTATION` or other implementations are set during initialization as needed.

### Rename `LOCAL_IMPLEMENTATION` to `BASE_IMPLEMENTATION`

We've discussed elsewhere that the name `LOCAL_IMPLEMENTATION` didn't express its intent clearly enough. Updating it to `BASE_IMPLEMENTATION` makes it clear that the implementation provides a minimal starting point that can be built upon. The `USE_WNFS_IMPLEMENTATION` implementation is an example of building from the base implementation.

## Test plan (required)

Many portions of the code in this PR are effectful in ways that remain challenging to test in webnative. The `initialization` and `createFilesystem` functions rely on data root lookup and update, which use `fetch`. `createFilesystem` and the `USE_WNFS_IMPLEMENTATION` rely on `keystore-db`.

The `readKey` function that gets or creates the read key is tested in `src/filesystem.test.ts`.

## Closing issues

Closes #368.

## After Merge
* [ ] Does this change invalidate any docs or tutorials? No.
* [ ] Does this change require a release to be made? Not immediately, but we may want to create an alpha release.

### Task list

* [x] Add `useWnfs` option to webnative initialize
* [x] Add `createFilesystem` function to create WNFS 
* [x] Create or load WNFS on webnative initialization
* [x] Add `use-wnfs` auth implementation to authorize WNFS when linking accounts
* [x] Ensure backward compatibility for apps that use the auth lobby

### DIFF
--- a/src/auth/implementation/use-wnfs.ts
+++ b/src/auth/implementation/use-wnfs.ts
@@ -55,6 +55,7 @@ export const linkDevice = async (data: Record<string, unknown>): Promise<void> =
       issuer
     })
     await ucanInternal.store([ucan.encode(fsUcan)])
+
   } else {
     throw new LinkingError(`Consumer received invalid link device response from producer. Given ucan is invalid: ${data.ucan}`)
   }
--- a/src/auth/implementation/use-wnfs.ts
+++ b/src/auth/implementation/use-wnfs.ts
@@ -8,6 +8,27 @@ import { LinkingError } from "../linking.js"
 
 import RootTree from "../../fs/root/tree.js"
 
+
+export const checkCapability = async (username: string): Promise<boolean> => {
+  const readKey = await storage.getItem("readKey")
+  if (!readKey) return false
+
+  const didFromDNS = await did.root(username)
+  const maybeUcan: string | null = await storage.getItem("ucan")
+
+  if (maybeUcan) {
+    const rootIssuerDid = ucan.rootIssuer(maybeUcan)
+    const decodedUcan = ucan.decode(maybeUcan)
+    const { ptc } = decodedUcan.payload
+
+    return didFromDNS === rootIssuerDid && ptc === "SUPER_USER"
+  } else {
+    const rootDid = await did.write()
+
+    return didFromDNS === rootDid
+  }
+}
+
 export const delegateAccount = async (username: string, audience: string): Promise<Record<string, unknown>> => {
   const readKey = await storage.getItem("readKey")
   const proof = await storage.getItem("ucan") as string
@@ -77,7 +98,7 @@ export const USE_WNFS_IMPLEMENTATION = {
     isUsernameValid: base.isUsernameValid,
     isUsernameAvailable: base.isUsernameAvailable,
     createChannel: base.createChannel,
-    checkCapability: base.checkCapability,
+    checkCapability,
     delegateAccount,
     linkDevice
   }
--- a/src/auth/implementation/use-wnfs.ts
+++ b/src/auth/implementation/use-wnfs.ts
@@ -1,0 +1,79 @@
+import * as base from "./base.js"
+import * as check from "../../common/type-checks.js"
+import * as did from "../../did/index.js"
+import * as storage from "../../storage/index.js"
+import * as ucanInternal from "../../ucan/internal.js"
+import * as ucan from "../../ucan/token.js"
+import { LinkingError } from "../linking.js"
+
+export const delegateAccount = async (username: string, audience: string): Promise<Record<string, unknown>> => {
+  const readKey = await storage.getItem("readKey")
+  const proof = await storage.getItem("ucan") as string
+
+  // UCAN
+  const u = await ucan.build({
+    audience,
+    issuer: await did.write(),
+    lifetimeInSeconds: 60 * 60 * 24 * 30 * 12 * 1000, // 1000 years
+    potency: "SUPER_USER",
+    proof,
+
+    // TODO: UCAN v0.7.0
+    // proofs: [ await localforage.getItem("ucan") ]
+  })
+
+  return { readKey, ucan: ucan.encode(u) }
+}
+
+function isUseWnfsLinkingData(data: unknown): data is { readKey: string; ucan: string } {
+  return check.isObject(data)
+    && "readKey" in data && typeof data.readKey === "string"
+    && "ucan" in data && typeof data.ucan === "string"
+}
+
+export const linkDevice = async (data: Record<string, unknown>): Promise<void> => {
+  if (!isUseWnfsLinkingData(data)) {
+    throw new LinkingError(`Consumer received invalid link device response from producer: Expected read key and ucan, but got ${JSON.stringify(data)}`)
+  }
+
+  const { readKey, ucan: encodedToken } = data
+  const u = ucan.decode(encodedToken as string)
+
+  if (await ucan.isValid(u)) {
+    await storage.setItem("ucan", encodedToken)
+    await storage.setItem("readKey", readKey)
+
+    // Create and store filesystem UCAN
+    const issuer = await did.write()
+    const fsUcan = await ucan.build({
+      potency: "APPEND",
+      resource: "*",
+      proof: encodedToken,
+      lifetimeInSeconds: 60 * 60 * 24 * 30 * 12 * 1000, // 1000 years
+
+      audience: issuer,
+      issuer
+    })
+    await ucanInternal.store([ucan.encode(fsUcan)])
+  } else {
+    throw new LinkingError(`Consumer received invalid link device response from producer. Given ucan is invalid: ${data.ucan}`)
+  }
+}
+
+
+
+// 🛳
+
+
+export const USE_WNFS_IMPLEMENTATION = {
+  auth: {
+    init: base.init,
+    register: base.register,
+    isUsernameValid: base.isUsernameValid,
+    isUsernameAvailable: base.isUsernameAvailable,
+    createChannel: base.createChannel,
+    checkCapability: base.checkCapability,
+    delegateAccount,
+    linkDevice
+  }
+}
--- a/src/auth/implementation/use-wnfs.ts
+++ b/src/auth/implementation/use-wnfs.ts
@@ -6,6 +6,8 @@ import * as ucanInternal from "../../ucan/internal.js"
 import * as ucan from "../../ucan/token.js"
 import { LinkingError } from "../linking.js"
 
+import RootTree from "../../fs/root/tree.js"
+
 export const delegateAccount = async (username: string, audience: string): Promise<Record<string, unknown>> => {
   const readKey = await storage.getItem("readKey")
   const proof = await storage.getItem("ucan") as string
@@ -42,6 +44,8 @@ export const linkDevice = async (data: Record<string, unknown>): Promise<void> =
   if (await ucan.isValid(u)) {
     await storage.setItem("ucan", encodedToken)
     await storage.setItem("readKey", readKey)
+
+    await RootTree.storeRootKey(readKey)
 
     // Create and store filesystem UCAN
     const issuer = await did.write()
--- a/src/filesystem.ts
+++ b/src/filesystem.ts
@@ -189,6 +189,9 @@ async function addSampleData(fs: FileSystem): Promise<void> {
 // 🔑
 
 
+/**
+ * Create or get a read key for accessing the WNFS private root.
+ */
 export async function readKey(): Promise<string> {
   const maybeReadKey = await storage.getItem("readKey") as unknown as string
   if (maybeReadKey) return maybeReadKey
--- a/src/filesystem.ts
+++ b/src/filesystem.ts
@@ -102,11 +102,13 @@ export async function loadFileSystem(
 
 
 /**
- * Create a filesystem
+ * Create a new filesystem and assign it to a user.
+ *
+ * Warning: This function will override a user's filesystem with a empty one.
  *
  * @param permissions The permissions to initialize the filesystem
  */
-export const createFilesystem = async (permissions: Permissions): Promise<FileSystem> => {
+export const installFileSystem = async (permissions: Permissions): Promise<FileSystem> => {
   // Get or create root read key
   const rootKey = await readKey()
 
--- a/src/filesystem.ts
+++ b/src/filesystem.ts
@@ -8,11 +8,11 @@ import * as crypto from "./crypto/index.js"
 import * as debug from "./common/debug.js"
 import * as dataRoot from "./data-root.js"
 import * as did from "./did/index.js"
+import * as token from "./ucan/token.js"
 import * as ucan from "./ucan/internal.js"
 import * as protocol from "./fs/protocol/index.js"
 import * as versions from "./fs/versions.js"
 
-import { build as buildUcan, encode as encodeUcan } from "./ucan/token.js"
 import { Branch } from "./path.js"
 import { Maybe, authenticatedUsername, decodeCID } from "./common/index.js"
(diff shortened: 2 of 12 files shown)


src/index.ts
```diff
@@ -48,13 +52,42 @@ export async function initialise(options: InitOptions): Promise<State> {
...
+        permissions,
+        authedUsername,
+        await maybeLoadFs(authedUsername)
+      )
     } else {
       return scenarioNotAuthorised(permissions)
     }
 
+  } else if (authedUsername && useWnfs) {
+    if (options.loadFileSystem === false) throw new Error("Must load filesystem when using the useWnfs option.")
```
bgins, line 65: Using WNFS only makes sense when loading the filesystem.

icidasset: I'm not entirely sure about having this check here. I would remove it for flexibility. If I wanted to load my filesystem in a web worker instead, I'd use the `loadFileSystem` option, I imagine we want to still do that with `useWnfs`?

src/filesystem.ts
```diff
@@ -97,6 +101,46 @@ export async function loadFileSystem(
...
+
+  // Update filesystem and publish data root
+  const rootCid = await fs.root.put()
+  const res = await dataRoot.update(rootCid, token.encode(fsUcan))
+  if (!res.success) console.warn("Failed to update data root.")
+
+  // Update CID log
+  await cidLog.add(rootCid.toString())
+
+  return fs
```
icidasset, line 140: We should also add the sample data here (see load filesystem func).

src/filesystem.ts
```diff
@@ -97,6 +101,46 @@ export async function loadFileSystem(
 }
 
 
+/**
+ * Create a filesystem
+ *
+ * @param permissions The permissions to initialize the filesystem
+ */
+export const createFilesystem = async (permissions: Permissions): Promise<FileSystem> => {
```
icidasset, line 109: Same comment as with the `useWnfs` naming, I think this function name might be confusing. The `loadFileSystem` function above actually can create a filesystem as well, but this function here does a few additional things.

There's a few use-cases in which you want to create a file system separate from your account's file system. This function would be confusing, since it directly overrides your account's filesystem. I think it should be called `assignNewSelfContainedFileSystem()` or something like that.

src/filesystem.ts
```diff
@@ -138,3 +182,17 @@ async function addSampleData(fs: FileSystem): Promise<void> {
   await fs.mkdir({ directory: [ Branch.Private, "Video" ] })
   await fs.publish()
 }
+
+
+
+// 🔑
+
+
+export async function readKey(): Promise<string> {
```
icidasset, line 191: Probably could use some documentation to indicate that this read key is a top-level read key used with auth. Webnative also stores other read keys in the key store which are tied to specific private paths.

icidasset (commented):
I have some concerns around naming and a few missing pieces, but this is looking great otherwise 👌 Very cool that we can link between apps with a WNFS now! 🎉

src/init/types.ts
```diff
@@ -19,4 +19,5 @@ export type InitOptions = {
   autoRemoveUrlParams?: boolean
   loadFileSystem?: boolean
   rootKey?: string
+  useWnfs?: boolean
```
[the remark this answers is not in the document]
bgins: Good question, and something that I wrestled with quite a bit. 

I'm writing up a proposal for some broader changes to initialization where this will make more sense. In short, all apps that want to use WNFS would set this option to true.

For this PR, perhaps we could do something to make it clear that apps don't need to use this option if they initialize with `permissions`. Maybe throw an exception when `useWnfs` is true and `permissions` are used?

src/auth/implementation/use-wnfs.ts
```diff
@@ -0,0 +1,80 @@
...
+  if (!isUseWnfsLinkingData(data)) {
+    throw new LinkingError(`Consumer received invalid link device response from producer: Expected read key and ucan, but got ${JSON.stringify(data)}`)
+  }
+
+  const { readKey, ucan: encodedToken } = data
+  const u = ucan.decode(encodedToken as string)
+
+  if (await ucan.isValid(u)) {
+    await storage.setItem("ucan", encodedToken)
+    await storage.setItem("readKey", readKey)
```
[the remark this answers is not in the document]
bgins: Yep, in part for backward compatibility. Also, non-auth-lobby apps may become authorizers when we implement app-owned app linking. I'd propose we revisit these names then.

src/index.ts
```diff
@@ -63,12 +96,6 @@ export async function initialise(options: InitOptions): Promise<State> {
 }
 
 
-/**
- * Alias for `initialise`.
- */
-export { initialise as initialize }
```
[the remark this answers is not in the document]
bgins: Oh whoops, completely by accident! Will put that back, thanks for catching that. 😅

src/filesystem.ts
bgins: What are the use cases where you want to create a filesystem separate from your account's filesystem? I see one case where we use it for testing. Are there others?

Perhaps `createFilesystem` could be modified to support the other use cases.

src/filesystem.ts
bgins: Discussed this with @therealjeffg. The decision was to only provide `public` and `private` directories by default in app-owned apps.

src/filesystem.ts
icidasset: The other main use case I had mind is when you have a temporary file system, ie. a file system for a user of the app that doesn't have a Fission account yet.

Maybe add an option to auto publish, so we can avoid accidental data loss situations? 🤷

src/filesystem.ts
```diff
@@ -97,6 +101,46 @@ export async function loadFileSystem(
...
+  // Add filesystem UCAN to store
+  await ucan.store([token.encode(fsUcan)])
+
+  // Update filesystem and publish data root
+  const rootCid = await fs.root.put()
+  const res = await dataRoot.update(rootCid, token.encode(fsUcan))
+  if (!res.success) console.warn("Failed to update data root.")
+
+  // Update CID log
+  await cidLog.add(rootCid.toString())
```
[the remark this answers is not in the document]
bgins: Yeah, good call! Added that. 👍

src/init/types.ts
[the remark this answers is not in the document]
bgins: Yep, it will be a bit awkward at first, but it should be be backward compatible while we transition into a refactored API. Added a comment to `initialise` to document it.

src/filesystem.ts
bgins: Would this work be accurate?

https://github.com/fission-suite/webnative/blob/514a380480eaa599f88f986a901769a1c5aa94b4/src/filesystem.ts#L192-L195

My understanding is the read key can be used outside of an auth scenario to read the private filesystem with root access. Is that correct? If not, I may need to add more read keys for this PR to work.

src/filesystem.ts
```diff
@@ -97,6 +101,46 @@ export async function loadFileSystem(
...
+    issuer
+  })
+
+  // Add filesystem UCAN to store
+  await ucan.store([token.encode(fsUcan)])
+
+  // Update filesystem and publish data root
+  const rootCid = await fs.root.put()
+  const res = await dataRoot.update(rootCid, token.encode(fsUcan))
+  if (!res.success) console.warn("Failed to update data root.")
```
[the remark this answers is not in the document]
bgins: Yep, I see that also does the `put` internally and returns the CID. Replaced it with `fs.publish`. 👍

src/index.ts
bgins: Ah right, great point! Forgot about that use case. And indeed, it is more flexible to be able to load the filesystem after initialization. 👌

src/filesystem.ts
icidasset: That's accurate and correct 👍 Other read keys are only received when coming back from the lobby with certain authorised permissions.

src/index.ts
```diff
@@ -48,13 +52,42 @@ export async function initialise(options: InitOptions): Promise<State> {
...
+      return scenarioContinuation(
+        rootPermissions,
+        authedUsername,
+        await createFilesystem(rootPermissions)
+      )
+    } else {
+      return scenarioContinuation(
+        rootPermissions,
+        authedUsername,
+        await maybeLoadFs(authedUsername),
```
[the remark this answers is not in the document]
bgins: Indeed, this was missing but we may only need it here:

https://github.com/fission-suite/webnative/blob/93d34534ee666c3123e7b206ead69fb6ca893869/src/auth/implementation/use-wnfs.ts#L48

If I understand correctly, we only need to store the root key once before we first load a filesystem from a CID. Adding it to the device linking implementation stores the key at linking time. The developer can load the filesystem after linking.

The key will either be stored when creating the filesystem or on linking a device. On subsequent visits, the key should be stored and ready for use.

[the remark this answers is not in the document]
icidasset: I agree, that'd be a better place to put it 👍

src/filesystem.ts
bgins: I gave this some thought, and I think that renaming this function is the way to go for now. What would you think about the name `initializeFilesystem`? Or maybe `setupFilesystem`? We can make its effects explicit in the documentation comment.

I'd prefer a less verbose name because developers will call this function directly after registering a user. At the point of registration, webnative will already be initialized, and we probably want to avoid requiring a reload to setup the filesystem.

icidasset: > I'd prefer a less verbose name because developers will call this function directly after registering a user. At the point of registration

Makes sense, in that case I'd pick `setupFileSystem`. But if I had to pick a single word, I'd probably pick something like `installFileSystem` or `establishFileSystem` (because 'setup' is too generic and too similar to 'load').

bgins:
Added a `checkCapability` to the auth implementation:

https://github.com/fission-suite/webnative/blob/f6d7700ca67147553a419a4ff2980aeca655bad3/src/auth/implementation/use-wnfs.ts#L12-L30

The reason is we need a stricter check than the base implementation, which doesn't verify the read key is set.

src/filesystem.ts
bgins: Yep, agreed that setup is too generic. 👍 

Updated it to `installFileSystem` and added some comments to it. We can go with that for now and change it later if we come up with something else.

bgins: Was chatting with @jessmartin about this. What would you think about `bootstrapFileSystem` instead of `installFileSystem`?

expede: > There's a few use-cases in which you want to create a file system separate from your account's file system. This function would be confusing, since it directly overrides your account's filesystem. I think it should be called assignNewSelfContainedFileSystem() or something like that.

Random suggestion from the process-based world (not sure if this captures all of the functionality): `startFileSystem` or `spawnFileSystem` (kind of like [`start_link`](https://hexdocs.pm/elixir/Supervisor.html#start_link/2) or [`spawn_link`](https://hexdocs.pm/elixir/Kernel.html#spawn_link/3))

icidasset: > Was chatting with @jessmartin about this. What would you think about `bootstrapFileSystem` instead of `installFileSystem`?

Technically speaking `establish` is more correct, but I like bootstrap as well 👍

bgins:
Added a couple of convenience functions for bootstrapping and loading root filesystems:

https://github.com/fission-suite/webnative/blob/9f76131204360cd44bc306c095081d18b3af3f73/src/filesystem.ts#L177-L193

The goal is to simplify these calls in apps that have root authority and don't need to know about the details of permissions.

matheus23:
Let me know when this is ready for another round of review :)

icidasset (approved):
🎉 Looks good!

matheus23 (approved):
Nice. Looking good! I like the added APIs in `filesystem.ts`.
:shipit:

appcypher:
@bgins How do I run the tests for this?

appcypher (approved):
This is not my area of expertise yet but after following the discussion for better context understanding, this looks good to me. I just have a single nitpick.

src/filesystem.ts
```diff
@@ -138,3 +206,20 @@ async function addSampleData(fs: FileSystem): Promise<void> {
...
+
+
+// 🔑
+
+
+/**
+ * Create or get a read key for accessing the WNFS private root.
+ */
+export async function readKey(): Promise<string> {
+  const maybeReadKey = await storage.getItem("readKey") as unknown as string
```
appcypher, line 219:
```suggestion
  const maybeReadKey: string | null = await storage.getItem("readKey")
```

bgins: Yeah, that is better. Accounts for the possible `null` 👍

PR #387, opened 2022-07-06
Update nix-shell to use flakes

walkah:
The nix-shell setup had gotten stale (and lorri was unable to evaluate on my new machine). This PR:
* updates to use nix flakes
* pins dependencies to the nix 22.05 release (which defaults to nodejs 16x)
* adds flake-compat layer for shell.nix. 
* removes nix directory / niv


bgins (approved):
Looks good and working well over here! ✨ 🚀

PR #388, opened 2022-07-20
Add multiple initialization entrypoints

### LINKED ISSUE #390: Separate intialize into multiple entrypoints
# Summary

## Problem

The API for app-owned auth apps is awkward because developers need to import and set implementations.

### Impact

The API is harder to document and the extra steps may be confusing to developers.

## Solution

Separate `webnative.initiase` into `webnative.app` (app-owned auth) and `webnative.permissionedApp` (existing apps and any future apps that request authorization from another app).

# Detail

**Additional context**

Separating the entrypoints will also free us up for internal refactoring. We could for example pass endpoints to `app` or `permissionedApp` instead of setting them before initialization.

We can also add more entrypoints in the future. For example, it might make sense to add a `walletApp` which sets implementations for apps that use wallet auth.

bgins:
## Summary

* [x] Separates initialization entrypoints for app-owned and linked apps
* [x] Adds new `AppState`, `AppScenario` and `AppInitOptions` types for app-owned apps
* [x] Moves the lobby auth implementation with the other implementations

This PR is a proof-of-concept for adding multiple entrypoints to webnative. The goal is to give ourselves flexibility in the way we initialize webnative. Rather than overloading a single entrypoint, we can have many entrypoints that each fulfill a specific purpose.

In this PR, app-owned initialization is separated out from the original `initialise` function. The `initialise` function is renamed internally to ~~`fissionApp` (open to naming changes!)~~ `permissionedApp` (we have a better name!), which sets the auth implementation to use the auth lobby. App-owned initialization uses the `app` function, which takes a `use-wnfs` flag to initialize with WNFS or not and to set an appropriate auth implementation.

The entrypoints use different state types. App-owned apps will never be linked from another app, which means their state type can be simplified:

https://github.com/fission-suite/webnative/blob/fd4077962c567687f826541b43352ad9550e019b/src/auth/state/app.ts#L41-L43

Apps that are linked continue to use the original `State` type, here renamed to reflect that they are ~~linked~~ permissioned:

https://github.com/fission-suite/webnative/blob/08966f9bac5f5eadc0a8ecf805397af9739fc225/src/auth/state/permissionedApp.ts#L87-L91

Naming suggestions welcome here! Not entirely sure about the best way to separate these out.

_See discussion below. Originally this was `LinkedAppState`, but we think `PermissionedAppState` will be better._

The init options types are also separated out because each entrypoint has different needs:

https://github.com/fission-suite/webnative/blob/08966f9bac5f5eadc0a8ecf805397af9739fc225/src/init/types.ts#L17-L33

Backward compatibility for existing apps is maintained in this PR. In my opinion, if we go forward with this change, we should schedule a breaking change because `initialise` and `initialize` will no longer make sense when considering the overall API. (Same goes for `State` and `Scenario`.)

## Test plan (required)

Existing apps that use the auth lobby should work without any changes.

An app-owned app can be testsed here: https://github.com/fission-suite/aol-demo/pull/1

## Closing issues

Closes #390 

## After Merge
* [ ] Does this change invalidate any docs or tutorials? Not yet, but changes should be scheduled if we deprecate the initialise function.
* [ ] Does this change require a release to be made? Yes, it will.


src/index.ts
```diff
@@ -21,15 +89,15 @@ import FileSystem from "./fs/index.js"
  * See `loadFileSystem` if you want to load the user's file system yourself.
  * NOTE: Only works on the main/ui thread, as it uses `window.location`.
  *
- * The `useWnfs` option is only necessary for apps that create their own WNFS. Apps that
- * request filesystem `permissions` from the auth lobby can ignore `useWnfs` in this version
- * of webnative.
  */
-export async function initialise(options: InitOptions): Promise<State> {
+export async function fissionApp(options: LinkedAppInitOptions): Promise<LinkedAppState> {
```
bgins, line 93: Renaming of the existing initialize function. 

Not sure about the name! How do we name an app to make it clear that it uses the Fission Auth Lobby?

src/index.ts
```diff
@@ -7,12 +7,80 @@ import * as dataRoot from "./data-root.js"
...
+ * NOTE: Only works on the main/ui thread, as it uses `window.location`.
+ *
+ */
+export async function app(options: AppInitOptions): Promise<AppState> {
+  options = options || {}
+
+  const { useWnfs = false } = options
+
+  if (useWnfs) {
+    setImplementations(USE_WNFS_IMPLEMENTATION)
```
bgins, line 41: Dependecy-injected implementations are set internally. The developer does not need to be aware of them unless they have an advanced use case.

src/index.ts
```diff
@@ -7,12 +7,80 @@ import * as dataRoot from "./data-root.js"
...
+  } else {
+    setImplementations(BASE_IMPLEMENTATION)
```
bgins, line 43: We could eventually make the base implementation the default because it assumes the least of any implementation.

src/index.ts
```diff
@@ -56,56 +125,35 @@ export async function initialise(options: InitOptions): Promise<State> {
...
 }
 
 
 /**
- * Alias for `initialise`.
+ * Alias for `fissionApp`.
  */
-export { initialise as initialize }
+ export { fissionApp as initialise }
+ export { fissionApp as initialize }
```
bgins, line 156: Maintain backward compatibility for now.

src/index.ts
```diff
@@ -133,9 +181,11 @@ export * from "./filesystem.js"
 export * from "./common/version.js"
 
 export const fs = FileSystem
-export { Scenario, State } from "./auth/state.js"
-export { AuthCancelled, AuthSucceeded, Continuation, NotAuthorised } from "./auth/state.js"
-export { InitialisationError, InitOptions } from "./init/types.js"
+export { AppScenario, AppState } from "./auth/state/app.js"
+export { AuthCancelled, AuthSucceeded, Continuation, NotAuthorised } from "./auth/state/linkedApp.js"
+export { Authed, NotAuthed } from "./auth/state/app.js"
+export { LinkedAppScenario as Scenario, LinkedAppState as State } from "./auth/state/linkedApp.js"
```
bgins, line 187: Maintain backward compatibility for now.

src/index.ts
icidasset: What makes an app a Fission app? An AOL app still registers a user with the Fission servers and stores data on it, so I guess that's a Fission app too?

Maybe we name this according to what the lobby gives us? The lobby serves as a "safety zone", in that people can trust the lobby so that it can provide (limited) access to subsets of their data to other apps. You could replicate that behaviour in different ways, but this function expects a specific way of doing so.

Maybe `lobbyApp` makes sense? A lobby app is an app that redirects to a root authority app and comes back with all necessary UCANs and keys to provide our app access to the data, this in the specific lobby format.

That makes me think, what if you have two AOL apps, where one delegates to the other. One would the root authority, and the second wants to have access to a subset of the user's data. How would that process work? Does (1) "upgrade" to the Fission lobby, or does it serve as something similar to the lobby for (2)? What would the init function be called for app (2)?

Sorry, thinking out loud 😅

bgins: Yeah, that's a great point! AOL apps are still Fission apps when they use the Fission server.

`lobbyApp` does seem like a better name. This function could still set some other backend by setting endpoints, so this should be more generic. Though it might be confusing because should a `lobbyApp` be a lobby?

Another possibility is `linkedApp`, which hints at the fact that it needs needs to be linked by some app with root authority. (Though in theory, it could be linked by any app with equal or greater authority. Future possibilities here of course 😅 .)

> That makes me think, what if you have two AOL apps, where one delegates to the other. One would the root authority, and the second wants to have access to a subset of the user's data. How would that process work? Does (1) "upgrade" to the Fission lobby, or does it serve as something similar to the lobby for (2)? What would the init function be called for app (2)?

(1) would use the `app` initialization function which would start it off with root authority, and (2) would use `lobbyApp`/`linkedApp`/`someOthername` to initialize with permissions to request from (1). And yeah, (1) has the same level of authority as the auth lobby, so it could be used to serve that role (after we've AOL app linking).

bgins: Another possibility, `permissionedApp`? Or maybe there is something else along those lines.

icidasset: > (1) would use the ...

Yes exactly 👍 I do like the sound of `permissionedApp`, with it having an option called `permissionedBy: URL` option 🤔

src/index.ts.bak
```diff
@@ -0,0 +1,154 @@
+import localforage from "localforage"
```
icidasset, line 1: File to be removed? 👀

src/auth/state/app.ts
```diff
@@ -0,0 +1,61 @@
...
+}
+
+
+
+// STATE
+
+
+export type AppState
+  = NotAuthed
+  | Authed
```
icidasset, line 42: The only inconsistency here is that "continuation" is missing (ie. was authed before, continuing existing sessions). Was that intentional?

icidasset (approved):

src/index.ts
bgins: Yeah for sure! Not on this PR, but we could refactor the way that endpoints are handled so they passed into initialization functions instead of set globally. That would make it explicit who a `permissionedApp` is `permissionedBy`.

Let's change this to `permissionedBy`, and if we come up with something else we can change again.

src/index.ts.bak
bgins: Ah yes, it should be! Removed it. 👌

src/auth/state/app.ts
bgins: Yep, intentional. `Authed` is roughly `Continuation`.

We don't need an `AuthSucceeded` because the user will register without a page load and re-initialization of webnative. `Continuation` can be renamed to `Authed` because we don't need to differentiate it from `AuthSucceeded`.

PR #389, opened 2022-07-21
Use rs-wnfs in webnative under a setup flag

matheus23:
This is a set of changes that should allow running both WNFS version 2.0.0 and WNFS version 3.0.0.
There is now a setup setting `setup.fsVersion({ version: "3.0.0" })` that will ensure that any newly created filesystem will be backed by rs-wnfs.
This only affects the public tree for now.

There's bigger blockers for this right now:
- We need to figure out garbage collection of wasm-owned objects: https://github.com/WebNativeFileSystem/rs-wnfs/issues/40
- We need to figure out bundling wasm with esbuild/ship the wasm to some kind of CDN/load it from ipfs/whatever

Here's a list of changes in this PR:
- Add a `PublicTreeWasm` class that extends `UnixTree` and `Puttable`, so can be a replacement for `RootTree`'s `publicTree` field
- Split up callback in `runOnNode`
- remove `onUpdate` callback in `mkdir` (it's never used)
- some updates to `instanceOf` calls, less `any` types, more `unknown`
- Fix a bunch of eslint warnings & add `_` as unused var pattern
- Split `runOnTree` into mutating and querying parts
- add `historyStep` function to the filesystem & use it after all mutations in the tests
- add `instanceOf` to `BareTree` and a field that allows us to check that
- add some tests specifically testing `PublicTreeWasm` in wasm/public.test.ts
- changed the return type of `UnixTree.get` to be `Promise<PuttableUnixTree | File | null>` instead of `Tree`, because `Tree` exposes some internal functions like `updateChildTree` (and they don't make sense with the rs-wnfs stuff)

For now you can't run this yourself, this PR depends on a private fork of `rs-wnfs` that I've built locally, but we'll get there!


matheus23:
I don't think we want this to be published yet. I think I can get everything in terms of testing out of this at its current form and the public WNFS structure is still somewhat in-flux and I don't want to waste the `3.0.0` version identifier on that.

matheus23:
We don't have to waste the `3.0.0` version identifier. It'll just change under our (experimental!) user's feet until it's stable.
Verison 3.0.0 won't be default for a while, and it'll be explicitly experimental. So use at your own risk.

gotjoshua:
So what is the exact procedure to enable the experimental features and try this out?

matheus23:
I tested this in drive & the auth lobby locally, and it seems to be running smoothly (& faster!).
Ready to be reviewed :)

matheus23:
> So what is the exact procedure to enable the experimental features and try this out?

In your `package.json`, add a dependency on this PR:
```json
{
  "dependencies": {
    "webnative": "https://github.com/fission-codes/webnative#matheus23/wnfs-wasm"
    ...
  }
  ...
}
```

Before you initialize webnative, add a call to `setup.fsVersion`:
```js
import * as webnative from "webnative"

// ...

webnative.setup.fsVersion({ version: "3.0.0" })

// ...

webnative.initialise(...).then(...)
```

Unless you're building an "AOL" (App owned login) app right now, you'd also need to fork the [auth lobby](https://github.com/fission-codes/auth-lobby) and configure it with the same instructions above. And you'd need to create a new account on that forked auth lobby and connect your app to that auth lobby, while being careful about staging vs. production environments on our infra.

So maybe it's best to wait until we've pushed an update of the auth lobby to staging :v:

Another option would be to just try running the file system locally completely offline. To do that create a new file system `new webnative.FileSystem({ localOnly: true, ... })` instead of calling `webnative.initialise`.

---

In any case, we'll notify you once we've published this webnative version to the staging auth lobby @gotjoshua :+1:

src/fs/filesystem.ts
```diff
@@ -336,24 +395,36 @@ export class FileSystem {
...
+      },
+      private: async (node, relPath) => {
+        if (typeCheck.isFile(node)) {
+          throw new Error("Tried to `mv` within a file")
+        }
 
-      const [ _, ...nextPath ] = pathing.unwrap(to)
-      return node.mv(relPath, nextPath)
+        const [_, ...nextPath] = pathing.unwrap(to)
+        // TODO FIXME: nextPath is wrong if you use a node that's deeper in the tree.
```
icidasset, line 409: Fix for after this PR?

src/fs/filesystem.ts
```diff
@@ -624,86 +735,99 @@ export class FileSystem {
...
+  async checkMutationPermissionAndAddProof(path: DistinctivePath, isMutation: boolean): Promise<void> {
+    const operation = isMutation ? "make changes to" : "query"
 
     if (!this.localOnly) {
       const proof = await ucan.dictionary.lookupFilesystemUcan(path)
       const decodedProof = proof && ucan.decode(proof)
 
       if (!proof || !decodedProof || ucan.isExpired(decodedProof) || !decodedProof.signature) {
-        throw new NoPermissionError(`I don't have the necessary permissions to ${operation} the file system at "${pathing.toPosix(path)}"`)
+        throw new NoPermissionError(`I don't have the necessary permissions ${operation} to the file system at "${pathing.toPosix(path)}"`)
```
icidasset, line 746: Should stay as is I think? Change would result in "permissions make changes to to"

src/fs/v3/IpfsBlockStore.ts
```diff
@@ -0,0 +1,62 @@
+import * as uint8arrays from "uint8arrays"
+import { IPFS } from "ipfs-core"
```
icidasset, line 2: Better to do `import type` so we don't need to add `ipfs-core` to the dependencies (unless Typescript removes this too?). Also, I've replaced this with `ipfs-core-types` everywhere, because it was creating issues with the unstable packages for ipfs-core and ipfs-core-types I decided to use (to avoid bundler problems which the currently stable versions have).

src/fs/v3/PublicRootWasm.ts
```diff
@@ -0,0 +1,370 @@
...
+  async ls(path: Path): Promise<Links> {
+    const root = await this.root
+    const { result: entries } = await root.ls(path, this.store) as OpResult<DirEntry[]>
+    const result: Links = {}
+    for (const entry of entries) {
+      result[entry.name] = {
+        name: entry.name,
+        isFile: entry.metadata.unixMeta.kind === "file",
+        size: 0, // TODO size?
+        cid: "not provided for performance ", // TODO do we really need a CID here?
```
icidasset, line 109: I think Drive depends on this. Why is this a performance win?

tests/helpers/in-memory-ipfs.ts
```diff
@@ -1,15 +1,26 @@
...
+import { BlockCodec } from "multiformats/codecs/interface"
 
+const CODECS: Record<string, BlockCodec<number, unknown>> = {
+  [dagPB.code]: dagPB,
+  [dagPB.name]: dagPB,
+  [dagCBOR.code]: dagCBOR,
+  [dagCBOR.name]: dagCBOR,
+  [raw.code]: raw,
+  [raw.name]: raw,
+}
```
icidasset, line 23: Why does this have the `.name`s as well, compared to what's in [src/fs/v3/IpfsBlockStore.ts](https://github.com/fission-codes/webnative/pull/389/files#diff-5384c7b03d3c6c7419ebd93f9a1868887231145c4484c7078f87febca960c415)? Also, I guess we should add `[raw.code]: raw` to the our new `ipfs/node/repo.ts` (which I added with my PR that just got merged)?

icidasset (approved):
Aww yisss WASM 😎


Looks great! Just noted a few small things, some questions and nitpicks.

src/fs/filesystem.ts
matheus23: Yeah, IIRC, when I was going though the code I noticed this was probably the source of this bug: https://github.com/fission-codes/webnative/issues/238

I can try that in another PR after this :)

src/fs/filesystem.ts
matheus23: Oh wow. Thanks for catching that. Must've been some multi-cursor thing messing this up!

src/fs/v3/PublicRootWasm.ts
matheus23: This requires us to serialize the directory entries in order to hash them. If we always provide a CID, we will always force a serialization.

I guess we should rather include it than leaving it out!

matheus23: I should really change the rs-wnfs `ls` API back from returning metadata to just returning the CIDs of nodes.

matheus23: In the mean time I'm making this get the CID of entries from rs-wnfs :+1:

matheus23: Oh, and I just tested drive, apparently it doesn't depend on that? It works fine like that. Will still change it.

icidasset: Ah nice 👍 Can it still copy the CID as well? (via the context menu of a public item)

tests/helpers/in-memory-ipfs.ts
matheus23: The `CODECS` list in `IpfsBlockStore` is something that we manage between JS and rs-wnfs, it doesn't touch js-ipfs.
`rs-wnfs` only speaks codec codes, not codec names, so it doesn't need entries for e.g. `raw.name`.
I've taken this configuration from some issue comment that achingbrain made *somewhere*, where he's configuring a totally in-memory js-ipfs node. I can't find that comment anymore though. But it seems like js-ipfs expects there to be entries for both `codec.code` and `codec.name`.

Yep, having `raw` in there would be good!

src/fs/v3/PublicRootWasm.ts
matheus23: Yep! The listing in drive actually uses `ipfs.files.stat` on the pretty tree.

While I was looking at the code I also saw some usage of `fs.get` in `resolveItem` that refers to the `cid`, but I think that logic isn't used anywhere. And if it is, it's faulty:

```js
const cid = resolved.cid || resolved.header.content || resolved.header.bareNameFilter
```

The `bareNameFilter` shouldn't be a replacement for a cid.

---

In any case, I still went ahead and fixed this :+1:

PR #394, opened 2022-07-27
Update `register` to make email optional

### LINKED ISSUE #393: Make email optional at registration
The Fission server supports user registration with or without an email. At the time of writing, this change is live on staging. See https://github.com/fission-suite/fission/pull/619 for more details.

The `register` function used to register users requires an email. We should relax this constraint by making email an optional parameter in this function.

bgins:
## Summary

* [x] Make email optional at registration

This PR makes the email parameter to the `register` function optional. It goes along with the changes on the server (https://github.com/fission-suite/fission/pull/619) where a user can register with a only a username _or_ a username and email.

## Test plan (required)

Tested by registering a user with and without an email.

## Closing issues

Closes #393 

## After Merge
* [x] Does this change invalidate any docs or tutorials? No, but we need new ones in the future.
* [ ] Does this change require a release to be made? Yes, I will release an alpha.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,8 +4,9 @@
 
 #### Features
 
-- Adds app owned WNFS.
+- Adds app owned WNFS
 - Separates initialize into app and permissionedApp entrypoints
+- Make email optional at registration
 
 ### v0.32.0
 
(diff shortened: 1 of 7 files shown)


walkah (approved):
thumbs-gif-by-sesame-street

Looks good to me!

PR #397, opened 2022-08-01
Add filesystem reset protection

### LINKED ISSUE #396: Add filesystem reset protection
# Summary

## Problem

The current implementation of `bootstrapFileSystem` resets an existing filesystem.

### Impact

Developers might call this function and accidentally reset a filesytem. It's a bit of a footgun.

## Solution

Add a `reset` option that defaults to false but can be set to true when a reset is desired.

bgins:
## Summary

* [x] Adds a `reset` option to `bootstrapFileSystem`
* [x] Adds a check for authed user in `bootstrapFileSystem`

The existing `boostrapFileSystem` function will override a user's filesystem with an empty one. This function is dangerous and could use some guard rails.

When the `reset` option is set to `false`, we check if a filesystem exists and throw an exception if so. The `reset` option can be explicitly set to `true` if a filesystem reset is desired. The `reset` option defaults to `false`.

The check for an authed user was needed for checking the filesystem, but is a general improvement because we need a user here.

## Test plan (required)

The change can be tested by calling `bootstrapFileSystem` in an app where a user already has a filesystem. The call should fail, but setting the `reset` option to `true` should succeed and reset the filesystem.

## Closing issues

Closes #396 

## After Merge
* [ ] Does this change invalidate any docs or tutorials? No docs for this one yet
* [ ] Does this change require a release to be made? Yes, I'll release an alpha with it

### DIFF
--- a/src/filesystem.ts
+++ b/src/filesystem.ts
@@ -124,7 +124,7 @@ export const bootstrapFileSystem = async (
       const logCid = await cidLog.newest() // data root in browser
 
       if (dataCid !== null || logCid !== undefined) {
-        throw new Error("Bootstrap operation will reset an existing filesystem. Please set reset to true if this behavior is desired.")
+        throw new Error("Bootstrap operation will \"reset\" an existing filesystem. Please set reset to true if this behavior is desired.")
       }
     }
 
--- a/src/filesystem.ts
+++ b/src/filesystem.ts
@@ -3,6 +3,7 @@ import { CID } from "multiformats/cid"
 import FileSystem from "./fs/index.js"
 
 import * as cidLog from "./common/cid-log.js"
+import * as common from "./common/index.js"
 import * as crypto from "./crypto/index.js"
 import * as debug from "./common/debug.js"
 import * as dataRoot from "./data-root.js"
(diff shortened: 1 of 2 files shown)


src/filesystem.ts
```diff
@@ -105,11 +106,32 @@ export async function loadFileSystem(
...
+  const authedUsername = await common.authenticatedUsername()
+
+  // Check for authed user and existing filesystem
+  if (authedUsername) {
+    if (!reset) {
+      const dataCid = navigator.onLine ? await dataRoot.lookup(authedUsername) : null // data root on server or DNS
+      const logCid = await cidLog.newest() // data root in browser
+
+      if (dataCid !== null || logCid !== undefined) {
+        throw new Error("Bootstrap operation will reset an existing filesystem. Please set reset to true if this behavior is desired.")
```
icidasset, line 127: Maybe put 'reset' in quotation marks or something like that.

icidasset (approved):

src/filesystem.ts
bgins: Sounds good, will do. 👌

src/filesystem.ts
```diff
@@ -105,11 +106,32 @@ export async function loadFileSystem(
...
-export const bootstrapFileSystem = async (permissions: Permissions): Promise<FileSystem> => {
+export const bootstrapFileSystem = async (
+  permissions: Permissions,
+  options: { reset?: boolean } = { reset: false }
+): Promise<FileSystem> => {
+  const { reset } = options
+  const authedUsername = await common.authenticatedUsername()
+
+  // Check for authed user and existing filesystem
+  if (authedUsername) {
```
jessmartin, line 121: Personally, I prefer this kind of check to be structured as a [guard clause](https://refactoring.com/catalog/replaceNestedConditionalWithGuardClauses.html).

PR #406, opened 2022-08-11
Remove internal auth module

### LINKED ISSUE #402: Improve auth dependency injection implementation
# Summary

## Problem

The internal implementation in `src/auth/internal.ts` uses the lobby auth implementations, but we use implementations from `src/auth/implementation.ts` internally.

### Impact

It's confusing. Internal currently means implementations we use for the auth lobby, but my tendency is to think it means auth functions we use internally in webnative.

## Solution

One solution would be to remove `src/auth/internal.ts` and call the lobby implementation directly where needed. Needs a bit more investigation to be sure.

bgins:
## Summary

* [x] Remove the internal auth module
* [x] Call auth init from the DI implementation
* [x] Light bit of dead code elimination

The main goal of this PR is to remove `src/auth/internal.ts` which has been a source of confusion (what does internal mean?) and was barely used. We were only calling the `init` function from it, and we can call that directly from the DI implementation.

Also cleaned up a few unused imports along the way.

## Test plan (required)

Test it with any app that uses the auth lobby and initializes with permission.

## Closing issues

Closes #402 

## After Merge
* [ ] Does this change invalidate any docs or tutorials? No
* [ ] Does this change require a release to be made? No, it can come out with 0.33


matheus23 (approved):
Yay to deleting lines! :P

(especially when functionality stays the same :eyes:)

PR #407, opened 2022-08-17
Add dependency-injected DNS implementation

### LINKED ISSUE #403: Add dependency injected DNS implementation
# Summary

## Problem

It's difficult to test the parts of our code that use DNS lookups.

### Impact

Insufficient test coverage.

## Solution

Add a DI implementation for the DNS lookups in `src/dns/index.ts`. In tests, we can inject implementations that fit the context of the test and avoid network calls.

# Detail

**Describe alternatives you've considered**

Mocking. We've attempted to set up Sinon and a some other mocking libraries, but we hit a bundler wall. Dependency injection is a great alternative because we control the magic, and it can open doors for advanced use cases beyond testing.

bgins:
## Summary

* [x] Dependency-injected DNS lookups
* [x] Tests for did getters
* [x] Moves the in-memory storage implementation alongside the other test helpers

Adding a DI implementation for DNS lookups gives us greater flexibility when testing our code base by allowing us provide temporary implementations to fake DNS lookups. As an example, this PR adds a test suite for the DID getters, which use `dns.lookupTxtRecord` to get a user's DID.

## Test plan (required)

The existing lookups can be tested with anything that looks up DIDs (e.g. account linking) and anything that looks up a DNS link (e.g. looking up a data root).

The new tests can be individually run with `npx mocha -g "root"` and `npx mocha -g "ownRoot"`

## Closing issues

Closes #403 

## After Merge
* [ ] Does this change invalidate any docs or tutorials? No
* [ ] Does this change require a release to be made? Not on its own, can come in with the next release


src/dns/implementation.ts
```diff
@@ -0,0 +1,8 @@
+import { DNS_IMPLEMENTATION } from "./implementation/dns.js"
```
icidasset, line 1: I think we should call this implementation "dns-over-http" or just plainly "http".

icidasset (approved):
One note, looks great otherwise 👍

src/dns/implementation.ts
bgins: Good call! I changed it over to "http".

PR #415, opened 2022-08-26
Improve peer list storage

### LINKED ISSUE #416: Vite warning in webantive app template related to webnative
I see this warning in the console when I run the webnative-app-template:

```
1:46:18 p.m. [vite] warning:
/Users/jeff/code/fission/clones/webnative-app-template/node_modules/.vite/deps/webnative.js
36358|    return import(
36359|      /* webpackIgnore: true */
36360|      cdn_url
   |      ^
36361|    ).then((_) => self.IpfsCore);
36362|  };
The above dynamic import cannot be analyzed by vite.
See https://github.com/rollup/plugins/tree/master/packages/dynamic-import-vars#limitations for supported dynamic import formats. If this is intended to be left as-is, you can use the /* @vite-ignore */ comment inside the import() call to suppress this warning.
```

We should at minimum consider mitigating this warning so developers don't see it. I prioritize this as most important so that developers don't see this warning when they run `npm run dev`

icidasset:
This fixes an issue with certain endpoint configurations and empty peer lists.

Sometimes it was using `/api/vx/ipfs/peers` instead of `/ipfs/peers`, and it didn't work with empty peer lists (or peer lists that didn't have usable peers in it).

Also fixes #416

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 - Fixes the depedency-injected `lookupDnsLink` function
 - Fixes issue with IPFS peer-list storage
+- Removes Vite warning caused by a dynamic import
 
 
 ### v0.34.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,8 @@
 
 #### Bug fixes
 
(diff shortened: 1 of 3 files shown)


bgins (approved):
Looks good! 🎉

PR #421, opened 2022-09-13
Add username schemes for global usernames

### LINKED ISSUE #385: Add identify function for cases where username and/or email not available (AOL)

bgins:
## Summary

* [x] Adds a dependency-injected auth `transformUsername` function
* [x] Adds a `toGlobalUsername` function
* [ ] Updates Webnative to use global username where needed
* [ ] Updates app-to-app linking to expect Base64 URL encoded usernames in query params

The primary goal of this PR is to provide a mechanism for apps to define user namespaces. In our current system, all usernames exist in a global namespace. We will continue to use a global namespace, but we will let developers define username schemes where they can, for example, hash the combination of username and domain to create a scoped pool of usernames.

To achieve this goal, this PR introduces a distinction between usernames and global usernames:
-  **Username.** A human-friendly name that users interact with.
- **Global Username.** A globally unique username that is registered with the Fission server, used to look up DIDs, and generally used internally by Webnative.

A global username is created from a username passing it to `transformUsername` to apply any desired username scheme, then to `toGlobalUsername` to hash the transformed username if desired.

Note that a username and global username may be the same if a the username scheme is the identity function and hashing is not requested. Existing usernames in Webnative will be treated in this way to maintain backward compatibility.

### Tasks

The bulk of the work in this PR will be determining where to use usernames and global usernames. Listing these out seems helpful:

- [ ] Initialization
  - [ ] `username` returned from entrypoints on `state`
  - [ ] `globalUsername` when validating permissions
  - [ ] `globalUsername` when loading the filesystem
  - [ ] `globalUsername` when looking up data root
- [ ] Account registration
  - [ ] `globalUsername` checking if a username if valid and available
  - [x] `globalUsername` when registering with the server
  - [x] `username` stored in browser storage
- [ ] DID lookups
  - [ ] `globalUsername` when looking up DID
- [ ] Data root lookups
  - [ ] `globalUsername` when looking up data root
- [ ] Looking up filesystem UCANs
  - [ ] `globalUsername` when looking up filesystem UCANs (will be used in filesystem prefix)
- [ ]  User domain
  - [ ] `globalUsername` in username (for example `<global-username>.fission.name`)
- [ ] Account linking
  - [ ] `globalUsername` when creating channel
  - [ ] `globalUsername` when checking producer capability
  - [ ] `username` entered by user on consumer side
- [ ] App linking
  - [ ] **Base64 URL encoded `username`** in query param sent when linking (N.B. username might be unsafe, like an email.)
  - [ ] `username` in browser storage on linking the app
  - [ ] `globalUsername` when validating permissions
  - [ ] `globalUsername` when loading the filesystem
- [ ] Shared private files
  - All public methods accept `username` but use `globalUsername` internally
  - _More details to follow..._

## Closing issues

Closes #385

## After Merge

### DIFF
--- a/package.json
+++ b/package.json
@@ -112,6 +112,7 @@
     "localforage": "^1.10.0",
     "multiformats": "^9.5.4",
     "one-webcrypto": "^1.0.3",
+    "punycode": "^2.1.1",
     "throttle-debounce": "^3.0.1",
     "tweetnacl": "^1.0.3",
     "uint8arrays": "^3.0.0",
--- a/package.json
+++ b/package.json
@@ -112,7 +112,6 @@
     "localforage": "^1.10.0",
     "multiformats": "^9.5.4",
     "one-webcrypto": "^1.0.3",
-    "punycode": "^2.1.1",
     "throttle-debounce": "^3.0.1",
     "tweetnacl": "^1.0.3",
     "uint8arrays": "^3.0.0",
--- a/src/auth/channel.ts
+++ b/src/auth/channel.ts
@@ -1,5 +1,4 @@
 import * as did from "../did/index.js"
-import { impl as auth } from "./implementation.js"
 import { setup } from "../setup/internal.js"
 import { LinkingError } from "./linking.js"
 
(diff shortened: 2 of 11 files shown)


bgins:
Closing this for now. Our approach to usernames has changed considerably as has the code base. We can revisit the ideas documented here in the future if they seem useful.

Issue #283, opened 2021-08-31
identifiers should be namespaced


icidasset:
Currently identifiers, which are used for read keys and name filters, aren't namespaced. This makes it impossible to have multiple filesystems loaded at the same time, for example, when you have a temporary filesystem (anonymous/unauthorised user) and a filesystem from a Fission user.

matheus23:
Another way to do achieve this would be to have two keystores under different "storeName"s.

```js
import * as wn from "webnative"
import RSAKeyStore from "keystore-idb/rsa"

wn.keystore.set(new RSAKeyStore({ storeName: "2ndStore" })
```

icidasset:
Oh nice, hadn't thought of that, I'll have to test that out.

Issue #274, opened 2021-08-06
Change redirectToLobby to accept a single object parameter 


bgins:
# Summary

## Problem

`redirectToLobby` requires default arguments to fill positional arguments in some cases.

### Impact

It may be inconvenient or feel like boilerplate for users to fill in defaults for these arguments.

## Solution

Update `redirectToLobby` to accept an object with `permissions`, `redirectTo`, and `extraLobbyParams`.

# Detail

**Is your feature request related to a problem? Please describe.**

The problem will only surface when the `extraLobbyParams` in #273 are added. When these changes are added, users will always need to pass a value for `redirectTo` in order to use the `extraLobbyParams`.

**Describe the solution you'd like**

It would be better if we passed a single object to `redirectToLobby`, for example:

```
redirectToLobby({ permissions: state.permissions, extraLobbyParams: { theme: <CID> } }) 
```

would allow users to skip `redirectTo`.

**Describe alternatives you've considered**

It might be possible to use arrays or some other data structure, but objects seem idiomatic for this.

**Additional context**

This will be a breaking change, but it isn't super impactful. We should wait to bring it in with other breaking changes to minimize disruption to codebases that are currently using webnative.

matheus23:
Throwing in another idea:
```ts
redirectToLobby(state.permissions, { extraLobbyParams: ..., redirectTo: ...})
```

I'd prefer this, because it [seems to be a common pattern](https://github.com/ipfs/js-ipfs/blob/master/docs/core-api/FILES.md). There's one "main", required argument (in this case the permissions), and then an optional second parameter "options", which is then a record where each field is optional.


Small win: Would still be a breaking change, because the second parameter `redirectTo` would become a record, but it's not *that* much of a breaking change, because everyone using `redirectToLobby` with only the permissions argument wouldn't get broken.

bgins:
> Throwing in another idea:
> 
> ```ts
> redirectToLobby(state.permissions, { extraLobbyParams: ..., redirectTo: ...})
> ```
> 
> I'd prefer this, because it [seems to be a common pattern](https://github.com/ipfs/js-ipfs/blob/master/docs/core-api/FILES.md). There's one "main", required argument (in this case the permissions), and then an optional second parameter "options", which is then a record where each field is optional.
> 
> Small win: Would still be a breaking change, because the second parameter `redirectTo` would become a record, but it's not _that_ much of a breaking change, because everyone using `redirectToLobby` with only the permissions argument wouldn't get broken.

[...]

PR #422, opened 2022-09-15
The global state refactor

### LINKED ISSUE #418: Improve read key management and errors
When you try to load an existing filesystem and don't have a `readKey` set in the "webnative storage" (aka. localforage indexeddb in the browser), you'll get the error message `Could not parse a valid private tree using the given key`, which can be confusing. The reason for this is because webnative generates a read key on demand, so you'll actually be using the wrong key in this scenario.

It's probably also confusing that you'll also need to call `wn.lobby.storeRootKey` for the filesystem to actually be able to use the filesystem. It might be better to only store in the web crypto api, and make it extractable when needed?

Linking or delegating transfers this read key from one secure context to another.

icidasset:
## Summary

This refactor gets rid of the global state throughout webnative. Things such as the `setup` object and dependency injection now work entirely different. Instead of accessing a global object, we now create `Program` instances which have pointers to various instances of components that we pass to functions. For example, we create a crypto component and that is passed to a function that does crypto things.

See CHANGELOG.md for more info.

## Tasks

- [x] Remove global state
- [x] Component system
- [x] Refactor crypto
- [x] Refactor auth component to make it compatible with walletauth
- [x] Extract lobby functionality into a component (confidences)
- [x] Extract IPFS functionality into a component (depot)
- [x] Abstract functionality of cid log and ucan storage (now called repositories)
- [x] Create new component called `reference` (dns + did lookup/update + data lookup/update + repos)
- [x] Create new component called `manners` (responsible for config dependent behaviours, eg. debug loggging)
- [x] Merge entrypoints and add customisable component behaviour for `program` entrypoint
- [x] Explain the different auth implementations, the role of the lobby, and the component system (see code docs in `index.ts`)
- [x] Implement sessions
- [x] Make webnative independent (ie. not tied to Fission infra)
- [x] No longer store root key in localstorage, only keystore.
- [x] Abstract root key functionality
- [x] Namespacing all the things! Allow for multiple apps on the same domain (better experience on localstorage)
- [x] Add ability to load multiple filesystems at the same time (ie. avoid identifier clash)
- [x] Remove global references from the filesystem and make instances with components
- [x] Refactor "apps platform" code (publishing apps via webnative, etc)
- [x] Finish refactor of filesystem creating/loading (partially done, removed "bootstrap filesystem" function) (see #424 for more info)
- [x] Refactor UCAN creation & DID validation (depends on crypto component, should we remove automatic setting of `iss` and have the dev always set it?)
- [x] Remove references to IPFS specific code in the filesystem and use the `depot` component instead. Use `Uint8Array`s as much as possible.
- [x] Backwards compatibility. We'll need to copy over data/keys from the correct locations. See https://talk.fission.codes/t/webnative-state-management-initialization-refactor/3363#backwards-compatibility-considerations-3 for specifics.
- [x] Refactor linking code (use components, remove keystore-idb, etc)
- [x] https://github.com/fission-codes/webnative/issues/131
- [x] Easy way to use Fission staging?
- [x] Fix tests
- [x] Test compatibility with older webnative apps
- [x] Write new README
- [x] Ensure compatibility with walletauth

## Code quality checks

- [x] `keystore-idb` code is only used in the `Crypto` implementation code and nowhere else. `SymmAlg` should only be imported from `components/crypto/implementation`
- [x] IPFS is not used/mentioned anywhere except in the `Depot` implementation, everything else should be Depot + DAGs /IPLD + CIDs

## Manual testing

- [x] Loading old filesystem
- [x] Backwards compatibility (use old webnative on same domain, upgrade to newer)
- [x] Linking
- [ ] Sharing private data
- [ ] Fission Drive

## Will need updating

Test with alpha version before release.

- [ ] Fission Dashboard
- [ ] Fission Drive
- [ ] Webnative Elm
- [ ] Webnative Walletauth

## Closes issues

Closes #418, #283, #274


CHANGELOG.md
```diff
@@ -1,5 +1,28 @@
 # Changelog
 
+
+### v0.35.0
```
jessmartin, line 3: I'm sure we've discussed it, but should this be `1.0`?

In favor of `1.0`:
- This is a _complete rewrite._ Feels weird to bump a minor version when it's a ground-up rewrite!
- According to semver, this is a breaking change to APIs and therefore should be a major version bump.
- If someone has `'>=0.34'` in there `package.json`, they'll get updated to `0.35` (I think that's how it works if you haven't specified a patch version) and stuff will break.

In favor of `0.35`:
- We wouldn't want to indicate we think things are production-ready by slapping a 1.0 on this.
- Maybe we want some sort of marketing around a 1.0 release and want to save it?
- Perhaps we're saving 1.0 for the Rust WASM rewrite?
- Prior to 1.0, it's expected that many changes will be breaking changes! So no worries about violating semver.

Curious what others think.

CHANGELOG.md
```diff
@@ -1,5 +1,28 @@
...
+
+Full rewrite of webnative.
```
jessmartin, line 5: This deserves a 🎉 . 😄

CHANGELOG.md
```diff
@@ -1,5 +1,28 @@
...
+- Introduction of a `confidences` components: Responsible for accepting UCANs and filesystem secrets from external sources (eg. Fission Lobby)
+
+- Removal of the `webnative.initialise`, `app` and `permissionedApp` functions. These have been replaced by the the `program` function which serves as the single entrypoint. You can customise all components and configuration via this function. It'll give you a `Program` with possibly a `Session` (and many other things) instead of the `State` we had before. Basically, when you get a session you're "logged in", otherwise you're not.
+
+- Webnative can now have many apps run on the same domain without any conflicts. This should help a lot when developing apps on the same localhost port. All storage and filesystems are namespaced by default.
+
+- Allows for multiple filesystems to be loaded at the same time (conflicts with identifiers have been resolved) and allows for a temporary filesystem (progressive login).
+
+- Removal of various confusing filesystem parameters, all data and read keys are `Uint8Array`s now.
+- When loading a filesystem the data root is tried multiple times (to get around the DNS issue)
```
jessmartin, line 21: I did a "full rewrite" of this section, reorganizing it logically. I'm not sure every change I made makes sense. And there's probably a bit more that could be done, but this feels like it reads more smoothly.

Here's what I'm thinking:

---

Full rewrite of webnative. 🎉 

**Components**
We've moved to a component system to make webnative more customizable and independent of Fission infrastructure. The system removes all global state, avoiding bundler issues.  We previously had dependency injection; this is basically more of the same without global state.

The documentation should give you some information on how the various components fit together. If you want to write implementations for a component, see the directories in `src/components` (or the lib folder). All existing components have been expanded and refactored, and we've introduced the following:
- Introduction of a `depot` component: All IPFS functionality has been moved into here. Can now be swapped out with something else. Note that webnative still uses CIDs & IPLD.
- Introduction of a `reference` component: Data root lookups & updating, DID root lookups, DNS and various repositories (cid log & UCANs) have been moved in here.
- Introduction of a `manners` component: Debug functionality has moved to here.
- Introduction of a `confidences` components: Responsible for accepting UCANs and filesystem secrets from external sources (eg. Fission Lobby)

**Improvements**
- The `program` function now serves as the single entrypoint. You can customise all components and configuration via this function. It'll give you a `Program` with possibly a `Session` (and many other things) instead of the `State` we had before. Basically, when you get a session you're "logged in", otherwise you're not.
- Webnative can now have many apps run on the same domain without any conflicts. This should help a lot when developing apps on the same localhost port. All storage and filesystems are namespaced by default.
- When loading a filesystem the data root is tried multiple times (to get around the DNS issue)
- Multiple filesystems can be loaded at the same time (conflicts with identifiers have been resolved), which allows for a temporary filesystem (progressive login).

**Removed**
- Removal of various confusing filesystem parameters, all data and read keys are `Uint8Array`s now.
- Removal of the `webnative.initialise`, `app` and `permissionedApp` functions. These have been replaced by the `program` function mentioned above.

jessmartin: Checked this in already.

CHANGELOG.md
jessmartin: Talked to @therealjeffg about it today and he's in favor of `0.35` as we don't want to indicate production-readiness with the `1.0` label.

Marking this as resolved!

therealjeffg: In reaction to this comment:

> If someone has '>=0.34' in there package.json, they'll get updated to 0.35 (I think that's how it works if you haven't specified a patch version) and stuff will break.

I agree with this sentiment - I think we'll get away with it this time because we don't have a lot of downstream adoption currently but in the future we need to be much better wrt helping developers consuming the library stay up to date. This isn't at the standard we should strive for, for exactly the reason Jess points out.

icidasset: Just wanted to this here: Normally people will install webnative using `npm install`, and that'll set the version to `^0.34.2`. That means people **will not** automatically get `0.35.0`, because the version starts with a zero, according to https://docs.npmjs.com/cli/v6/using-npm/semver#caret-ranges-123-025-004

src/components/depot/implementation.ts
```diff
@@ -0,0 +1,30 @@
...
+
+export type Implementation = {
+  // Get the data behind a CID
+  getBlock: (cid: CID) => Promise<Uint8Array>
+  getUnixFile: (cid: CID) => Promise<Uint8Array>
+  getUnixDirectory: (cid: CID) => Promise<DirectoryItem[]>
+
+  // Keep data around
+  putBlock: (data: Uint8Array, codec: BlockCodec<number, any>) => Promise<CID>
+  putChunked: (data: Uint8Array) => Promise<PutResult>
```
icidasset, line 13: I should probably explain here that chunked means a collection of various blocks grouped together by a DAG-
PB node.

src/components/depot/implementation.ts
```diff
@@ -0,0 +1,30 @@
+import { BlockCodec } from "multiformats/codecs/interface"
+import { CID } from "multiformats/cid"
+
+
+export type Implementation = {
```
icidasset, line 5: Preparation for CAR mirror/pool, allows us to switch out js-ipfs.

src/fs/protocol/basic.ts
```diff
@@ -1,89 +1,163 @@
...
+
+  // NOTE: Somehow the try/catch is needed by the integration test with the CAR file.
+  //       Maybe a mistake in the CAR file or how the file system code worked before?
+  let withAlgorithm
+
+  try {
+    withAlgorithm = DagCBOR.decode(buf)
+  } catch {
+    // Not CBOR?
+    return buf
```
icidasset, line 37: Not sure what this is about see note.

src/fs/protocol/private/mmpt.ts
```diff
@@ -53,23 +59,23 @@ export default class MMPT implements Puttable {
...
 
     // if already in tree, then skip
     if (name === nextNameOrSib) {
-      if (setup.debug && this.links[name]?.cid !== value) {
-        console.warn(`Adding \`${name}\` to the MMPT again with a different value. This should not happen. The original value will still be used and can loading issues on other ipfs repos. Current CID is \`${this.links[name]?.cid}\`, new CID is \`${value}\`.`)
-      } else {
-        // skip
-      }
+      // if (this.links[ name ]?.cid !== value) {
+      //   this.manners.warn(`Adding \`${name}\` to the MMPT again with a different value. This should not happen. The original value will still be used and can loading issues. Current CID is \`${this.links[ name ]?.cid}\`, new CID is \`${value}\`.`)
```
icidasset, line 70: Disabled this as it's of no use anymore.

src/fs/v1/PrivateTree.ts
```diff
@@ -137,13 +206,17 @@ export default class PrivateTree extends BaseTree {
   async putDetailed(): Promise<PrivateAddResult> {
     // copy the object, so we're putting the current version & don't include any revisions
     const nodeCopy = Object.assign({}, this.header)
-    return protocol.priv.addNode(this.mmpt, nodeCopy, this.key)
+
+    // ensure all CIDs in skeleton are in string form, not sure where these CID objects are coming from
+    nodeCopy.skeleton = ensureSkeletonStringCIDs(nodeCopy.skeleton)
```
icidasset, line 211: Not sure about this either.

src/repository.ts
```diff
@@ -0,0 +1,107 @@
...
+
+  getByIndex(idx: number): T | null {
+    return this.memoryCache[ idx ]
+  }
+
+  async getAll(): Promise<T[]> {
+    const storage = await this.storage.getItem(this.storageName)
+    const storedItems = TypeChecks.isString(storage)
+      // TODO: ? - Need partial JSON decoding for this
+      ? storage.split("|||").map(this.fromJSON)
```
icidasset, line 68: I'd like something better here, but don't know how to approach this.
I thought I could `JSON.stringify` the whole collection, but I need to call `fromJSON` on every item 🤔

src/ucan/token.ts
```diff
@@ -39,21 +42,23 @@ export async function build({
...
-  const decodedProof = proof && decode(proof)
-  const ksAlg = await crypto.keystore.getAlg()
+  const decodedProof = proof
+    ? (typeof proof === "string" ? decode(proof) : proof)
+    : null
 
   // Header
   const header = {
-    alg: jwtAlgorithm(ksAlg) || "UnknownAlgorithm",
+    alg: await dependents.crypto.keystore.getUcanAlg(),
```
icidasset, line 61: Allow for other types of UCANs such as the ones created with walletauth.

src/ucan/token.ts
```diff
@@ -74,14 +79,14 @@ export async function build({
     aud: audience,
     exp: exp,
     fct: facts,
-    iss: issuer || await did.ucan(),
```
icidasset, line 77: No more default issuer, required param.

tests/fs/wasm/public.test.ts
```diff
@@ -1,69 +1,71 @@
...
-import * as ipfsConfig from "../../../src/ipfs/config.js"
+
 import { HardLinks } from "../../../src/fs/types.js"
 import { CID } from "multiformats"
+import { components } from "../../helpers/components.js"
+
 
 
+// TODO: Fix GC issues
+describe.skip("the wasm public root", () => {
```
icidasset, line 13: Temporarily disabled this because of a GC issue.
Need to create an issue for @matheus23

tests/helpers/components.ts
```diff
@@ -0,0 +1,303 @@
...
+    loadImmediately: false
+  },
+}
+
+
+
+// CRYPTO
+
+
+export async function createCryptoComponent(): Promise<Crypto.Implementation> {
```
icidasset, line 57: Could be moved to implementations as "in-memory keystore + webcrypto"

tests/helpers/components.ts
```diff
@@ -0,0 +1,303 @@
...
+
+
+// const [ ipfs, repo ] = await createInMemoryIPFS()
+// const depot: Depot.Implementation = await IpfsDepot.implementation(ipfs, repo)
+
+
+export const inMemoryDepot: Record<string, Uint8Array> = {}
+
+
+const depot: Depot.Implementation = {
```
icidasset, line 138: This avoids us having to use js-ipfs in the tests.
Bit hacky I guess because I shove large chunked files inside a single block, but it works.

tests/helpers/components.ts
```diff
@@ -0,0 +1,303 @@
...
+
+
+
+// STORAGE
+
+
+let inMemoryStorage: Record<string, any> = {}
+
+
+const storage: Storage.Implementation = {
```
icidasset, line 197: Could be moved to implementations, in-memory storage.

src/index.ts
```diff
@@ -1,162 +1,625 @@
...
+@@@@@@@    %@@@@@@@@@@@@@@@     @@@@%    @@@@@@@
+@@@@@@@     %@@%     @@@@@@     %@@%     @@@@@@@
+@@@@@@@              @@@@@@              @@@@@@@
+@@@@@@@%            %@@@@@@%            %@@@@@@@
+@@@@@@@@@%        %@@@@@@@@@@%        %@@@@@@@@@
+%@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@%
+  @@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@
+    %@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@%
+
+ */
```
icidasset, line 24: Most important file, everything comes together here. The documentation should be self-explanatory, note if anything is not clear or missing.

src/index.ts
```diff
@@ -1,162 +1,625 @@
...
+
+
+export type ShortHands = {
+  loadFileSystem: (username: string) => Promise<FileSystem>
+  loadRootFileSystem: (username: string) => Promise<FileSystem>
+}
+
+
+export type AuthenticationStrategies = Record<
+  string,
```
icidasset, line 122: Should note here that this will be the auth types. Every activated auth strategy type will have a property in this record.

src/index.ts
```diff
@@ -1,162 +1,625 @@
...
-  } else {
-    setImplementations(BASE_IMPLEMENTATION)
-  }
 
-  // Check if browser is supported
-  if (globalThis.isSecureContext === false) throw InitialisationError.InsecureContext
-  if (await isSupported() === false) throw InitialisationError.UnsupportedBrowser
 
-  const authedUsername = await common.authenticatedUsername()
+// PREDEFINED COMPONENT COMBINATIONS
```
icidasset, line 166: See this section for a detailed explanation on the components and how they work together. A short explanation is located in the changelog.

src/index.ts
```diff
@@ -1,164 +1,634 @@
...
+  loadRootFileSystem: (username: string) => Promise<FileSystem>
+}
+
+
+export type AuthenticationStrategies = Record<
+  string,
+  {
+    implementation: Auth.Implementation<Components>
+
+    accountConsumer: (username: string) => Promise<AccountLinkingConsumer>
```
icidasset, line 126: I wanted to separate the functions that a user/dev directly uses from the implementation. Mixing the implementation functions in here would be confusing, for example, `linkDevice` does not fully link a device, you have to create an account linking consumer & producer.

src/index.ts
```diff
@@ -1,164 +1,633 @@
...
+ * - `session`, a `Session` object if a session was created before.
+ * - `auth`, a means to control the various auth strategies you configured. Use this to create sessions. Read more about auth components in the toplevel `auth` object documention.
+ * - `confidences`, a means to control confidences. Use this to collect & request confidences, and to create a session based on them. Read more about confidences in the toplevel `confidences` object documentation.
+ * - `components`, your full set of `Components`.
+ *
+ * This object also has a few other functions, eg. to load a filesystem.
+ * These are called "shorthands" because they're the same functions available
+ * through other places in webnative, but you don't have to pass in the components.
  *
+ * See `assemble` for more information.
```
icidasset, line 158: Can we somehow make it so that typedoc inserts a link here to the documentation of this `assemble` function?

README.md
```diff
@@ -1,130 +1,181 @@
...
-      // We need a secure context to do cryptography
-      // Usually this means we need HTTPS or localhost
+})
+```
 
-    case wn.InitialisationError.UnsupportedBrowser:
-      // Browser not supported.
-      // Example: Firefox private mode can't use indexedDB.
-  }
+That gives us a `Program` object, with this we can create a new user session or reuse the existing session. There's two distinctive ways to create a user session, either by using an authentication strategy or by using something we call "confidences". More on the latter in a bit, let's start with the default authentication strategy.
```
bgins, line 53:
```suggestion
`wn.program` returns a `Program` object, which can create a new user session or reuse an existing session. There are two ways to create a user session, either by using an authentication strategy or by requesting access from another app through the "confidences" system. Let's start with the default authentication strategy.
```

Possible rewording. Also added a suggestion for how we might hint at confidences and how they are different from selecting an authentication strategy.

README.md
```diff
@@ -1,130 +1,181 @@
...
+```ts
+let session
+
+// Do we have an existing session?
+if (program.session) {
+  session = program.session
+
+// If not, let's authenticate.
+// (a) new user, register a new Fission account
+} else if (registerNewUser) {
```
bgins, line 64: How would a developer decide what to set `registerNewUser` to?

README.md
```diff
@@ -1,130 +1,181 @@
...
+  consumer.on("link", ({ approved, username }) => {
+    if (approved) {
+      console.log(`Successfully authenticated as ${username}`)
+      session = program.auth.session()
+    }
+  })
+}
+```
 
+Alternatively you can use "confidences", this system is used when you want partial access to a file system. At the moment of writing this only supported through the "Fission auth lobby", which is a Webnative app that uses the auth strategy shown above.
```
bgins, line 101:
```suggestion
Alternatively you can use the "confidences" system when you want partial access to a file system. At the moment of writing, confidences are only supported through the "Fission auth lobby", which is a Webnative app that uses the auth strategy shown above.
```

README.md
```diff
@@ -1,130 +1,181 @@
...
 
-switch (state.scenario) {
+This Fission auth lobby flow works as follows:
+1. You get redirected to the Fission lobby from your app.
+2. Here you create an account like in the normal auth strategy flow shown above.
+3. The lobby shows what your app wants to access in your file system.
+4. You approve or deny these permissions and get redirected back to your app.
+5. Your app collects the encrypted information (UCANs & file system secrets).
+6. You can create a user session.
```
bgins, line 109:
```suggestion
6. Your app can create a user session.
```
To be consistent with the other ones.

README.md
```diff
@@ -1,130 +1,181 @@
...
 
+```ts
+const fs = session.fs
+```
 
+__Notes:__
 
-# Webnative File System
+- You can use alternative authentication strategies, such as [webnative-walletauth](https://github.com/fission-codes/webnative-walletauth).
+- You can remove all traces of the user using `await session.destroy()`
```
bgins, line 142: Yeah, nice to have this exposed off the session. 🧨

README.md
```diff
@@ -1,130 +1,181 @@
...
+- You can load the file system separately, in case you're using web worker. This is done using the combination of `configuration.filesystem.loadImmediately = false` and `program.loadFileSystem()`
```
bgins, line 143:
```suggestion
- You can load the file system separately if you're using a web worker. This is done using the combination of `configuration.filesystem.loadImmediately = false` and `program.loadFileSystem()`
```

bgins (commented):
Added a few comments to the README. Next up, looking at the code.

README.md
```diff
@@ -1,130 +1,181 @@
...
-    // Ask the user permission to additional filesystem paths
-    fs: {
-      private: [ wn.path.directory("Audio", "Music") ],
-      public: [ wn.path.directory("Audio", "Mixtapes") ]
-    }
+const program = await wn.program({
+  // Can also be a string, used as an identifier for caches.
+  // If you're developing multiple apps on the same localhost port,
+  // make sure these differ.
+  id: { creator: "Nullsoft", name: "Winamp" }
```
avivash, line 39: Mentioned in the WAT PR the idea of using `namespace` in place of `tag` or `id`. what do you think?
```suggestion
  namespace: { creator: "Nullsoft", name: "Winamp" }
```

README.md
```diff
@@ -1,130 +1,181 @@
...
+    if (userInput === challenge.pin) challenge.confirmPin()
+    else challenge.rejectPin()
+  })
+
+  producer.on("link", ({ approved }) => {
+    if (approved) console.log("Link device successfully")
+  })
+
+  // On device without session:
+  //     Somehow you'll need to get a hold of the username.
```
avivash, line 84: small typo. `ahold` is one word in this context :) 
```suggestion
  //     Somehow you'll need to get ahold of the username.
```

README.md
```diff
@@ -1,130 +1,181 @@
...
-}
+// (b) When you get redirected back and your program is ready,
+// you will have access to your user session.
+session = program.session
 ```
 
-`redirectToLobby` will redirect you to [auth.fission.codes](https://auth.fission.codes) our authentication lobby, where you'll be able to make a Fission an account and link with another account that's on another device or browser. The function takes a second, optional, parameter, the url that the lobby should redirect back to (the default is `location.href`).
-
-`initialise` will return a rejected Promise if the browser, or context, is not supported.
+Once you have your `Session`, you have access to the file system 🎉
```
avivash, line 133: Maybe use `your` in place of `the`, as it is your file system?
```suggestion
Once you have your `Session`, you have access to your file system 🎉
```

src/common/version.ts
```diff
@@ -1,2 +1,2 @@
-export const VERSION = "0.34.2"
+export const VERSION = "0.35.0"
```
avivash, line 1: more of an idea than a suggestion, but maybe it would make sense to pull the version directly from the `package.json` here to reduce the potential for human error when bumping the version 🤔

src/common/cid.ts
```diff
@@ -10,17 +18,26 @@ export function decodeCID(val: CID | object | string): CID {
...
 
-  if (typeof val === "string") return CID.parse(val as string)
+  if (typeof val === "string") return CID.parse(val)
   if (typeof val === "object" && "version" in val && "code" in val && "multihash" in val) {
     return CID.create(val.version, val.code, val.multihash)
   }
 
-  throw new Error(`Could not decode CID: ${val}`)
+  // Result of CID.toJSON()
+  if (typeof val === "object" && hasProp(val, "version") && val.version === 1 && hasProp(val, "code") && isNum(val.code) && hasProp(val, "hash") && isObject(val.hash) && Object.values(val.hash).every(isNum)) {
```
avivash, line 27: this condition is a little verbose. I wonder if it would make sense to save it to a variable with a more descriptive name and check against that? Maybe with a comment too? something like:
```suggestion
  /**
   * A valid encoded CID object must meet the following criteria:
   * - has a `version` equal to 1
   * - has a `code` that is a number
   * - has a `hash` object containing only numerical values
  */
  const isCIDValid = typeof val === "object" && hasProp(val, "version") && val.version === 1 && hasProp(val, "code") && isNum(val.code) && hasProp(val, "hash") && isObject(val.hash) && Object.values(val.hash).every(isNum)
  
  if (isCIDValid) {
```

src/index.ts
```diff
@@ -172,32 +624,133 @@ export async function isSupported(): Promise<boolean> {
...
+    }
+  })
+}
+
+
+
+// 🛠
+
+
+export function extractConfig(opts: Partial<Components> & Configuration): Configuration {
```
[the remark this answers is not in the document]
icidasset: Ah yes, that should fail because `tag` is a required parameter.
I just omitted it in the readme.

I wasn't sure if it should infer it from permissions.
I think I prefer explicitness here.
I'll add the `tag` to the example.

src/common/cid.ts
icidasset: Your comment suggestion is not entirely correct. A valid CID can take many shapes, hence the other parts of that function. This is a particular format coming from `multiformats/cid -> instance.toJSON`. I'll add a better explanation 👍

src/common/version.ts
icidasset: We do that already yeah. See `scripts/gen-version.js`

README.md
icidasset: Good to know, thanks 😄

README.md
icidasset: Oh great idea! That's the word I was looking for 👌

README.md
```diff
@@ -1,130 +1,181 @@
...
+import * as wn from "webnative"
 
 // Browser/UMD build
 const wn = globalThis.webnative
 ```
 
-# Authentication
+## Creating a Program
+
+A Webnative program is an assembly that makes up a part of, or a whole distributed web application. It consists out of several components that can be customised. _We'll stick with the default components for now, which means we'll be using the Web Crypto API._
```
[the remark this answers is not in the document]
icidasset: Well, I would argue that "of components" is implied, one of definitions of assembly is:
> a unit consisting of components that have been fitted together

I agree with the other changes 👍

README.md
icidasset: It's decided by the user wether or not they have an account, eg. this screen in WAT:

Screenshot 2022-11-18 at 11 23 59

The wording could be better here yeah.
Maybe `userChoseToRegister`?

matheus23 (commented):
Love that we don't have global state anymore!
This makes working on webnative easier in the future, enables nicer tests, we can reason about code changes better etc! :tada: 
I also really liked browsing the docs and specifically all the types. It really helps me form a mental model of what is needed for webnative to function! Things like a list of UCANs, FileSystemSecrets and the username, etc.

---

### Name Bikeshedding

I don't quite like the name 'confidences'. Could be my non-native-enligsh-speakerness, but at first I didn't even think there existed a plural of "confidence". Maybe we can name that 'capabilities'? Or, perhaps something like 'AccountCapabilities'? Or 'AccountSecrets'? (Although the UCANs technically aren't secrets, and there's more/other secrets in the keystore).

### Refactor

I think we may be able to factor the components *slightly* differently.
E.g. `manners` is mostly logging, so it could've been named `logger`, except that it has "wnfsWasmLookup", which I guess could live in `reference` instead? (not sure)

Is there someone using dependency injection for `crypto.aes`, `crypto.hash`, `crypto.misc` or `crypto.rsa`?

I know it was originally Daniel who set up dependency-injection for them, but I *think* we don't have to do that now that `one-webcrypto` is a thing. They could all just become utility functions based on that.

I know there's some untanglement needed with `keystore-idb` though (essentially I'd propose to just inline `keystore-idb`s functions in webnative as a start). Let me know what you think of that in principle. I guess that's perhaps ... another refactor away :sweat_smile:

src/components/manners/README.md
```diff
@@ -0,0 +1 @@
+Behaviour changes based on the configuration given to an entry point.
```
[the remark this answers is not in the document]
icidasset: Right, I started documenting the components this way, but ended up documenting them in `index.ts`. I'll delete it 👍

README.md
```diff
@@ -1,130 +1,182 @@
...
+
 await fs.write(
-  fs.appPath(wn.path.file("Sub Directory", "hello.txt")),
-  "👋"
+  contentPath,
+  new TextEncoder().encode("👋") // Uint8Array
 )
 
-// Announce the changes to the server
+// Announce the changes to your other devices
```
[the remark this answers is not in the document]
icidasset: I like it!

icidasset:
@matheus23 Yeah, confidences is a difficult one. It comes from this definition:

> [count noun] (often confidences) a secret or private matter told to someone under a condition of trust: the girls exchanged confidences about their parents.

Which I think fits the description perfectly, exchanging secrets. The component is not responsible for storing the secrets, it's responsible for getting them to the app. I considered "capabilities" too, but it found it confusing because we have UCAN capabilities and it's more than that. Also does capabilities clash with permissions? Confidences felt more distinct and appropriate. That said, I am open to changing it to capabilities if other people find it a better fit too.

Thoughts @bgins @avivash?

---

> I think we may be able to factor the components slightly differently.

Yeah, definitely room for improvement. I left all the crypto in there, because I could see some people preferring WASM for all their crypto. I could see it being better having smaller components, but on the other hand it's easy to inherit from other components. Does it hurt leaving that possibility open, not sure? 🤷‍♂️ 

I'm not happy with the keystore stuff either, and I've been looking for a better approach but haven't found anything yet. I need to see how walletauth exactly fits in here. Same thing with auth, most auth strategies won't need linking, so why is it part of the auth component? Things to think about. I feel like I'm going to tackle that right after this, since I'll be fitting walletauth and hopefully passkeys into this picture.

I kept `wnfsWasmLookup` in `manners` because it determines how webnative works. I wanted the reference to be:

> the use of a source of information in order to ascertain something

So it's all about looking at "sources of truth", or updating them. Whereas "manners" is more how something is done. Does that make sense?

README.md
icidasset: Updated to `namespace`

README.md
bgins: Yeah, that wording would be better! 👌

src/index.ts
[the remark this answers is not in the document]
bgins: Sounds good, I'm with explicit. 💯

bgins:
> @matheus23 Yeah, confidences is a difficult one. It comes from this definition:
> 
> > [count noun] (often confidences) a secret or private matter told to someone under a condition of trust: the girls exchanged confidences about their parents.
> 
> Which I think fits the description perfectly, exchanging secrets. The component is not responsible for storing the secrets, it's responsible for getting them to the app. I considered "capabilities" too, but it found it confusing because we have UCAN capabilities and it's more than that. Also does capabilities clash with permissions? Confidences felt more distinct and appropriate. That said, I am open to changing it to capabilities if other people find it a better fit too.
> 
> Thoughts @bgins @avivash?

At least in American English, that use of confidence has fallen out of common use. It's more something you would see in a 19th century novel.

I can see a connection between permissions in capabilities. You request permission and receive capability when permission is granted. But in the `request` call would that be clear?

```typescript
program.capabilities.request(permissions)
```

It's true that it's a bit inaccurate to call it capabilities, because you are receiving more than just UCANs. On the other hand, one could make the argument that you are incapable of accessing the filesystem without the other secrets.

All in all, I think capabilities is a bit better than confidences, but it doesn't seem like the perfect thing.

src/common/version.ts
avivash: ahh, cool!

src/common/cid.ts
avivash: good to know, thanks! I was just trying to distill the condition down into a comment 😅

avivash:
> @matheus23 Yeah, confidences is a difficult one. It comes from this definition:
> 
> > [count noun] (often confidences) a secret or private matter told to someone under a condition of trust: the girls exchanged confidences about their parents.
> 
> Which I think fits the description perfectly, exchanging secrets. The component is not responsible for storing the secrets, it's responsible for getting them to the app. I considered "capabilities" too, but it found it confusing because we have UCAN capabilities and it's more than that. Also does capabilities clash with permissions? Confidences felt more distinct and appropriate. That said, I am open to changing it to capabilities if other people find it a better fit too.
> 
> Thoughts @bgins @avivash?

I would agree with Brian, I have never really heard `confidence(s)` used in this context before. I like the suggestion of using `capabilities`. As a developer looking at the code for the first time, that would make more sense to me

icidasset:
Capabilities it is 👍

src/index.ts
```diff
@@ -1,164 +1,616 @@
...
-  } else {
-    return appState.scenarioNotAuthed()
+/**
+ * Predefined confidences configurations.
+ *
+ * If you want partial read and/or write access to the filesystem you'll want
+ * a "confidences" component. This component is responsible for requesting
+ * and receiving UCANs, read keys and namefilters from other sources to enable this.
+ *
+ * NOTE: This uses all the default components as the dependencies for the confidences component.
```
bgins, line 229: Hmm was about to make a similar rewrite suggestion as the one for the auth component above, but not so sure. Might be alright as is. 🤔

src/index.ts
```diff
@@ -1,164 +1,616 @@
...
+ *       If you're, for example, using a non-default crypto component, you'll want to
+ *       pass that in here as a parameter as well.
+ *
+ *       Dependents: crypto, depot.
+ */
+export const confidences = {
+  /**
+   * A secure enclave in the form of a webnative app which serves as the root authority.
+   * Your app is redirect to the lobby where the user can create an account or link a device,
```
bgins, line 238:
```suggestion
   * Your app is redirected to the lobby where the user can create an account or link a device,
```

src/index.ts
```diff
@@ -1,164 +1,616 @@
...
+   * and then request permissions to the user for reading or write to specific parts of the filesystem.
```
bgins, line 239:
```suggestion
   * and then request permissions from the user for reading or write to specific parts of the filesystem.
```

src/index.ts
```diff
@@ -1,164 +1,616 @@
...
+  }
+}
 
+/**
+ * Predefined depot configurations.
+ *
+ * The depot component gets data in and out your program.
+ * For example, say I want to load and then update a file system.
+ * The depot will get that file system data for me,
+ * and when updating it, bring the data to where it needs to be.
```
bgins, line 267:
```suggestion
 * and after updating it, send the data to where it needs to be.
```

Not at all sure about these ones. If I recall correctly, we tell the server "come and get the data". So perhaps this is more like

> and after updating it, will send the data or notify an external service to fetch the data.

Service also maybe isn't right? Consider it an initial idea. 😅

src/index.ts
```diff
@@ -1,164 +1,616 @@
...
-  }
+
+/**
+ * Build a Webnative Program based on a given set of `Components`.
+ * These are various customisable components that determine how a Webnative app works.
+ * Use `program` to work with a default, or partial, set of components.
+ *
+ * Additionally this does a few other things:
+ * - Checks if the browser is supported.
+ * - Restores a session if one was made before, and load the user's file system if needed.
```
bgins, line 338:
```suggestion
 * - Restores a session if one was made before and loads the user's file system if needed.
```

src/index.ts
```diff
@@ -1,164 +1,616 @@
...
- * NOTE: Only works on the main/ui thread, as it uses `window.location`.
+ * This will give you a `Program` object which has the following properties:
+ * - `session`, a `Session` object if a session was created before.
+ * - `auth`, a means to control the various auth strategies you configured. Use this to create sessions. Read more about auth components in the toplevel `auth` object documention.
+ * - `confidences`, a means to control confidences. Use this to collect & request confidences, and to create a session based on them. Read more about confidences in the toplevel `confidences` object documentation.
+ * - `components`, your full set of `Components`.
  *
+ * This object also has a few other functions, eg. to load a filesystem.
+ * These are called "shorthands" because they're the same functions available
+ * through other places in webnative, but you don't have to pass in the components.
```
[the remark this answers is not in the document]
bgins: Ah reading below, I may not be understanding the issue here. Is it possible to directly import these helpers from a module? Or do they need to be configured, and we are really trying to avoid developers interacting with the lower-level components directly?

src/common/fission.ts
```diff
@@ -0,0 +1,80 @@
...
+export const PRODUCTION: Endpoints = {
+  apiPath: "/v2/api",
+  lobby: "https://auth.fission.codes",
+  server: "https://runfission.com",
+  userDomain: "fission.name"
+}
+
+
+export const STAGING: Endpoints = {
+  apiPath: "/api/v2",
```
bgins, line 29:
```suggestion
  apiPath: "/v2/api",
```
Should this be `/v2/api`?

src/components/confidences/implementation.ts
```diff
@@ -0,0 +1,15 @@
...
+
+
+export type RequestOptions = {
+  extraParams?: Record<string, string>
+  permissions?: Permissions
+  returnUrl?: string
+}
+
+export type Implementation = {
+  collect: () => Promise<Maybe<Confidences>>,
```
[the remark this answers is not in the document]
icidasset: I'm just doing what eslint tells me to do 😄 Weird that it doesn't pick up these ones.

src/index.ts
```diff
@@ -1,164 +1,616 @@
...
+}
+
+
+export type Program = ShortHands & {
+  auth: AuthenticationStrategy
+  components: Components
+  confidences: {
+    collect: () => Promise<Maybe<string>> // returns username
+    request: () => Promise<void>
+    session: (username: string) => Promise<Maybe<Session>>
```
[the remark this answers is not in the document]
icidasset: `confidences.session()` like `auth.session()` would be used whenever the "authenticated" state has assumed to be changed. So in the case of confidences/capabilities, whenever the confidences/capabilities are collected. Which happens automatically in the `assemble` function when we create our `Program`.

Using the lobby, you get redirect back to your app, a program is created, the confidences are collected and you get your session. The reason it is exposed is, you could write an implementation for confidences/capabilities that doesn't use a redirect, so you would need to call `.session()` manually after the program is already created (just like with auth you need to call it manually in some situations).

src/index.ts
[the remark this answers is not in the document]
icidasset: > Or do they need to be configured, and we are really trying to avoid developers interacting with the lower-level components directly?

Exactly. You can import them from a module, but you will need to pass in the components yourself if you do.

src/index.ts
icidasset: Yeah I'm not entirely happy with this line either.
This gives the assumption that you can't customise it.

Maybe we just drop this line entirely and say:
> If you're using a non-default component, you'll want to pass that in here as a parameter as well.
> Dependents: crypto, manners, reference, storage.

src/index.ts
icidasset: Well, it's not the depot that notifies the "external service". My wording is probably confusing here. It's really the data root we update via the reference component that notifies the server to get the data from our app's depot.

src/common/fission.ts
icidasset: Thanks for spotting that! 👍

src/index.ts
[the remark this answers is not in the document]
bgins: Got it, makes sense to keep it exposed like that. 👌

src/index.ts
bgins: Yeah, that reads better to me! 👍

src/components/capabilities/implementation/fission-lobby.ts
```diff
@@ -0,0 +1,329 @@
...
+    && TypeChecks.isObject(obj.fs)
+    && Object.values(obj.fs).every(a => TypeChecks.hasProp(a, "key") && TypeChecks.hasProp(a, "bareNameFilter"))
+    && Array.isArray(obj.ucans)
+    && obj.ucans.every(a => TypeChecks.isString(a))
+}
+
+async function translateClassifiedInfo(
+  { crypto }: Dependents,
+  classifiedInfo: LobbyClassifiedInfo
+): Promise<{ fileSystemSecrets: Capabilities.FileSystemSecret[], ucans: Ucan.Ucan[] }> {
```
bgins, line 226:
```suggestion
): Promise<{ fileSystemSecrets: Capabilities.FileSystemSecret[]; ucans: Ucan.Ucan[] }> {
```

src/components/capabilities/implementation/fission-lobby.ts
```diff
@@ -0,0 +1,329 @@
...
+
+export async function collect(
+  endpoints: Fission.Endpoints,
+  dependents: Dependents
+): Promise<Maybe<Capabilities.Capabilities>> {
+  const url = new URL(window.location.href)
+  const authorised = url.searchParams.get("authorised")
+  if (!authorised) return null
+
+  const username = url.searchParams.get("username") || ""
```
bgins, line 40:
```suggestion
  const username = url.searchParams.get("username") ?? ""
```

Maybe use nullish coalescing?

src/components/capabilities/implementation/fission-lobby.ts
```diff
@@ -0,0 +1,329 @@
...
+  { crypto }: Dependents,
+  classifiedInfo: LobbyClassifiedInfo
+): Promise<{ fileSystemSecrets: Capabilities.FileSystemSecret[], ucans: Ucan.Ucan[] }> {
+  // Extract session key
+  const rawSessionKey = await crypto.keystore.decrypt(
+    Uint8arrays.fromString(classifiedInfo.sessionKey, "base64pad")
+  )
+
+  // The encrypted session key and read keys can be encoded in both UTF-16 and UTF-8.
+  // This is because keystore-idb is to UTF-16 by default, and that's what webnative used before.
```
bgins, line 233:
```suggestion
  // This is because keystore-idb uses UTF-16 by default, and that's what webnative used before.
```

Possible typo.

bgins (commented):
A few more comments and suggestions. New auth implementation is looking great! ✨

src/components/auth/implementation/base.ts
```diff
@@ -0,0 +1,141 @@
...
+  const u = Ucan.decode(token as string)
+
+  if (await Ucan.isValid(dependents.crypto, u)) {
+    await dependents.storage.setItem(dependents.storage.KEYS.ACCOUNT_UCAN, token)
+    await SessionMod.provide(dependents.storage, { type: TYPE, username })
+  }
+}
+
+
+export async function register(
```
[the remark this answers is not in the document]
icidasset: Yeah I documented that below, but I can add it here too 👍

src/components/depot/implementation.ts
```diff
@@ -0,0 +1,30 @@
...
+  // Get the data behind a CID
+  getBlock: (cid: CID) => Promise<Uint8Array>
+  getUnixFile: (cid: CID) => Promise<Uint8Array>
+  getUnixDirectory: (cid: CID) => Promise<DirectoryItem[]>
+
+  // Keep data around
+  putBlock: (data: Uint8Array, codec: BlockCodec<number, any>) => Promise<CID>
```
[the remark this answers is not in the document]
icidasset: I haven't been able to get it to work. They changed something recently and because of it I had to change to `any` 🤷‍♂️

matheus23:
There's some remaining usage of "confidences" in the README and CHANGELOG :v: @icidasset

src/linking/consumer.test.ts
```diff
@@ -106,29 +106,29 @@ describe("handle session key", async () => {
   })
 
   it("returns a warning when it receives a session key it cannot decrypt with its temporary private key", async () => {
-    const cfg = config.normalize()
-    const { rsaSize, hashAlg } = cfg
-    const temporaryRsaPairNoise = await rsa.makeKeypair(rsaSize, hashAlg, KeyUse.Exchange)
-    const rawSessionKeyNoise = utils.arrBufToStr(utils.base64ToArrBuf(exportedSessionKey), CharSize.B16)
+    const temporaryRsaPairNoise = await crypto.rsa.genKey()
+    const rawSessionKeyNoise = exportedSessionKey // utils.arrBufToStr(utils.base64ToArrBuf(exportedSessionKey), CharSize.B16)
```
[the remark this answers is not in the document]
icidasset: Old code 👍

icidasset:
> There's some remaining usage of "confidences" in the README and CHANGELOG 

Thanks for spotting that!

src/components/depot/implementation.ts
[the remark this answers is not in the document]
matheus23: `BlockCodec<number, any>` actually resolves to this:

```ts
interface BlockCodec {
  name: string
  code: number
  encode: (data: any) => ByteView<any>
  decode: (bytes: ByteView<any>) => any
}
```

We never actually use any field other than `code` in any implementations of `putBlock`, so perhaps we can simplify the whole thing to just `code: number`?

If implementations want to go from `code: number` => `BlockCodec` they can keep a map of supported codecs.

A `putBlock` implementation already gets encoded data, so it shouldn't actually make use of `encode` or `decode`, so I think that'd be fine :)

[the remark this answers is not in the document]
icidasset: Good point, I'll simply this 👍 I did like the type specificity of this though, could we do something more specific than number here 🤔

bgins (approved):
Amazing work! So much accomplished here. It's looking great! 💯 🎉 

As a summary, here are the things I have tested manually:

- `webnative.compositions` with `staging` set to true
- Account linking (tested in the [Webnative app template](https://github.com/webnative-examples/webnative-app-template) on staging and production)
- Backwards compatibility (tested with staging and production)
- Custom manners component with alterations to logging
- Custom manners component with filesystem hooks (using #428)
- Tested with the [Stored Wasm example](https://github.com/webnative-examples/stored-wasm-example) which uses `capabilities` and the Fission Auth Lobby

Thanks for all these changes! Excited to start using them in our apps. 🕺

src/filesystem.ts
```diff
@@ -1,82 +1,72 @@
...
- *                Will be used if a filesystem hasn't been created yet.
  */
-export async function loadFileSystem(
-  permissions: Maybe<Permissions>,
-  username?: string,
-  rootKey?: string
-): Promise<FileSystem> {
+export async function loadFileSystem({ config, dependents, rootKey, username }: {
+  config: Configuration
+  dependents: Dependents
```
matheus23, line 22: I'm playing around with this a little right now, and that seemed off.

Aren't these `dependencies`, not `Dependents`?
I.e. the `FileSystem` depends on `crypto`, `storage`, `reference`, etc. not the other way around :thinking:

icidasset: Hmm true, I'll change it 👍 
By the way, in case you hadn't seen this, there's a version of this function available on the `Program` that doesn't require to pass in components.

src/common/type-checks.ts
```diff
@@ -20,6 +20,10 @@ export const isBool = (val: unknown): val is boolean => {
   return typeof val === "boolean"
 }
 
+export function isCryptoKey(val: unknown): val is CryptoKey {
+  return hasProp(val, "algorithm") && hasProp(val, "extractable") && hasProp(val, "type")
```
matheus23, line 24: Should this just be
```suggestion
  return val instanceof CryptoKey
```
?

matheus23 (approved):
I looked through quite a bit of this PR and I got a feel for the refactor when working a bit on the dashboard.

I'd say we can go ahead and :shipit: :)

src/common/hex.node.test.ts
```diff
@@ -1,18 +1,20 @@
...
 
 describe("hex", () => {
 
   it("round trips to bytes and back out", () => {
     fc.assert(
-      fc.property(fc.array(fc.integer(16, 255)), data => {
-        const hexData = []
-        const buffers = []
-        const returnData = []
+      fc.property(fc.array(fc.integer()), data => {
```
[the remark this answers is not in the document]
icidasset: They removed support for this: https://github.com/dubzzz/fast-check/pull/2939

src/common/type-checks.ts
icidasset: Probably yeah 😄 Don't remember why I did it this way.

src/common/hex.node.test.ts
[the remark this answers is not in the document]
matheus23: No, they removed the `fc.integer(16, 255)` syntax, but `fc.integer({ min: 16, max: 255 })` should work.

src/common/type-checks.ts
icidasset: Undoing this change, for some reason the tests freeze with this 🤨 
https://github.com/fission-codes/webnative/actions/runs/3541918255/jobs/5946780663

PR #428, opened 2022-11-22
Add file system hooks

icidasset:
Adds file system hooks to the manners component.

This enables an easier flow for walletauth. Walletauth creates a read key, encrypts it and puts it in the public file system. After that, when loading a file system, it needs to decrypt that read key and inject it into the system, before loading the filesystem. So rather than "monkey patching" the functions that load the filesystem we use these hooks.

Other changes:
- Had to use a type parameter for the manners component.
- Removed manners component parameter from MMPT functions, wasn't being used.

### DIFF
--- a/src/common/cid.ts
+++ b/src/common/cid.ts
@@ -3,6 +3,9 @@ import { decode as decodeMultihash } from "multiformats/hashes/digest"
 import { hasProp, isNum, isObject } from "./type-checks.js"
 
 
+export { CID }
+
+
 /**
  * CID representing an empty string. We use this to speed up DNS propagation
  * However, we treat that as a null value in the code
--- a/src/components/manners/implementation.ts
+++ b/src/components/manners/implementation.ts
@@ -1,10 +1,21 @@
 import { CID } from "multiformats"
+
 import type { Configuration } from "../../configuration.js"
+
+import * as Depot from "../../components/depot/implementation.js"
+import * as Reference from "../../components/reference/implementation.js"
+
 import * as FileSystem from "../../fs/types.js"
 
 
 export type ImplementationOptions = {
   configuration: Configuration
+}
+
+
+export type DataFlowComponents = {
+  depot: Depot.Implementation
+  reference: Reference.Implementation
 }
 
 
@@ -29,10 +40,10 @@ export type Implementation = {
      * Various file system hooks.
      */
     hooks: {
-      afterLoadExisting: (fs: FileSystem.API) => Promise<void>
-      afterLoadNew: (fs: FileSystem.API) => Promise<void>
-      beforeLoadExisting: (cid: CID) => Promise<void>
-      beforeLoadNew: () => Promise<void>
+      afterLoadExisting: (fs: FileSystem.API, dataFlowComponents: DataFlowComponents) => Promise<void>
+      afterLoadNew: (fs: FileSystem.API, dataFlowComponents: DataFlowComponents) => Promise<void>
+      beforeLoadExisting: (cid: CID, dataFlowComponents: DataFlowComponents) => Promise<void>
+      beforeLoadNew: (dataFlowComponents: DataFlowComponents) => Promise<void>
     }
   }
 }
--- a/src/components/manners/implementation.ts
+++ b/src/components/manners/implementation.ts
@@ -1,4 +1,5 @@
 import type { Configuration } from "../../configuration.js"
+import * as FileSystem from "../../fs/types.js"
 
 
 export type ImplementationOptions = {
@@ -6,7 +7,7 @@ export type ImplementationOptions = {
 }
 
 
-export type Implementation<FileSystem> = {
+export type Implementation = {
   log: (...args: unknown[]) => void
   warn: (...args: unknown[]) => void
 
@@ -27,8 +28,8 @@ export type Implementation<FileSystem> = {
      * Various file system hooks.
      */
     hooks: {
-      afterLoadExisting: (fs: FileSystem) => Promise<void>
-      afterLoadNew: (fs: FileSystem) => Promise<void>
+      afterLoadExisting: (fs: FileSystem.API) => Promise<void>
+      afterLoadNew: (fs: FileSystem.API) => Promise<void>
       beforeLoadExisting: () => Promise<void>
       beforeLoadNew: () => Promise<void>
     }
--- a/src/components/manners/implementation.ts
+++ b/src/components/manners/implementation.ts
@@ -40,10 +40,10 @@ export type Implementation = {
      * Various file system hooks.
      */
     hooks: {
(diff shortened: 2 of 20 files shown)


src/fs/root/tree.ts
```diff
@@ -39,7 +39,7 @@ import { PublicRootWasm } from "../v3/PublicRootWasm.js"
 type Dependents = {
   crypto: Crypto.Implementation
   depot: Depot.Implementation
-  manners: Manners.Implementation
+  manners: Manners.Implementation<any>
```
icidasset, line 42: I'm using `<any>` here, because I want to indicate that it doesn't matter what the type of the 'file system' is that the hooks are using, because we aren't using the hooks here. Another reason is if we were to use the `FileSystem` class, we'd be introducing some circular imports (same for all the other places I used `<any>`)

bgins: Would like to avoid introducing `any` into the code base in all these places. Let's explore this to see if we can find an alternative.

First question is whether it is necessary to introduce the type parameter. It looks like the type parameter is referenced in the manners implementation.

https://github.com/fission-codes/webnative/blob/b381bea661eb826890d817bb3845dcfdf8fb6b00/src/components/manners/implementation.ts#L30-L31

Would it be possible to import the `FileSystem` as a type in this module instead?

```typescript
import type FileSystem from "../../fs/filesystem.js"
```

Other alternatives might be checking if we can use `unknown` instead of `any`, but removing the type parameter if we don't need it would be better.

icidasset: > Would like to avoid introducing `any` into the code base in all these places.

Me too.

> First question is whether it is necessary to introduce the type parameter

It is, otherwise we have circular imports (you can test this with `npm run test:imports`).
Of course, there might be another way around this issue, but I haven't found any yet.

> Other alternatives might be checking if we can use unknown instead of any

I have considered that, but that doesn't fit as well, because that means you'd need to be type checking in the implementation if you're actually dealing with a `FileSystem`. So that would be a weird experience for devs to write hooks.

`any` actually makes sense here, the type could be "anything" in this case, because the code doesn't use the hooks. Would be great if there was another way to indicate this to the type system, but I haven't found it.

icidasset: Ok, scratch that, I've implemented Philipp's suggestion that uses an interface for the file system.

src/fs/types.ts
```diff
@@ -2,10 +2,57 @@ import type { CID } from "multiformats/cid"
...
 import { PutResult } from "../components/depot/implementation.js"
-import { Path } from "../path/index.js"
+import { DirectoryPath, DistinctivePath, FilePath, Path } from "../path/index.js"
 import { Ucan } from "../ucan/types.js"
 
 
+// 💾 TOP LEVEL
+// ------------
+
+export type API = Exchange & Persistence & Posix & Sharing
```
icidasset, line 12: Better name for this? Just `FileSystem`? 🤷‍♂️

src/common/cid.ts
```diff
@@ -3,6 +3,9 @@ import { decode as decodeMultihash } from "multiformats/hashes/digest"
 import { hasProp, isNum, isObject } from "./type-checks.js"
 
 
+export { CID }
```
icidasset, line 6: So people don't necessarily need to add the multiformats library as a dependency? Because most people don't know what that is anyway.

src/components/manners/implementation.ts
```diff
@@ -18,4 +31,19 @@ export type Implementation = {
...
+   * File system.
+   */
+  fileSystem: {
+    /**
+     * Various file system hooks.
+     */
+    hooks: {
+      afterLoadExisting: (fs: FileSystem.API, dataFlowComponents: DataFlowComponents) => Promise<void>
+      afterLoadNew: (fs: FileSystem.API, dataFlowComponents: DataFlowComponents) => Promise<void>
+      beforeLoadExisting: (cid: CID, dataFlowComponents: DataFlowComponents) => Promise<void>
```
icidasset, line 45: Requirement for walletauth, I need to load part of the public filesystem to extract the encrypted root read key. I can't pass in the reference as a dependency during the creation of the manners components. because the reference component itself depends on the manners component. And sadly the `PublicTree` and `PrivateTree` components of the file system depend on the reference component, because they need to do DNS lookups. Should we add these hooks elsewhere? 🤷‍♂️

src/fs/types.ts
```diff
@@ -2,10 +2,61 @@ import type { CID } from "multiformats/cid"
...
 
+// 💾 TOP LEVEL
+// ------------
+
+export type API = Exchange & Persistence & Properties & Posix & Sharing
+
+export interface Exchange {
+  addPublicExchangeKey(): Promise<void>
+  hasPublicExchangeKey(): Promise<boolean>
+}
```
matheus23, line 17: Can we make these not part of the `FileSystem.API`? I think they're very similar to the `addSampleData` functions, so perhaps they should be somewhere near that function?

src/components/manners/implementation.ts
```diff
@@ -1,11 +1,24 @@
...
+import * as FileSystem from "../../fs/types.js"
+
 
 export type ImplementationOptions = {
   configuration: Configuration
 }
 
 
+export type DataFlowComponents = {
+  depot: Depot.Implementation
```
icidasset, line 17: Should possibly add the crypto component here as well.

src/components/manners/implementation/base.ts
```diff
@@ -9,6 +11,17 @@ export function implementation(opts: ImplementationOptions): Implementation {
...
     warn: opts.configuration.debug ? console.warn : () => { },
 
-    wnfsWasmLookup: wnfsVersion => fetch(`https://unpkg.com/wnfs@${wnfsVersion}/wasm_wnfs_bg.wasm`)
+    // WASM
+    wnfsWasmLookup: wnfsVersion => fetch(`https://unpkg.com/wnfs@${wnfsVersion}/wasm_wnfs_bg.wasm`),
+
+    // File system
+    fileSystem: {
+      hooks: {
+        afterLoadExisting: (fs: FileSystem.API) => addSampleData(fs),
```
bgins, line 20: A couple of questions here. First, should this function be `async`?

```suggestion
        afterLoadExisting: async (fs: FileSystem.API) => addSampleData(fs),
```

Second, not sure about adding sample data for every app. In product discussions with Jeff, we decided to remove the sample data for app-owned apps with the idea that they should control what data is created. Defaults won't really make sense in all cases.

bgins (approved):
Looks good. Tested it out writing a custom manners component with logging in each hook, working great. 🎉 

Left one comment about adding sample data by default.

src/components/manners/implementation/base.ts
icidasset: That function is `async` yeah, `addSampleData` returns a Promise.
Yeah we can remove the default, could use it as a hook for Fission Drive now 👍

PR #431, opened 2022-11-28
Fix "Could not locate user DID in DNS" uncaught exceptions in device linking

### LINKED ISSUE #430: Could not locate user DID in DNS uncaught exception after device linking
# Summary

## Problem

After successfully linking a device, the linked device reports a "Could not locate user DID in DNS" uncaught exception in the console.

### Impact

Device linking works, so there should be no impact for users. It will be confusing for developers because they may think something went wrong.

## Solution

Catch the exceptions or don't throw them until we need to. The changes will be to https://github.com/fission-codes/webnative/blob/db89623bbab458ccfdfa673b5e0804e36a1e3e7a/src/components/auth/channel.ts#L53-L82

# Detail

**To Reproduce**
1. Clone [Webnative app template](https://github.com/webnative-examples/webnative-app-template/tree/icidasset/webnative-0.35) which uses Webnative at `0.35.0-alpha-2`
2. Register a new user
3. Link the user's account on a new device
4. Observe error in the console

**Expected behavior**

No uncaught exception is reported.

**Screenshots**

CleanShot 2022-11-28 at 10 26 07@2x

bgins:
## Summary

This PR fixes/implements the following **bugs**

* [x] Fixes "Could not locate user DID in DNS" uncaught exceptions

Uncaught exceptions are reported in the console after device linking, but they are spurious because we retry a few times until we are successful. The change here removes the thrown exceptions until the last attempt.

## Test plan (required)

To test the change:

1. Clone [Webnative app template](https://github.com/webnative-examples/webnative-app-template/tree/icidasset/webnative-0.35) and update Webnative to use `bgins/fix-did-root-exceptions`
2. Register a new user
3. Link the user's account on a new device
4. No uncaught exceptions should be reported, but warnings may be displayed on failed attempts

## Closing issues

Fixes #430

## After Merge
* [ ] Does this change invalidate any docs or tutorials? It does not
* [ ] Does this change require a release to be made? To be released with the other `0.35.0` changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,10 +28,6 @@ The documentation should give you some information on how the various components
 - Removal of the `webnative.initialise`, `app` and `permissionedApp` functions. These have been replaced by the `program` function mentioned above.
 - Removal of the `fs.appPath` function, replaced with the `appData` function located in the path module.
 
-**Fixes**
-
-- Fixes `Could not locate user DID in DNS` uncaught exception reported after successfully linking a device
-
 ### v0.34.2
 
 - Fixes `LinkError: import object field '__wbg_putBlock_88cdb3be9020efb7' is not a Function` when loading WASM.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,10 @@ The documentation should give you some information on how the various components
 - Removal of the `webnative.initialise`, `app` and `permissionedApp` functions. These have been replaced by the `program` function mentioned above.
 - Removal of the `fs.appPath` function, replaced with the `appData` function located in the path module.
 
+**Fixes**
+
+- Fixes `Could not locate user DID in DNS` uncaught exception reported after successfully linking a device
+
 ### v0.34.2
 
 - Fixes `LinkError: import object field '__wbg_putBlock_88cdb3be9020efb7' is not a Function` when loading WASM.
--- a/package-lock.json
+++ b/package-lock.json
@@ -59,7 +59,7 @@
         "util": "^0.12.4"
       },
       "engines": {
-        "node": ">=15"
+        "node": ">=16"
       }
     },
     "node_modules/@achingbrain/ip-address": {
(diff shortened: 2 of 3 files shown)


avivash:
> @bgins should i be on a particular branch of WAT? when i run it on `main` and install `"webnative": "fission-codes/webnative#bgins/fix-did-root-exceptions"`, i get these errors in the console and the app keeps reloading 🤔 image

@bgins ahhh now i see your link actually points to steven's branch 😅 let me try that

bgins:
> @bgins ahhh now i see your link actually points to steven's branch 😅 let me try that

Oh yeah, I should have mentioned that in the description! Lemme know if that doesn't work out.

avivash (approved):
LGTM! 👍🏼

avivash:
> > @bgins ahhh now i see your link actually points to steven's branch 😅 let me try that
> 
> Oh yeah, I should have mentioned that in the description! Lemme know if that doesn't work out.

all good, everything's working as expected now :)

CHANGELOG.md
```diff
@@ -28,6 +28,10 @@ The documentation should give you some information on how the various components
 - Removal of the `webnative.initialise`, `app` and `permissionedApp` functions. These have been replaced by the `program` function mentioned above.
 - Removal of the `fs.appPath` function, replaced with the `appData` function located in the path module.
 
+**Fixes**
+
+- Fixes `Could not locate user DID in DNS` uncaught exception reported after successfully linking a device
```
icidasset, line 33: Was this a problem before 0.35?

icidasset (approved):
👍  I should note, I have fix for `loadFileSystem` which has a similar problem.

CHANGELOG.md
bgins: We did not notice it before `0.35`, but I didn't see anything in the refactor that would have caused it to occur. Maybe drop it from the changelog though since it didn't manifest before?

icidasset: Yeah I’d just drop it.

PR #432, opened 2022-11-28
0.35 fixes

icidasset:
Couple more adjustments and fixes:
- Exchange key functions have been moved (weren't part of the interface, so couldn't access them in hooks)
- IPFS peers weren't namespaced
- Replaced `namespaceToString` with a generic `namespace` function
- Issue with backwards compat, needed to look up root DID
- Missing async & await keywords in readme
- Use a version stamp for the migration indicator
- Check if databases exists before trying to open them


src/index.ts
```diff
@@ -613,17 +625,24 @@ async function ensureBackwardsCompatibility(components: Components, config: Conf
...
+    .then(v => v && Semver.isBiggerThanOrEqualTo(v, currentVersion))
+
   if (migrationOccurred) return
 
   // Only try to migrate if environment supports indexedDB
   if (!self.indexedDB) return
 
   // Migration
-  const keystoreDB = await bwOpenDatabase("keystore")
+  const existingDatabases = (await self.indexedDB.databases()).map(db => db.name)
```
bgins, line 644: Seeing this error in Firefox:

idx-db

It looks like `indexedDB.databases` is not supported in Firefox: https://developer.mozilla.org/en-US/docs/Web/API/IDBFactory/databases

bgins (approved):
Looks good overall! 🎉 

One fix is needed to support Firefox. Mentioned it in the comments.

src/index.ts
icidasset: Thanks! Added a fallback for Firefox and older browsers.

PR #433, opened 2022-12-02
Perform browser checks before components init

icidasset:
## Summary

This checks if the context and browser are supported before initialising the default components, instead of after. Certain components use indexedDB, so we want the check to be done before that code is called. It only checks when calling `.program()`, not `.assemble()` (see code docs for more info). `.isSupported()` is also exported.


avivash (approved):
tested locally, LGTM!

PR #434, opened 2022-12-02
Make session components private

icidasset:
## Summary

Small change to make the session's components private so that the session class instance is less confusing.


bgins (approved):
Looks good! I think hiding `crypto` and `storage` is a good approach here. 💯

PR #435, opened 2022-12-02
0.35 Release 

### LINKED ISSUE #438: previousCid -> previousCids()[0]
# Summary
My linter and build chain complain about [this line](https://github.com/fission-codes/webnative/blob/db89623bbab458ccfdfa673b5e0804e36a1e3e7a/src/fs/v3/PublicRootWasm.ts#L439)

## Problem

image

### Impact

build fails

## Solution
fix is easy previousCids()[0]
but not sure if that has side effects that you may need to consider.

### LINKED ISSUE #439: Only instantiate an IPFS repo when needed
Depends on the depot component, but when using the default IPFS depot component, don't instantiate (js-)ipfs right away.

icidasset:
Final PR before 0.35 release.

Changes:
- [x] Added DID aliases for `write` and `exchange` named `agent` and `sharing` respectively.
- [x] Added DID shorthand functions `program.agentDID()` and `program.sharingDID()` 
- [x] Added `src` folder to the NPM distribution
- [x] Added emoji to `program` function to indicate in API docs that's where to start
- [x] Default to extension-less imports (for people using webnative, not in our code)
- [x] Support for imports with extension is still there (but not for esbuild see note nr. 1)
- [x] Removed `loadRootFileSystem`
- [x] Expose config from program for convenience .
- [x] Fixes UMD build & adds `.esm.min` suffix to minified ESM build files.
- [x] Removes remaining usages of yarn.
- [x] Fixes #438
- [x] Fixes some encryption and encoding issues with the lobby flow.
- [x] Replaced some UCAN root issuer lookups with did-root lookups.
- [x] Forgot to add request options argument to `program.capabilities.request()`
- [x] Fixed issue with configuration extraction function (forgot to copy permissions)
- [x] Less usage of `window` object in fission-lobby implementation 
- [x] Closes #439 
- [x] Removed all traces of `ipfs-core` besides the CDN url (caused some issues with bundlers)
- [x] Updated IPFS from CDN to 0.17.0 to align with used `ipfs-core-types` 

Notes:
(1) Extension imports do not work in `esbuild` (and possibly bundlers depending on it). Because esbuild does not support "alternatives" in the `exports` package.json field, it always uses the first entry of the list. See https://webpack.js.org/guides/package-exports/#support for more info.


src/components/capabilities/implementation/fission-lobby.ts
```diff
@@ -289,27 +294,20 @@ async function translateClassifiedInfo(
...
-          returnValue = await action()
-        } catch (e) {
-          if (tryNum == options.tries) {
-            throw e
-          }
-        }
-        tryNum++
-      }
-      return returnValue || null
-    })(),
```
icidasset, line 308: This was loading the iframe 20 times at the same time instead of waiting for the timeout.

src/configuration.ts
```diff
@@ -1,15 +1,42 @@
...
+     * Capability collection settings.
+     */
+    collect?: {
+      /**
+       * If set to `true` then collection won't happen automatically
+       * when creating a `Program` (via `program` or `assemble`).
+       *
+       * @default false
+       */
+      manually?: boolean
```
icidasset, line 26: Open to naming suggestions 👀

avivash: hmm 🤔 how about `manual` or maybe `automatic`?

README.md
```diff
@@ -142,6 +142,7 @@ __Notes:__
 - You can use alternative authentication strategies, such as [webnative-walletauth](https://github.com/fission-codes/webnative-walletauth).
 - You can remove all traces of the user using `await session.destroy()`
 - You can load the file system separately if you're using a web worker. This is done using the combination of `configuration.fileSystem.loadImmediately = false` and `program.loadFileSystem()`
+- If you're using the capabilities system and workers, you may need to disable automatic capability collection if the collection method of the used capabilities component depends on the `window` object. You can disable this using `configuration.capabilities.collect.manually = true`
```
avivash: in reference to my other comment, i think naming it `capabilities.collect.automatic` makes sense since that's what's being used in the docs 👍🏼

src/permissions.ts
```diff
@@ -30,6 +45,10 @@ export type RawPermission = {
 }
 
 
+
+// 🛠
```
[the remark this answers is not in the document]
icidasset: Sounds good! I'll work on this in the next release if that's alright.
Won't use `/*` thought but instead `//`, or it might show up in the API docs.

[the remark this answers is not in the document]
avivash: Sure, that sounds great! :)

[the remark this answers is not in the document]
icidasset: Made an issue for this here: https://github.com/fission-codes/webnative/issues/442

src/components/depot/implementation/ipfs-default-pkg.ts
```diff
@@ -12,13 +16,19 @@ export async function implementation(
...
-  const [ ipfs, repo ] = await Ipfs.nodeWithPkg(
-    dependencies,
-    await Ipfs.pkgFromCDN(Ipfs.DEFAULT_CDN_URL),
-    peersUrl,
-    repoName,
-    false
-  )
-
-  return IpfsBase.implementation(ipfs, repo)
+  let instance: Maybe<{ ipfs: IPFS, repo: IPFSRepo }> = null
```
bgins, line 19:
```suggestion
  let instance: Maybe<{ ipfs: IPFS; repo: IPFSRepo }> = null
```

src/components/depot/implementation/ipfs/node.ts
```diff
@@ -119,7 +120,7 @@ export async function createAndConnect(
   peersUrl: string,
   repoName: string,
   logging: boolean
-): Promise<[ IPFSCore, IPFSRepo ]> {
+): Promise<{ ipfs: IPFSCore, repo: IPFSRepo }> {
```
bgins, line 123:
```suggestion
): Promise<{ ipfs: IPFSCore; repo: IPFSRepo }> {
```

src/components/reference/implementation/fission-base.ts
```diff
@@ -16,7 +16,13 @@ export async function implementation(endpoints: Endpoints, dependencies: Depende
   base.dataRoot.domain = (username: string) => `${username}.files.${endpoints.userDomain}`
   base.dataRoot.lookup = (...args) => DataRoot.lookup(endpoints, dependencies, ...args)
   base.dataRoot.update = (...args) => DataRoot.update(endpoints, dependencies, ...args)
-  base.didRoot.lookup = (...args) => DID.root(endpoints, ...args)
+  base.didRoot.lookup = (...args) => {
+    try {
+      return DID.root(endpoints, ...args)
+    } catch {
+      return Base.didRootLookup(dependencies, ...args)
```
icidasset, line 23: Some context, recently I had an issue that the root DID in my UCANs was different from the one in DNS. I assume because of account recovery or an issue with the CLI (it's very old account).

In any case, I'm not sure about this approach. I do know that the DID returned from this function should always be the same, otherwise I introduce a bunch of issues, such as problems loading the file system. So ideally I think we pick one or the other, not the above approach where it would use the local-only approach (Base) if the DNS lookup fails (eg. when online or DNS services are down).

I would opt for the local-only approach, but then we have the new UCAN spec that no longer uses nested UCANs in the way we use them now (will be referenced with CIDs). So I'm not sure in how far that will work offline.

Thoughts @bgins @matheus23 ?

src/configuration.ts
```diff
@@ -63,6 +72,11 @@ export type UserMessages = {
 // 🛠
 
 
+export function addRootFileSystemPermissions(config: Configuration): Configuration {
+  return { ...config, permissions: { ...config.permissions, ...ROOT_FILESYSTEM_PERMISSIONS } }
+}
```
bgins, line 77: Is this something that developers would use in their apps? Not seeing it used in our code.

icidasset: Ah right, don't need this anymore, I'll remove it.

src/components/reference/implementation/fission-base.ts
icidasset: Will default to DNS for now.

bgins (approved):
Looks good! Nice set of cleanups ✨

PR #449, opened 2022-12-15
Move shareLink function

icidasset:
Moves the `shareLink` function to a more reachable location.
Should we export this Fission module from the main module as well?


bgins (approved):
Looks good! 🎉 

> Should we export this Fission module from the main module as well?

Yep, this seems like a good idea to me. Would make it easier to access. 👍

PR #455, opened 2023-01-06
Add function overload signatures to path.appData()

icidasset:
`webnative.path.appData` was missing the correct function overload type signatures, causing it always to return a path of the type `DistinctivePath` instead of specifically `FilePath` or `DirectoryPath` depending what type of suffix path you give it (if any).

```ts
const root: DirectoryPath = wn.path.appData(appInfo)
const dir: DirectoryPath = wn.path.appData(appInfo, wn.path.directory("example"))
const file: FilePath = wn.path.appData(appInfo, wn.path.file("example.txt"))
```

### DIFF
--- a/src/path/index.ts
+++ b/src/path/index.ts
@@ -62,9 +62,11 @@ export function root(): DirectoryPath {
 /**
  * Utility function create an app data path.
  */
-export function appData(app: AppInfo, suffix?: DistinctivePath): DistinctivePath {
+export function appData(app: AppInfo, suffix?: FilePath): FilePath
+export function appData(app: AppInfo, suffix?: DirectoryPath): DirectoryPath
+export function appData(app: AppInfo, suffix?: DistinctivePath): DistinctivePath
+export function appData(app: AppInfo, suffix?: any): any {
   const parent = directory(Branch.Private, "Apps", app.creator, app.name)
-
   if (suffix) return combine(parent, suffix)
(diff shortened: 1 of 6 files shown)


src/path/index.ts
```diff
@@ -62,9 +62,12 @@ export function root(): DirectoryPath {
 /**
  * Utility function create an app data path.
  */
-export function appData(app: AppInfo, suffix?: DistinctivePath): DistinctivePath {
+export function appData(app: AppInfo): DirectoryPath
+export function appData(app: AppInfo, suffix: FilePath): FilePath
+export function appData(app: AppInfo, suffix: DirectoryPath): DirectoryPath
+export function appData(app: AppInfo, suffix: DistinctivePath): DistinctivePath
+export function appData(app: AppInfo, suffix?: any): any {
```
bgins, line 69: Could this be written without using `any`? 

It might work to replace the `any` with `DistinctivePath` because it is broad enough to cover each of the `suffix` overloads and the return type overloads.

avivash (approved):
thanks for doing this! I agree with Brian's comment, but looks good otherwise!

src/path/index.ts
icidasset: Done 👍 I didn't realise you could have two times the same function type with overloads, because the last type signature isn't exported and is only used for the implementation.

PR #456, opened 2023-01-06
Use better trigger for docs Github action

icidasset:
This will cause the docs site to update whenever a "stable" Github release is created. I first had the idea to do `releases/*` branches to trigger the docs but that was never done, and instead triggered the action manually.


bgins (approved):
Looks good to me!

avivash (approved):
🚀

PR #461, opened 2023-01-12
Avivash/File System Recovery Shorthand Method

avivash:
## Summary

* [x] The `recoverFileSystem` shorthand

Moving the FileSystem recovery functionality into `webnative`. Devs can now call the `recoverFileSystem` shorthand method directly from the `program`. It is important to note the `newUsername` and `oldUsername` that are passed in are the hashed versions, as `webnative` doesn't yet have a notion of the unhashed versions. (That will be completed in https://github.com/fission-codes/webnative/issues/385)

```ts
const { success } = await $sessionStore.program.recoverFileSystem({
  newUsername,
  oldUsername,
  readKey
})
```

~~**Note: I'm not sure if we currently have the infrastructure in the codebase for this, but it would be cool if we could emit events at various parts of the recovery flow to update the message on the client 🤔**~~ Answered by Steven!

## Test plan (required)

Load [WAT](https://github.com/webnative-examples/webnative-app-template/) using this `webnative` branch and go through the recovery flow. It should work the same as it did before.

https://www.loom.com/share/b2ecfe8fa8a04f7e96dab76d8dc749a6

## Closing issues

Closes https://github.com/fission-codes/webnative/issues/427

## After Merge
* [x] Does this change require a release to be made? Is so please create and deploy the release

## Sibling PR

https://github.com/webnative-examples/webnative-app-template/pull/113

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ### v0.35.2
 
+Adds `program.recoverFileSystem({ newUsername, oldUsername, readKey })` shorthand method so apps will no longer have to implement the FS recovery flow manually.
+
+### v0.35.2
+
 Fixes issue with the types of the `path.appData` function. Now has the correct overloads.
 
 ### v0.35.1
@@ -354,7 +358,7 @@ Upgrade to CIDv1.
   });
   ```
 
-- Those prerequisites are passed to the `wn.redirectToLobby` function.  
+- Those prerequisites are passed to the `wn.redirectToLobby` function.
   (So the auth lobby has the correct parameters to determine the permissions to ask the user)
 - Adds the ability to use multiple apps with one file system (closes #73)
 - The SDK now handles multiple UCANs
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-### v0.35.3
+### v0.36.0
 
 Adds `program.recoverFileSystem({ newUsername, oldUsername, readKey })` shorthand method so apps will no longer have to implement the FS recovery flow manually.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-### v0.35.2
+### v0.35.3
 
 Adds `program.recoverFileSystem({ newUsername, oldUsername, readKey })` shorthand method so apps will no longer have to implement the FS recovery flow manually.
 
--- a/README.md
+++ b/README.md
@@ -142,7 +142,7 @@ __Notes:__
 - You can use alternative authentication strategies, such as [webnative-walletauth](https://github.com/fission-codes/webnative-walletauth).
 - You can remove all traces of the user using `await session.destroy()`
 - You can load the file system separately if you're using a web worker. This is done using the combination of `configuration.fileSystem.loadImmediately = false` and `program.loadFileSystem()`
-- You can recover a file system if you've downloaded a Recovery Kit by calling `program.recoverFileSystem({ newUsername, oldUsername, readKey })`. The `oldUsername` and `readKey` can be parsed from the uploaded Recovery Kit and the `newUsername` can be generated before calling the function. Please refer to [this example](https://github.com/webnative-examples/webnative-app-template/blob/5498e7062a4578028b8b55d2ac4c611bd5daab85/src/components/auth/recover/HasRecoveryKit.svelte#L49) from Fission's Webnative App Template
+- You can recover a file system if you've downloaded a Recovery Kit by calling `program.recoverFileSystem({ newUsername, oldUsername, readKey })`. The `oldUsername` and `readKey` can be parsed from the uploaded Recovery Kit and the `newUsername` can be generated before calling the function. Please refer to [this example](https://github.com/webnative-examples/webnative-app-template/blob/5498e7062a4578028b8b55d2ac4c611bd5daab85/src/components/auth/recover/HasRecoveryKit.svelte#L49) from Fission's Webnative App Template. Additionally, if you would like to see how to generate a Recovery Kit, you can reference [this example](https://github.com/webnative-examples/webnative-app-template/blob/main/src/lib/account-settings.ts#L186)
 
 
 ## Working with the file system
--- a/README.md
+++ b/README.md
@@ -142,6 +142,7 @@ __Notes:__
 - You can use alternative authentication strategies, such as [webnative-walletauth](https://github.com/fission-codes/webnative-walletauth).
 - You can remove all traces of the user using `await session.destroy()`
 - You can load the file system separately if you're using a web worker. This is done using the combination of `configuration.fileSystem.loadImmediately = false` and `program.loadFileSystem()`
+- You can recover a file system if you've downloaded a Recovery Kit by calling `program.recoverFileSystem({ newUsername, oldUsername, readKey })`. The `oldUsername` and `readKey` can be parsed from the uploaded Recovery Kit and the `newUsername` can be generated before calling the function. Please refer to [this example](https://github.com/webnative-examples/webnative-app-template/blob/5498e7062a4578028b8b55d2ac4c611bd5daab85/src/components/auth/recover/HasRecoveryKit.svelte#L49) from Fission's Webnative App Template
 
 
 ## Working with the file system
--- a/src/filesystem.ts
+++ b/src/filesystem.ts
@@ -1,26 +1,23 @@
 import { CID } from "multiformats/cid"
 
+import * as Crypto from "./components/crypto/implementation.js"
 import * as Depot from "./components/depot/implementation.js"
-import * as Reference from "./components/reference/implementation.js"
-
+import * as DID from "./did/index.js"
 import * as Protocol from "./fs/protocol/index.js"
+import * as Reference from "./components/reference/implementation.js"
+import * as RootKey from "./common/root-key.js"
+import * as Storage from "./components/storage/implementation.js"
+import * as Ucan from "./ucan/index.js"
 import * as Versions from "./fs/versions.js"
 
-import FileSystem, { Dependencies } from "./fs/filesystem.js"
-
-import * as Ucan from "./ucan/index.js"
-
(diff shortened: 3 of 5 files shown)


src/index.ts
```diff
@@ -49,12 +49,19 @@ import { Components } from "./components.js"
...
 
+// TYPES
+
+
+import { type RecoverFileSystemParams } from "./fs/types/params"
+
+
 // IMPLEMENTATIONS
 
+
```
avivash, line 64: adjusted the spacing here to be consistent with the other import/export sections at the top of the file

src/index.ts
```diff
@@ -485,8 +496,10 @@ export async function assemble(config: Configuration, components: Components): P
 
   // Shorthands
   const shorthands: ShortHands = {
-    loadFileSystem: (username: string) => loadFileSystem({ config, username, dependencies: components }),
-
+    loadFileSystem: (username: string) =>
+      loadFileSystem({ config, username, dependencies: components }),
+    recoverFileSystem: (params: RecoverFileSystemParams) =>
+      recoverFileSystem({ agentDID: () => DID.agent(components.crypto), auth, dependencies: components, ...params }),
```
avivash, line 502: i could maybe define `agentDID` before this section since it's used in two places now 🤔

avivash:
ahh, i'll update the changelog!

src/index.ts
avivash: i ended up doing this

src/filesystem.ts
```diff
@@ -8,10 +8,18 @@ import * as Versions from "./fs/versions.js"
...
 
+import * as Ucan from "./ucan/index.js"
+
+import * as RootKey from "./common/root-key.js"
+
 import { Branch } from "./path/index.js"
 import { Configuration } from "./configuration.js"
 import { Maybe, decodeCID, EMPTY_CID } from "./common/index.js"
 
+import { type RecoverFileSystemParams } from "./fs/types/params"
```
icidasset, line 19: `.js` suffix missing here. I think we still need it for some bundlers 😞

src/filesystem.ts
```diff
@@ -105,6 +113,70 @@ export async function loadFileSystem({ config, dependencies, rootKey, username }
...
+  readKey,
+}: {
+  agentDID: () => Promise<string>
+  auth: AuthenticationStrategy
+  dependencies: Dependencies
+} & RecoverFileSystemParams): Promise<{ success: boolean }> {
+  try {
+    const { crypto, reference, storage } = dependencies
+
+    const newRootDID = await agentDID()
```
icidasset, line 134: I would call `DID.agent(dependencies.crypto)` from `import * as DID from "./did/index.js"` instead of passing `agentDID` as a parameter.

src/filesystem.ts
```diff
@@ -105,6 +113,70 @@ export async function loadFileSystem({ config, dependencies, rootKey, username }
...
+
+    const oldRootCID = await reference.dataRoot.lookup(oldUsername)
+    if (!oldRootCID) {
+      throw new Error("Failed to lookup oldUsername")
+    }
+
+    // Update the dataRoot of the new user
+    await reference.dataRoot.update(oldRootCID, ucan)
+
+    // Store the accountDID, which is used to namespace the readKey
```
icidasset, line 164: Might be more correct to say: "Store the read key which namespaced using the account DID"?

src/filesystem.ts
```diff
@@ -105,6 +113,70 @@ export async function loadFileSystem({ config, dependencies, rootKey, username }
...
+    await RootKey.store({
+      accountDID: newRootDID,
+      crypto: crypto,
+      readKey,
+    })
+
+    return {
+      success: true
+    }
+  } catch (error) {
```
icidasset, line 174: Does it still print the error in the console when catching it?

src/index.ts
```diff
@@ -49,12 +49,19 @@ import { Components } from "./components.js"
...
 import { Session } from "./session.js"
-import { loadFileSystem } from "./filesystem.js"
+import { loadFileSystem, recoverFileSystem } from "./filesystem.js"
 import FileSystem from "./fs/filesystem.js"
 
 
+// TYPES
+
+
+import { type RecoverFileSystemParams } from "./fs/types/params"
```
icidasset, line 59: Missing `.js` suffix in the filename here too.

icidasset (commented):
Couple comments, looks great otherwise! 🥳

icidasset:
No event emitter yet in Webnative, but it's on the short-term todo list! Also, bit confused about the example shown here, this should work with regular usernames too right?

src/filesystem.ts
avivash: ahh, good point! I just removed this try/catch so the error can instead be caught from the call site in the client app 👍🏼

avivash:
>No event emitter yet in Webnative, but it's on the short-term todo list! 

Nice! 

>Also, bit confused about the example shown here, this should work with regular usernames too right?

Ahh yes good point! I should mention this works regardless of whether the username is hashed, I'll update the example 👌🏼

src/filesystem.ts
```diff
@@ -105,6 +115,65 @@ export async function loadFileSystem({ config, dependencies, rootKey, username }
...
+ */
+export async function recoverFileSystem({
+  auth,
+  dependencies,
+  oldUsername,
+  newUsername,
+  readKey,
+}: {
+  auth: AuthenticationStrategy
+  dependencies: Dependencies
```
icidasset, line 129: I'd only ask for the components this function actually needs, instead of using the ones needed to load a file system. Less friction if people want to experiment.
```ts
import * as Crypto from "../components/crypto/implementation.js"
import * as Reference from "../components/reference/implementation.js"
import * as Storage from "../components/storage/implementation.js"

dependencies: {
  crypto: Crypto.Implementation
  reference: Reference.Implementation
  storage: Storage.Implementation
}
```

icidasset (approved):
One more nitpick.
Otherwise I think it's good to go.

src/filesystem.ts
avivash: ahh good point, will update!

README.md
```diff
@@ -142,6 +142,7 @@ __Notes:__
 - You can use alternative authentication strategies, such as [webnative-walletauth](https://github.com/fission-codes/webnative-walletauth).
 - You can remove all traces of the user using `await session.destroy()`
 - You can load the file system separately if you're using a web worker. This is done using the combination of `configuration.fileSystem.loadImmediately = false` and `program.loadFileSystem()`
+- You can recover a file system if you've downloaded a Recovery Kit by calling `program.recoverFileSystem({ newUsername, oldUsername, readKey })`. The `oldUsername` and `readKey` can be parsed from the uploaded Recovery Kit and the `newUsername` can be generated before calling the function. Please refer to [this example](https://github.com/webnative-examples/webnative-app-template/blob/5498e7062a4578028b8b55d2ac4c611bd5daab85/src/components/auth/recover/HasRecoveryKit.svelte#L49) from Fission's Webnative App Template
```
bgins, line 145: We may want to add a bit more context about how a recovery kit is created. If the steps are simple enough, we could list them here. Or maybe link to where we create them in WAT.

Another possibility would be adding another shorthand that makes it easy for devs to get the `username` and `readKey` so they can create recovery kits with them.

Thoughts?

src/filesystem.ts
```diff
@@ -8,10 +8,20 @@ import * as Versions from "./fs/versions.js"
 
 import FileSystem, { Dependencies } from "./fs/filesystem.js"
 
+import * as Ucan from "./ucan/index.js"
```
bgins, line 11: Could we re-order the imports a bit? The order we have generally been using is:

- Imports from external libraries
- Imports from within Webnative

Within each of those:

- Namespace imports (e.g. `import * as name from "module-name"`)
- Named imports (e.g. `import { export1, export2 } from "module-name"`)
- Default imports (e.g. `import defaultExport from "module-name"`)

The imports in each of those should be alphabetized. 

For example: https://github.com/fission-codes/webnative/blob/730d8e1ae489bc1ed2a148ba8f756712b08c2b66/src/index.ts#L26-L53

bgins (approved):
Looking good! ✨ 

Left a few comments. One thing we want to consider how developers will create a recovery kit to use this shorthand function. More on that in the comments.

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+### v0.35.3
```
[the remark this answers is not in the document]
avivash: will do!

src/filesystem.ts
avivash: cool, are you thinking something more like this? 

```ts
import { CID } from "multiformats/cid"

import * as Crypto from "./components/crypto/implementation.js"
import * as Depot from "./components/depot/implementation.js"
import * as DID from "./did/index.js"
import * as Reference from "./components/reference/implementation.js"
import * as Storage from "./components/storage/implementation.js"
import * as Protocol from "./fs/protocol/index.js"
import * as Versions from "./fs/versions.js"
import * as RootKey from "./common/root-key.js"
import * as Ucan from "./ucan/index.js"

import { AuthenticationStrategy } from "./index.js"
import { Branch } from "./path/index.js"
import { Configuration } from "./configuration.js"
import { Maybe, decodeCID, EMPTY_CID } from "./common/index.js"
import { type RecoverFileSystemParams } from "./fs/types/params.js"

import FileSystem, { Dependencies } from "./fs/filesystem.js"
```

avivash: edited that, it wasn't in alphabetical order the first time i did it 😅

README.md
avivash: good point! I think i'll add a link to the WAT example for now. I think adding a shorthand for generating the kit will be a bit too complicated until we've ported the username scheme/hashing functionality over to webnative. What do you think?

src/filesystem.ts
avivash: done 👍🏼

README.md
bgins: Sounds good! We can add that down the road. 💯

src/filesystem.ts
bgins: Yep, that looks good! ✨

PR #463, opened 2023-01-18
Add shorthand accountDID method

### LINKED ISSUE #443: Add shorthand rootDID method
`program.accountDID` which is a shorthand for `components.reference.didRoot.lookup(username)`

icidasset:
Closes #443

```ts
program.accountDID(username)
```
I thought about doing just `program.accountDID()` without the username param which would depend on the session underneath and return `null` if not authed. But seeing that `loadFileSystem` also takes the `username` parameter, it makes sense to implement it that way.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,8 @@
 
 ### v0.36.0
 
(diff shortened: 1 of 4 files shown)


avivash (approved):
ahh nice, looks good!

PR #464, opened 2023-01-18
Add file system data functions as shorthand methods

### LINKED ISSUE #448: Export file system data functions from main module
Currently to get to the `addPublicExchangeKey` and other file system data functions you need to import `webnative/fs/data`. It might be good to export this from the main `webnative` module as well for convenience and people not using a bundler.

Maybe something like?
```ts
webnative.fileSystem.addPublicExchangeKey()
```

icidasset:
Closes #448

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -159,9 +159,6 @@ export type ShortHands = {
   sharingDID: () => Promise<string>
 
   // File system
-  loadFileSystem: (username: string) => Promise<FileSystem>
-  recoverFileSystem: (params: RecoverFileSystemParams) => Promise<{ success: boolean }>
-
   fileSystem: {
     addPublicExchangeKey: (fs: FileSystem) => Promise<void>
     addSampleData: (fs: FileSystem) => Promise<void>
@@ -440,7 +437,7 @@ export const storage = {
  * - Provides shorthands to functions so you don't have to pass in components.
  * - Ensure backwards compatibility with older Webnative clients.
  *
- * See `loadFileSystem` if you want to load the user's file system yourself.
+ * See the `program.fileSystem.load` function if you want to load the user's file system yourself.
  */
 export async function assemble(config: Configuration, components: Components): Promise<Program> {
   const permissions = config.permissions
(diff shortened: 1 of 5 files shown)


avivash (approved):
awesome, much more scalable! 🎉

src/index.ts
```diff
@@ -568,14 +573,16 @@ export async function assemble(config: Configuration, components: Components): P
...
-    loadFileSystem: (username: string) =>
-      loadFileSystem({ config, username, dependencies: components }),
-    recoverFileSystem: (params: RecoverFileSystemParams) =>
-      recoverFileSystem({
-        auth,
-        dependencies: { crypto: components.crypto, reference: components.reference, storage: components.storage },
-        ...params,
-      }),
+    loadFileSystem: (username: string) => loadFileSystem({ config, username, dependencies: components }),
+    recoverFileSystem: (params: RecoverFileSystemParams) => recoverFileSystem({ auth, dependencies: components, ...params }),
```
bgins, line 577: Now that we have `fileSystem.load` and `fileSystem.recover`, should we remove the `loadFileSystem` and `recoverFileSystem` shorthands?

src/index.ts
```diff
@@ -568,14 +573,16 @@ export async function assemble(config: Configuration, components: Components): P
...
+
+    fileSystem: {
+      addPublicExchangeKey: (fs: FileSystem) => FileSystemData.addPublicExchangeKey(components.crypto, fs),
+      addSampleData: (fs: FileSystem) => FileSystemData.addSampleData(fs),
```
bgins, line 581: What are the use cases where developer would want to use this shorthand? I could see where it could see it being useful for quick experimentation, but I don't think developers will want exactly the directories it creates.

If we are the primary consumer of the shorthand, we should consider leaving it buried to keep the top-level API as simple as possible.

src/index.ts
icidasset: We could yeah.

src/index.ts
icidasset: The public key ones would be for whenever you have an app that you want to receive a private share on. The behaviour of these functions will change over time (eg. rs-wnfs will keep exchange keys in the root).

The sample data one would be for apps like Drive, which is less useful, but doesn't hurt to have it there I guess, even if it's for quick experimentation.

bgins: Ah yeah, the public key ones are good additions 👌. I was questioning `addSampleData`.

Adding `addSampleData` has some cost because it's another bit of API for developers to learn. If it has a good use case for them, then it's worth adding. Otherwise, we should consider leaving it out.

It seems like quick experimentation is a secondary use case. If we want to support the quick experimentation, maybe `addSampleData` could add a file in addition to directories?

icidasset: I'll add a file 👍

bgins (approved):

PR #465, opened 2023-01-18
Remove fs.add and fs.cat

### LINKED ISSUE #440: Remove fs.add and fs.cat
Use `fs.read` and `fs.write` instead.

icidasset:
Matches the interface of rs-wnfs.

Closes #440 

Guide will need updating:
https://guide.fission.codes/developers/webnative/file-system-wnfs

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,8 +2,15 @@
 
 ### v0.36.0
 
+#### Breaking changes
+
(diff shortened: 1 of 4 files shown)


bgins (approved):
Yes! This is an excellent improvement, `read` and `write` are much clearer. 💯

avivash (approved):
nice, i like not having these aliased anymore 👍🏼 i'll look through the templates to confirm we aren't using any of the old aliased methods

PR #466, opened 2023-01-20
Use stricter types for path creation

### LINKED ISSUE #457: Improve path creation
In scenarios where a `Path.Branch` is needed a path should be provided that includes a branch.

Context: https://github.com/fission-codes/webnative/issues/362#issuecomment-1078850705

icidasset:
This restricts the paths that can be used with the file system POSIX interface. For example, `fs.write` now requires a file path that starts with `"private"` or `"public"` and must have at least one more path segment.

```ts
fs.write(
  path.file("private", "fileName")
)
```

The other neat thing about this is when you start writing the following in your editor with the Typescript LSP:
```ts
await fs.write(
  Path.file("")
)
```
It'll suggest `"private"` or `"public"` so people don't have to guess anymore what to type in there as the first item.

## Further notes

- `Branch` has been renamed to `RootBranch`
- Introduces the concept of `Partition`s which are `RootBranch`es that can be used the POSIX file system interface.
- You can also use the `RootBranch` enum to provide string values, for the people that prefer to use a constant.
- I might have unnecessarily restricted some functions, let me know if you find any.
- My only worry about this is that people won't be able to write these types themselves. That said, I don't think anyone will need to.

Thoughts?

## How to write these types by hand

```ts
import { Partitioned, PartitionedNonEmpty, Segments, SegmentsNonEmpty } from "webnative/path/index"
import { Partition, Private, Public } from "webnative/path/index"
import { DirectoryPath, FilePath } from "webnative/path/index"

// Regular old path, no restrictions
path.file() as FilePath<Segments> // { file: [] }

// Must have at least one segment
path.file("segment") as FilePath<SegmentsNonEmpty>

// Must use a WNFS partition, can be public or private partition
path.directory("public") as DirectoryPath<Partitioned<Partition>>>

// Must use the private partition
path.directory("private") as DirectoryPath<Partitioned<Private>>>

// Must use the private partition and have at least one segment besides the partition
path.directory("private", "dir") as DirectoryPath<PartitionedNonEmpty<Private>>>
```

## Closes

Closes #457

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,10 @@
 - Adds `program.accountDID(username)` shorthand method to retrieve the root DID of a given account username.
 - Adds the file system data functions as shorthand methods.
 
+#### Other
+
+- Introduces stricter types for paths to restrict the paths you can use with various functions and also to guide you a bit more.
+
 
 ### v0.35.2
 
--- a/src/fs/filesystem.ts
+++ b/src/fs/filesystem.ts
@@ -16,7 +16,7 @@ import * as TypeChecks from "../common/type-checks.js"
 import * as Ucan from "../ucan/index.js"
 import * as Versions from "./versions.js"
 
-import { Branch, Branched, DistinctivePath, DirectoryPath, FilePath } from "../path/index.js"
+import { RootBranch, Partitioned, PartitionedNonEmpty, Partition, DistinctivePath } from "../path/index.js"
 import { Permissions } from "../permissions.js"
 import { SymmAlg } from "../components/crypto/implementation.js"
 import { decodeCID } from "../common/index.js"
@@ -204,7 +204,7 @@ export class FileSystem implements API {
   // POSIX INTERFACE (DIRECTORIES)
   // -----------------------------
 
-  async ls(path: Path.Directory<Branched>): Promise<Links> {
+  async ls(path: Path.Directory<Partitioned<Partition>>): Promise<Links> {
     if (Path.isFile(path)) throw new Error("`ls` only accepts directory paths")
     return this.runOnNode(path, {
       public: async (root, relPath) => {
@@ -220,7 +220,7 @@ export class FileSystem implements API {
     })
   }
 
-  async mkdir(path: Path.Directory<Branched>, options: MutationOptions = {}): Promise<this> {
+  async mkdir(path: Path.Directory<PartitionedNonEmpty<Partition>>, options: MutationOptions = {}): Promise<this> {
     if (Path.isFile(path)) throw new Error("`mkdir` only accepts directory paths")
 
     await this.runMutationOnNode(path, {
@@ -246,7 +246,7 @@ export class FileSystem implements API {
   // -----------------------
 
   async write(
-    path: Path.Distinctive<Branched>,
+    path: Path.Distinctive<PartitionedNonEmpty<Partition>>,
     content: Uint8Array | SoftLink | SoftLink[] | Record<string, SoftLink>,
     options: MutationOptions = {}
   ): Promise<this> {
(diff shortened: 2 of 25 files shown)


CHANGELOG.md
```diff
@@ -13,6 +13,10 @@
 - Adds `program.accountDID(username)` shorthand method to retrieve the root DID of a given account username.
 - Adds the file system data functions as shorthand methods.
 
+#### Other
+
+- Introduces stricter types for paths to restrict the paths you can use with various functions and also to guide you a bit more.
```
icidasset, line 18: I've put this here since it shouldn't be a breaking change.
Unless you've been using the file system incorrectly 🙈

src/fs/filesystem.ts
```diff
@@ -43,12 +43,6 @@ import * as Sharing from "./share.js"
 // TYPES
 
 
-export interface AppPath {
-  (): DirectoryPath
-  (path: DirectoryPath): DirectoryPath
-  (path: FilePath): FilePath
-}
```
icidasset, line 50: Forgot to remove this, leftovers from 0.34

avivash (approved):
woot! 🚀

src/fs/types.ts
```diff
@@ -21,32 +23,32 @@ export interface Properties {
 }
 
 export interface Posix {
-  exists(path: DistinctivePath): Promise<boolean>
-  get(path: DistinctivePath): Promise<PuttableUnixTree | File | null>
-  mv(from: DistinctivePath, to: DistinctivePath): Promise<this>
-  rm(path: DistinctivePath): Promise<this>
+  exists(path: DistinctivePath<PartitionedNonEmpty<Partition>>): Promise<boolean>
```
bgins, line 26: Are the `public` and `private` partitions guaranteed to exist? Is there any time that a developer would need to check for them? If so, this type may prevent them from doing so.

src/fs/types.ts
```diff
@@ -21,32 +23,32 @@ export interface Properties {
...
   // Directories
-  ls(path: DirectoryPath): Promise<Links>
-  mkdir(path: DirectoryPath, options?: MutationOptions): Promise<this>
+  ls(path: DirectoryPath<Partitioned<Partition>>): Promise<Links>
+  mkdir(path: DirectoryPath<PartitionedNonEmpty<Partition>>, options?: MutationOptions): Promise<this>
 
   // Files
   write(
-    path: DistinctivePath,
+    path: DistinctivePath<PartitionedNonEmpty<Partition>>,
```
bgins, line 40: Can this be restricted further to a `FilePath`? Are there cases where it is possible to write to a directory?

src/path/index.ts
```diff
@@ -116,52 +163,68 @@ export function toPosix(
...
+export function isPartition(partition: Partition, path: DistinctivePath<Segments>): boolean {
+  return unwrap(path)[ 0 ] === partition
+}
+
+/**
+ * Is this `DistinctivePath` of the given `RootBranch`?
  */
-export function isFile(path: DistinctivePath): path is FilePath {
-  return !!(path as FilePath).file
+export function isRootBranch(rootBranch: RootBranch, path: DistinctivePath<Segments>): boolean {
```
bgins, line 206: The name `isRootBranch` makes me think we are checking if a path is the `private` or `public` directory. What would you think about renaming this to something like `isOnRootBranch`?

src/path/index.ts
```diff
@@ -116,52 +163,68 @@ export function toPosix(
...
 /**
- * Is this `DistinctivePath` a file?
+ * Is this `DistinctivePath` of the given `Partition`?
+ */
+export function isPartition(partition: Partition, path: DistinctivePath<Segments>): boolean {
+  return unwrap(path)[ 0 ] === partition
+}
+
+/**
+ * Is this `DistinctivePath` of the given `RootBranch`?
```
bgins, line 204:
```suggestion
 * Is this `DistinctivePath` on the given `RootBranch`?
```

Possible edit.

bgins (approved):
Looks good! 🎉 

I left a couple of comments around where the types might be too restrictive or not restrictive enough, and one naming thing.

Love how the LSP suggests "public" or "private". Nice added bonus. ❤️

src/path/index.ts
icidasset: Sounds good, will rename!

PR #467, opened 2023-01-23
remove program.session property

nichoth:
Remove the property `program.session` from the program object, because it was confusing having multiple ways to access the same `session` object. 

After this PR, there is one way to access the session:
```js
session.value = await program.auth.session()
```

This eliminates the potential confusion around requesting capabilities vs using an authentication strategy — https://github.com/snail-situation/hermes/discussions/42

------

[see chat log](https://discord.com/channels/478735028319158273/678353918752718848/1066454227435659355)

------

The tests still pass after I removed the `program.session` property.

----

Is there something I have overlooked or am not aware of? Would this break any potential use cases I am not considering?

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -143,7 +143,6 @@ export type Program = ShortHands & {
   }
   configuration: Configuration
   components: Components
-  session: Maybe<Session>
 }
 
 
@@ -579,14 +578,12 @@ export async function assemble(config: Configuration, components: Components): P
     }
   }
 
-  // Fin
   return {
     ...shorthands,
     configuration: { ...config },
     auth,
     components,
-    capabilities,
-    session,
+    capabilities
   }
 }
 


icidasset:
@nichoth I agree that this makes sense, but there's one issue. The `program.capabilities.session` function requires `username` as the parameter and with capabilities the only way to get the username is via `program.capabilities.collect` (eg. with the lobby you only get the username when redirecting back from the lobby). In other words, the username wouldn't always be available to you, you would need to call `await SessionMod.restore(components.storage)`, which is exactly what `.program()` is doing here for you.

Thoughts @bgins ?

bgins:
Agreed that this would increase the complexity of retrieving a capabilities-based session. Out of curiosity, would it be possible to update `program.capabilities.session` so that it does not require the username parameter and figures out the username internally?

A separate consideration is that it is a nice DX convenience to make the `session` available on the `program` at initialization. It spares the developer from calling a function to determine the `session` when Webnative can determine it for them.

Perhaps we could document this more effectively to remove confusion?

icidasset:
> Out of curiosity, would it be possible to update `program.capabilities.session` so that it does not require the username parameter and figures out the username internally?

Not possible at the moment, the flow depends on the developer calling `program.capabilities.collect()` which returns the username. I didn't want the session to call `.collect()` for you because that would introduce some overhead, and it's also implementation dependant so it might do something entirely unexpected.

That said, the flow can definitely be improved, but I want to hold off on that until we get to app to app linking.

> Perhaps we could document this more effectively to remove confusion?

Yes 👍 

---

Thanks @nichoth ! We'll improve the documentation for now and we'll revisit this later when we're going to look at app to app linking.

bgins:
I've updated the description of `program.session` in the guide: https://guide.fission.codes/developers/webnative/initialization

> **Session.** A session including a username and a filesystem instance. A session will be created from an authentication strategy or requested capabilities depending on which approach your application takes.

Issue #414, opened 2022-08-22
Add an interface to subscribe to root CID changes


bgins:
# Summary

## Problem

Developers need to poll the filesystem to detect changes.

### Impact

Extra work and complexity for developers.

## Solution

Add an interface to subscribe to any change to WNFS. The interface should notify developers when a change occurs locally and remotely.

# Detail

It's currently possible to add a callback to the WNFS publish hooks, which are called when a change is made locally. We can likely use the existing work as a starting point or hook into it.

bgins:
A couple of rough ideas for how we might implement this feature:

- Add an event emitter that emits local and remote change events
- Make the event emitter available to developers at webnative initialization
- Emit local events based on existing publish hooks
- After each local event, poll internally for remote changes and and emit an event when change is confirmed remotely

Note that this would be a first quick implementation with a focus on getting it working and establishing a good interface. A subsequent implementation could set up PubSub with the server over a websockets channel to remove the internal polling.

Issue #413, opened 2022-08-22
TRACKING: WNFS event listeners


bgins:
This issue tracks changes to be made to implement event listeners in WNFS. This feature was initially requested in #136.

As a quick summary, it would be helpful to provide an interface where users can subscribe to changes when:
 
- A change is made anywhere in the filesystem
- A specific file or directory changes

In both cases, it would be useful to know when a change occurs locally and on a remote node.

### Tasks

- [ ] Add an interface to subscribe to root CID changes (a change anywhere in the filesystem)
- [ ] Add an interface to subscribe to file and directory changes

therealjeffg:
Note: remote changes are blocked on the server project.

Issue #163, opened 2021-01-13
WNFS Conflict Resolution


expede:
NB: Feature requests will only be considered if they solve a pain

# Summary

## Problem

Concurrent updates across devices can cause data loss by:

1. Dropping history (overwriting revisions)
2. Forgetting files in the latest generation when reconciling with new data

There is a server-side conflict detection portion of this, but the front end needs to know how to reconcile changes first.

### Impact

Sometimes users loose recent changes or files

## Solution

"Rebase" changes of the local WNFS on top of a remote source of truth at a conflict point.

This means being able to "pop off" the changes since the last common CID, and place them on top of the latest CID from the single source of truth.

The one wrinkle is that we don't want to loose any new files, so this needs to add any missing links back to the new CID head, and propagate those through the history. If you recall this image:

<img width="925" alt="Screen Shot 2021-01-12 at 9 35 08 PM" src="https://user-images.githubusercontent.com/1052016/104410869-1256f280-551e-11eb-88be-473cf96b44bb.png">

We're keeping the blue boxes (new files), but adjusting the blue lines to point to all existing files at the previous step.

Not that this does _not_ include sub-file merging. We cannot understand their content structure ahead of time, and even then it often goes awry. _Complete individual files_ can be overwritten, and we keep the history if they want the old copy or to do manual merging of the inner contents.

expede:
> > There is a server-side conflict detection portion of this, but the front end needs to know how to reconcile changes first.
> 
> I think there's value in implementing the server-side portion of this without implementing the front-end change reconciling in webnative. E.g. I could already implement conflict resolution on the app layer in flatmate if the server-side portion existed!
> 
> At the moment I can't do this, as any `.publish` call might overwrite any changes made by anyone else. So I'd need some kind of `.publishIfTheServerAgreesWithMeThatTheMostRecentStateWas(cid)`. 😄

Right, or you could ask for the latest head before you apply your changes https://runfission.com/docs/#tag/User%2Fpaths%2F~1user~1data~1%7Busername%7D%2Fget

You'll need to do the reconciliation in either case. Just knowing that you're behind only gets you so far, you still need WNFS to merge the changes.

matheus23:
> Right, or you could ask for the latest head before you apply your changes with the `GET /user/data/{username} endpoint`

That's my plan for now. It just leaves an uneasy feeling as two concurrent GET + PATCH requests can race each other.

therealjeffg:
[...]

Issue #136, opened 2020-10-13
Add file and directory event listener hooks


bgins:
# Summary

It would be nice to be able to add event listeners to files and directories to detect when they have changed. This would permit event-driven programming as an alternative to polling on the filesystem for changes.

## Problem

To detect a change to a file or directory, we have to ask if a change has occurred.

### Impact

Polling has its CPU and memory costs and opens the door for weird timing issues.

## Solution

Add hooks that permit event-driven programming. That way we can do things when and only when a change occurs.

# Detail

**Is your feature request related to a problem? Please describe.**

I have written  a [simple app](https://github.com/bgins/fission-stored-wasm-example) where a user can add two numbers and see the result on another device. A user will see the result when they first open the app on the second device, but it would be nice if the result would also update when two new numbers are added.

The app could periodically check the WinFS file for new results, but updating when an event signals a change would be better.

**Describe the solution you'd like**

Add hooks for event listeners on files and directories.

**Describe alternatives you've considered**

Polling. A hook on the root of the filesystem has also been suggested (`wnfs.publishHooks.push(hook)`), but this calls for filtering for the desired file or directory.

bmann:
Thanks @bgins! There’s a brief sketch in the whitepaper around pieces of this https://whitepaper.fission.codes/file-system/file-system-basics/journal

therealjeffg:
@walkah I want to revisit this in the next sprint - it would be ideal if tooling such as a browser extension could listen to webnative events and react to them.

therealjeffg:
More notes from discussing this with @bgins - 

If I am building an app and want to have my UI update based on state in webnative, i think I might actually have to poll properties currently to get more info on the login state, etc. My use case is I'm building a browser extension that will represent webnative internals in the browser devtools in order to help developers debug their use of webnative. It woudl be great to be able to get events from webnative when changes happen:

 - auth events - logout, etc
 - filesystem events, eg a file is added, deleted, etc
 - what is the ( ipfs ) status of a file, eg get an event when we add it locally, then again when it's been passed into js-ipfs and maybe later when we can confirm it is published and available via a gateway.

bgins:
Adding a few notes as well, the events that I think would be useful from WNFS are around mutations:

- Something changed in the filesystem (any mutation)
- Something changed in a specific directory or a file was changed

[...]

PR #469, opened 2023-01-25
Add event emitter interface and two file system events

icidasset:
__Adds event emitters to the program.__ This allows you to say:

```ts
program.fileSystem.on("local-change", ({ path, root }) => {
  console.log("The file system has changed locally 🔔")
  console.log("Changed path:", path)
  console.log("New data root CID:", root)
})
```

Events added so far
- `fileSystem["local-change"]`
- `fileSystem["published"]`

Events are documented automatically using the types:
Screenshot 2023-01-25 at 18 33 31

## Related issues

#136
#413 
#414 

I'm not sure these issues can be closed since they may involve remote changes, which this PR does **not** include. Server changes are needed in order to achieve this. Personally I'd also like to wait for #163 before we go down that road. In any case, this is a good start and people can set up their own notification system using this.

### DIFF
--- a/package.json
+++ b/package.json
@@ -91,6 +91,7 @@
     "blockstore-datastore-adapter": "^4.0.0",
     "datastore-core": "^8.0.2",
     "datastore-level": "^9.0.4",
+    "emittery": "^1.0.1",
     "events": "^3.3.0",
     "fission-bloom-filters": "1.7.1",
     "ipfs-core-types": "0.13.0",
--- a/package.json
+++ b/package.json
@@ -91,7 +91,6 @@
     "blockstore-datastore-adapter": "^4.0.0",
     "datastore-core": "^8.0.2",
     "datastore-level": "^9.0.4",
-    "emittery": "^1.0.1",
     "events": "^3.3.0",
     "fission-bloom-filters": "1.7.1",
     "ipfs-core-types": "0.13.0",
@@ -127,7 +126,7 @@
     "rimraf": "^3.0.2",
     "ts-node": "^10.9.1",
     "tslib": "^2.4.1",
-    "typedoc": "^0.23.21",
+    "typedoc": "^0.23.24",
     "typedoc-plugin-missing-exports": "^1.0.0",
     "typescript": "^4.8.4",
     "util": "^0.12.4"
--- a/src/events.ts
+++ b/src/events.ts
@@ -1,0 +1,27 @@
+import Emittery from "emittery"
+import { CID } from "./common/cid.js"
+import { DistinctivePath } from "./path/index.js"
+
+
+export type Events = {
+  "fileSystem:local-change": { root: CID, path: DistinctivePath }
+  "fileSystem:published": { root: CID }
+}
+
+
+/**
+ * Events interface.
+ *
+ * Subscribe to events using `on` or `once` (multiple events are allowed),
+ * and unsubscribe using `off`.
+ *
+ * More info on the [emittery Github readme](https://github.com/sindresorhus/emittery/tree/f0b3c2bf8dc985a7dde0e39607e30950394be54b#usage).
+ */
+export type EventEmitter = Emittery<Events>
+
+
+export function createEmitter(): EventEmitter {
+  return new Emittery({
+    debug: { name: "Webnative" }
+  })
+}
--- a/src/events.ts
+++ b/src/events.ts
@@ -15,6 +15,16 @@ export type Events = {
  * Subscribe to events using `on` or `once` (multiple events are allowed),
  * and unsubscribe using `off`.
  *
+ * ```ts
+ * program.events.on("fileSystem:local-change", ({ path, root }) => {
+ *   console.log("The file system has changed locally 🔔")
+ *   console.log("Changed path:", path)
+ *   console.log("New data root CID:", root)
+ * })
+ *
+ * program.events.off("fileSystem:published")
+ * ```
+ *
  * More info on the [emittery Github readme](https://github.com/sindresorhus/emittery/tree/f0b3c2bf8dc985a7dde0e39607e30950394be54b#usage).
  */
 export type EventEmitter = Emittery<Events>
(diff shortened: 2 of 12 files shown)


src/index.ts
```diff
@@ -445,6 +447,9 @@ export async function assemble(config: Configuration, components: Components): P
   // Backwards compatibility (data)
   await ensureBackwardsCompatibility(components, config)
 
+  // Event emitter
+  const eventEmitter = createEmitter()
```
avivash, line 451: I wonder if we should add this is a `component`(dependency) so we don't need to pass it in as a separate arg to each function that uses it? 🤔 though I supposed that may expose it to developers within the context of their client app, which we maybe don't want? Sorry, I'm just thinking out loud 😅

package.json
```diff
@@ -91,6 +91,7 @@
     "blockstore-datastore-adapter": "^4.0.0",
     "datastore-core": "^8.0.2",
     "datastore-level": "^9.0.4",
+    "emittery": "^1.0.1",
```
bgins, line 94: Would it be possible to use our event emitter (https://github.com/fission-codes/webnative/blob/main/src/common/event-emitter.ts) instead of using emittery?

src/events.ts
```diff
@@ -0,0 +1,37 @@
+import Emittery from "emittery"
+import { CID } from "./common/cid.js"
+import { DistinctivePath, Partition, Partitioned } from "./path/index.js"
+
+
+export type Events = {
+  "fileSystem:local-change": { root: CID; path: DistinctivePath<Partitioned<Partition>> }
+  "fileSystem:published": { root: CID }
+}
```
bgins, line 9: We are prefixing "fileSystem" here in case we add other types of event emitters in the future? If so, what would you think about namespacing at the top-level on the program. Something like:

```typescript
program.events.fileSystem.on("local-change", ({ path, root }) => {
  ...
})
```

package.json
icidasset: How did I miss that? 😅 Yeah sure thing! I'll rename `removeListener` to `off` though so that it matches `on`.

src/index.ts
icidasset: Hmm maybe, I don't see this needing to be a component, since event emitting is such a generic thing. We can always revise later, maybe I'll change my mind as we add more events.

src/events.ts
icidasset: Done 👍

package.json
bgins: The name `removeListener` is a better than `off` in my opinion. I've always read the `on` as "on this event, do this thing". The `off` doesn't really make sense in this context, but `removeListener` is explicit about what you are doing.

That said, I'd be alright with either one. We modeled this off of the node event emitter, and they define [emitter.off](https://nodejs.org/api/events.html#emitteroffeventname-listener) as an alias for `removeListener`.

bgins (approved):
Excellent! Super excited to have these first event emitters. 🕺 

Agreed that it doesn't close the existing issues, but it's a great start on them! 💯

package.json
icidasset: I’ll define the ‘on’ and ‘off’ aliases for ‘addListener’ and ‘removeListener’ respectively.

avivash (approved):
Awesome! I can add events to the `recoverFileSystem` method in a follow-up PR. There have been a couple comments saying it would be nice to know what's going on the behind the scenes when the recovery route is `processing`(particularly for the times when it takes a little while) 👌🏼

icidasset:
> Awesome! I can add events to the `recoverFileSystem` method in a follow-up PR. There have been a couple comments saying it would be nice to know what's going on the behind the scenes when the recovery route is `processing`(particularly for the times when it takes a little while) 👌🏼

Sounds good! I was thinking we could do something like:
```ts
fileSystem.on("recovery:progress", progress => {
  switch (progress.stepId) {
    case "REGISTERING_NEW_USER" // This could also be a constant, maybe `progress.steps.REGISTER_NEW_USER`
  }
  
  // Also: stepNumber and stepAmount, or maybe a percentage? 🤷‍♂️
})
```

Also, should we convert the account linking to this system?
I feel like the set up for account linking could be a bit better.
What if we did something like the following instead?
```ts
// on device with existing session
program.accountLinking.listen()

program.accountLinking.on("challenge", (pin) => {
  if (userInput === pin) program.accountLinking.confirmPin()
})

// on device without session
program.accountLinking.link(username)

program.accountLinking.on("challenge", (pin) => {
  showPinInUI(pin)
})
```

src/events.ts
icidasset: Actually now that I think more about adding other events, I'm changing this to:
```ts
program.fileSystem.on("...")
```
Because we already have `program.fileSystem`

avivash:
> > Awesome! I can add events to the `recoverFileSystem` method in a follow-up PR. There have been a couple comments saying it would be nice to know what's going on the behind the scenes when the recovery route is `processing`(particularly for the times when it takes a little while) 👌🏼
> 
> Sounds good! I was thinking we could do something like:
> 
> ```ts
> fileSystem.on("recovery:progress", progress => {
>   switch (progress.stepId) {
>     case "REGISTERING_NEW_USER" // This could also be a constant, maybe `progress.steps.REGISTER_NEW_USER`
>   }
>   
>   // Also: stepNumber and stepAmount, or maybe a percentage? 🤷‍♂️
> })
> ```

I like the idea of having the `currentStep` and `totalSteps` visible 👍🏼 that will open a lot of UI options for the client 👌🏼. Maybe instead of a `switch` we could make it into a map to make it a bit more performant, but i like the general idea! I agree with Brian though, maybe let's move it to its own issue and discuss the implementation details there. I'm happy to open the issue, unless you've already opened one?

icidasset:
@avivash I haven't created any issues for this, go ahead 🙏

avivash:
> @avivash I haven't created any issues for this, go ahead 🙏

Added an issue here: https://github.com/fission-codes/webnative/issues/474. I'll add some implementation proposals in the comments today 👍🏼

PR #473, opened 2023-01-27
Fixes #471

### LINKED ISSUE #471: Issue with history.list() from directory and file
# Summary

## Problem

Cannot list a file or directory history due to the `list` function looking to encode `node.key` as a base64 string as a utf-8 array, but `node.key` is already a utf-8 array. 

### Impact

I believe this makes you entirely unable to list the history for a file or directory - at least in the private directory. 

## Solution

I don't know what the functional intent of having node.key be a Uint8Array, but it seems this would need to be converted to a base64 string - or possibly that is unnecessary and the `fromString` function can be removed from the code.

# Detail

**Describe the bug**

When you call `history.list()` for a file or directory, an error is produced from the file at `src/fs/v1/PrivateHistory` code line `const key = Uint8arrays.fromString(this.node.key, "base64pad");`

This occurs because the node.key is in-fact a Uint8Array. It produces the error `SyntaxError: Non-base64pad character`

**To Reproduce**
`fs.get()` any file or directory and run await `[node].history.list()`

**Expected behavior**
A lovely list of the file/directory history

**Screenshots**

**Smartphone (please complete the following information):**

**Additional context**

icidasset:
Fixes #471 history retrieval.

Didn't catch this in the refactor because I used an `as unknown` before.
I replaced this with better types.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,11 @@
 
 #### Other
 
-- Introduces stricter types for paths to restrict the paths you can use with various functions and also to guide you a bit more.
+Introduces stricter types for paths to restrict the paths you can use with various functions and also to guide you a bit more.
+
+#### Fixes
+
+Fixes various issues with file/directory history.
 
 
 ### v0.35.2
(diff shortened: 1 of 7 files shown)


bgins (approved):
Looks good! Tested this out and it fixes the issue. 💯

PR #476, opened 2023-01-30
Inject debug programs in globalThis context

### LINKED ISSUE #405: Enhance setup.debug to always expose wn object if true
Problem: some developers using webnative ( like me ) will use tooling like sveltekit and initialize our code in a component, effectively hiding it from the DOM and any kind of tooling we or others might want to build. It's currently impossible to access the wn instance from the devtools console, for example, unless you set a breakpoint in the scope where that variable is created.

From a chat w/Brian:

Hey! Question for you. I have some code in my test sveltekit / wn app now to make it easier for me to create a content script:

```javascript
onMount(() => {
  const DEBUG = true; 
  if (DEBUG && DEBUG === true) {
    navigator.__wn = wn;    
  }
});
```

...so I can write code in my content script like:

```javascript
if (navigator.__wn) {
  // this page is definitely using webnative
  navigator.__wn.authenticatedUsername().then(console.log)
}
```

In the codebase we currently have a debug option in setup.ts:

https://github.com/fission-codes/webnative/blob/main/src/setup.ts#L12-L20

When this debug option is set, we could optionally do something like what I have in code example 1 - exposing the wn instance to the top level DOM for convenience.

icidasset:
Closes #405 

```ts
wn.program({
  namespace: "example",
  debug: true
})

const program = globalThis.__webnative.programs["example"]
```

Added a section to the [📜 README](https://github.com/fission-codes/webnative/tree/icidasset/debug-inject#debugging) that explains it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 - Adds `program.fileSystem.recover({ newUsername, oldUsername, readKey })` shorthand method so apps will no longer have to implement the FS recovery flow manually.
 - Adds `program.accountDID(username)` shorthand method to retrieve the root DID of a given account username.
 - Adds the file system data functions as shorthand methods.
+- Injects programs into the global namespace when in debug mode. Can be disabled via the configuration.
 
 #### Other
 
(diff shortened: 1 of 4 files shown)


bgins (approved):
Looks good! 🎉 💯

avivash (approved):
LGTM!

PR #481, opened 2023-02-02
`fs.mv` is a mutation

### LINKED ISSUE #480: mv function not publishing
# Summary

Inability to publish the results of moving a file or directory

## Problem

The `fs.mv` function is returning the proper filesystem result from its promise, but the result cannot be published successfully. 

### Impact

`mv` is rendered ineffective. 

## Solution

No error is produced, so knowing exactly what the a likely fix is would be hard for me to say. Something to do with refactoring the trees for old and new parent directories a file or directory is moved into - I'm not familiar with what exactly this entails. 

# Detail

**Describe the bug**

**To Reproduce**
Please see example project here for reproducible error https://github.com/bgins/webnative-fs-mv-test

**Expected behavior**
It is expected that `fs.mv` and then `publish` would produce updated persistent directory trees for the affected directories. Again, worth noting is that the `mv` method does return the correct filesystem response, but it is reverted when trying to `publish`. 

**Screenshots**

**Smartphone (please complete the following information):**

**Additional context**

icidasset:
Fixes #480

We forgot to mark `mv` as a mutation 🙈

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,7 +22,8 @@
 
 #### Fixes
 
-Fixes various issues with file/directory history.
+- Fixes persistance issue with `fs.mv()`
+- Fixes various issues with file/directory history.
 
 
 ### v0.35.2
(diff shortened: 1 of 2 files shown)


bgins (approved):
Looks good! Tested it out with the reproduction and this fixes it. 💯

PR #487, opened 2023-02-03
Fixes #486

### LINKED ISSUE #486: `loadFileSystem` fails to sync if `cidLog` is non-empty but does not contain `dataCid`
# Summary

## Problem

If `cidLog` is non-empty but does not contain `dataCid`, `loadFileSystem` determines that DNSlink is `-1` updates behind, and skips adding the new CID to the log.

### Impact

Trailing devices will not synchronize with the revision in DNSLink

## Solution

The issue likely stems from this change here: https://github.com/fission-codes/webnative/commit/d988bfa304d4a6f387a3cd3c0e8f8182b5215140#diff-25d1f7f8e769c78a1c45070f524bf6d51a1702a349e2541daf1435c64e6673bdR47

If `cidLog.indexOf(dataCid)` is `-1`, then `logIdx` will be equal to `cidLog.length()`. If `cidLog.length()` is positive, then this triggers the case [here](https://github.com/fission-codes/webnative/blob/main/src/filesystem.ts#L67), preventing synchronization.

Locally, I've fixed this by guarding that case against the absence of `dataCid` in `cidLog`:

```typescript
if (logIdx > 0 && cidLog.indexOf(dataCid) != -1)
```

There's likely a solution that makes these cases clearer, though.

icidasset:
Testing the various states:
- **No data cid**: Create new account.
- **Up to date**: Refresh the page after a publish has successfully completed.
- **Outdated**: Refresh the page just after making a change in the file system and just before the publish has finished. That way the DNS becomes out of sync with your local state. Refreshing the page should say "outdated, 1 newer local entry". Repeating this process should increase this counter (ie. 2 newer local entries, and so on).
- **Newer**: Make a change in one session, wait for publish to finish and open app in other session.

Fixes #486 

You can test this with the alpha version: `0.36.1-alpha-1`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+### v0.36.1
+
+- Fixes issue with the CID log which in certain scenarios caused Webnative to load the wrong version of an account's file system.
+
 ### v0.36.0
 
 #### Breaking changes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ### v0.36.1
 
-Fixes issue with the CID log which in certain scenarios caused Webnative to load the wrong version of an account's file system.
+Fixes an issue with the CID log, which, in certain scenarios, caused Webnative to load the wrong version of an account's file system.
 
 ### v0.36.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ### v0.36.1
 
-- Fixes issue with the CID log which in certain scenarios caused Webnative to load the wrong version of an account's file system.
+Fixes issue with the CID log which in certain scenarios caused Webnative to load the wrong version of an account's file system.
 
 ### v0.36.0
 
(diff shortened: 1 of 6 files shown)


avivash:
Tested all those scenarios locally and it looks good 👍🏼 One thing I noticed though was if I refreshed before a `publish` completed  and saw the `newer local entry` log 👇🏼 
image
then opened up a new session, I wouldn't see the local entry in that gallery, but I think that is probably expected behaviour, right?

**Original Session:**
image

**New Session:**
image

icidasset:
> but I think that is probably expected behaviour, right?

Yeah exactly @avivash

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+### v0.36.1
+
+Fixes issue with the CID log which in certain scenarios caused Webnative to load the wrong version of an account's file system.
```
avivash, line 5: Small grammar nitpick, but entirely necessary maybe since it's just the changelog
```suggestion
Fixes an issue with the CID log, which, in certain scenarios, caused Webnative to load the wrong version of an account's file system.
```

avivash (approved):
LGTM! 🚀

PR #490, opened 2023-02-06
Allow symlinks to other file systems to be created

### LINKED ISSUE #489: Soft links to other users' trees cannot be created with `fs.symlink`
# Summary

## Problem

The [docs](https://guide.fission.codes/developers/webnative/sharing-private-data#soft-links) suggest that soft links pointed at other users' trees can be created, however the [function itself](https://github.com/fission-codes/webnative/blob/main/src/fs/filesystem.ts#L457) does not take `username` as an argument, and instead just uses the current user.

## Solution

Support passing in an optional `username`, defaulting to the current user. I've hacked up roughly this locally, and it seems to work as intended.

icidasset:
Closes #489

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+### v0.36.2
+
+Allow symlinks to other file systems to be created.
+
 ### v0.36.1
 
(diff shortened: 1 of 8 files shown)


bgins (approved):
Looks good! 🎉

avivash (approved):
LGTM! 🚀

PR #495, opened 2023-02-14
Fix & re-enable rs-wnfs integration tests

### LINKED ISSUE #429: Re-enable WASM tests
WASM tests are currently skipped because of a garbage-collection issue.

matheus23:
Fixes #429 

I added the "no changelog" label since this is only a fix for a feature behind an experimental feature flag. I don't think this needs a changelog entry. Right?

### DIFF
--- a/src/fs/v3/PublicRootWasm.ts
+++ b/src/fs/v3/PublicRootWasm.ts
@@ -13,10 +13,15 @@ import { BaseFile } from "../base/file.js"
 import { Metadata } from "../metadata.js"
 
 
+// This is some global mutable state to work around global mutable state
+// issues with wasm-bindgen. It's important we *never* accidentally initialize the
+// "wnfs" Wasm module twice.
 let initialized = false
 
 async function loadWasm({ manners }: Dependencies) {
-  // MUST NOT initialize twice
+  // MUST be prevented from initializing twice:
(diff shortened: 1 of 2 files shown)


matheus23:
I wrote more about the issue itself here:

https://github.com/rustwasm/wasm-bindgen/issues/3307

matheus23:
So apparently, the original code already had that `initialized` variable which prevents the bug:
https://github.com/fission-codes/webnative/blob/2d0a93e4f435a51a7618ce47544fb17b391352ad/src/fs/v3/PublicRootWasm.ts#L19-L22

But then it got reverted in https://github.com/fission-codes/webnative/pull/422/files?diff=unified&w=0#diff-4952ebe25ed816f4c8f9f5d03618522f41ab5e00941e4a69f46990b1f9755094L19-R28

This is just reverting the revert.

I realize the refactor's goal when it removed the `initialized` variable tried to get rid of global mutable state in general. Unfortunately, in this case we can't get rid of it, as we need to be careful about correctly maintaining `wasm-bindgen`-generated code (which contains global mutable state), and which is somewhat out of our control. The best we can do is file an issue (see the previous message).

So, I added more comments signalling that this code must not be deleted.

icidasset (approved):
Thanks! 🙏

Issue #459, opened 2023-01-06
`v0.35` issue with `fs.write` — "Could not decode CID"


nichoth:
# Summary

The only change was updating to `0.35`, was working previously. Can see the code here: https://github.com/snail-situation/hermes/blob/14f8e5178a0e4e4153177edb43b6ed9feefe3d9e/src/pages/whoami.tsx#L78

## Problem

```js
        try {
            const filepath = wn.path.appData(
                PERMISSIONS.app,
                wn.path.file(CONSTANTS.avatarPath)
            )
            await fs.write(filepath, pendingImage.file)
            console.log('file path written...', filepath)

            // read the file we just wrote
            const content = await fs.cat(filepath)
            appAvatar.value = URL.createObjectURL(
                new Blob([content as BlobPart], { type: 'image/jpeg' })
            )

            await fs.publish()

            setPendingImage(null)
        } catch (err) {
            console.log('could not write file...')
            console.log('errrrrrrrrrrr', err)
        }
```

this is console log:
```
could not write file...
whoami.tsx:68 errrrrrrrrrrr Error: Could not decode CID: {"/":"bafkreicu646jao2xjpkbmk3buom6hmxsexmbwyju22k6wmtnky2ljisv3e"}
    at decodeCID (cid.ts:47:9)
    at PrivateFile.fromInfo (PrivateFile.ts:145:7)
    at PrivateFile.fromLatestName (PrivateFile.ts:117:24)
    at async PrivateTree.getDirectChild (PrivateTree.ts:254:11)
    at async PrivateTree.createOrUpdateChildFile (PrivateTree.ts:191:22)
    at async PrivateTree.addRecurse (tree.ts:82:7)
    at async PrivateTree.add (tree.ts:70:5)
    at async Object.private (filesystem.ts:323:13)
    at async FileSystem.runMutationOnNode (filesystem.ts:762:7)
    at async FileSystem.add (filesystem.ts:312:7)
```

nichoth:
Reproduced in another place — https://github.com/snail-situation/hermes/blob/v0.35/src/pages/whoami.tsx#L114

icidasset:
Thanks @nichoth ! Is this with a file system that existed before and was used with an older version of Webnative? Will look into this tomorrow.

nichoth:
> Thanks @nichoth ! Is this with a file system that existed before and was used with an older version of Webnative? Will look into this tomorrow.

[...]

PR #496, opened 2023-02-15
Parse DAG-JSON CIDs

icidasset:
Should fix issues such as #459 where CIDs in the format of `{"/":"CID_STRING"}` (DAG-JSON) are encountered.


bgins (approved):
Looks good! 👌

PR #504, opened 2023-03-27
Add browser extension support

### LINKED ISSUE #460: Add browser extension support
# Summary

## Problem

Webnative does not expose information in a way that is accessible to browser extensions.

### Impact

Browser extensions cannot easily interface with apps that use Webnative.

## Solution

Implement an API that browser extensions can interface with to query Webnative for:

- Webnative version
- App namespace
- Username
- Account DID
- Agent DID
- Filesystem skeleton
- Data root CID
- Capabilities (what parts of WNFS does the app have access to?)
- Bitswap activity and connection details (Maybe not? We may be abandoning bitswap soon.)

The information should only be exposed when the developer has set `debug` to `true` when initializing a Webnative program.

Webnative should send messages to a browser extension when any of the above values change. We want to avoid polling for changes.

bgins:
## Summary

* [x] Adds browser extension support
  * [x] Add `config.debugging.emitWindowPostMessages` flag to gate window post messages intended for the extension
* [x] Move events to `program.on` and namespace them with string prefixes 
* [x] Adds session event listeners
* [x] Rename `window.__webnative` to `window.__odd`

### Browser extension support

This PR adds support for connecting and sending messages to the [Webnative devtools extension](https://github.com/webnative-examples/webnative-devtools/tree/release-prep). Messages are sent as window post messages when Webnative debug mode is set to `true`. Messages are tagged with the browser extension ID or as `odd-devtools-ready-message` to inform the extension that Webnative is ready to connect.

Window post messages can be completely disabled by setting `config.debugging.emitWindowPostMessages` to false.

### Move events onto program

This PR moves events onto the top-level program and renames them. For example:

```typescript
program.fileSystem.on("local-change", ({ path, root }) => { ... })
```

```typescript
program.on("fileSystem:local-change", ({ path, root }) => { ... })
```

### Session events

Two new events have been added:

```typescript
export type Session<S> = {
  "session:create": { session: S }
  "session:destroy": { username: string }
}
```

These events are emitted when a session is created or destroyed. The `S` generic is filled in with a `Session` in the implementation.

### Let the renaming begin

This PR accounts for renaming Webnative to ODD at points of interaction with the extension, but does not rename everything in the code base.

## Test plan (required)

Install the extension locally as described in the Webnative devtools extension repo. Use the `release-prep` branch.

Run an app with Webnative debug mode set to `true`. Open the devtools and find the Webnative panel. The panel should show Webnative connected if the extension has successfully connected with Webnative.

See the Test Plan in this PR for a rundown of the functionality provided by the extension: https://github.com/webnative-examples/webnative-devtools/pull/21.

Check that event emitters are properly exposed at `program.on`.

## Closing issues

Closes #460 

## After Merge
* [ ] Does this change invalidate any docs or tutorials? Yes, the guide will need some updates
* [ ] Does this change require a release to be made? We will hold off on release until we have renamed everything internally to ODD and are ready to publish under the new name to npm


avivash (approved):
Looks awesome, this is going to be so useful for devs! 🚀

src/components/auth/implementation/base.ts
```diff
@@ -108,16 +109,23 @@ export async function register(
...
+    const session = new Session({
       crypto: components.crypto,
       storage: components.storage,
+      eventEmitter: eventEmitters.session,
       type: TYPE,
       username: authedUsername
     })
 
+    // Emit session create event
+    eventEmitters.session.emit("session:create", { session })
```
icidasset, line 125: Why not call this in the Session constructor so that you only have to `emit` the event in one place?

src/configuration.ts
```diff
@@ -20,6 +20,13 @@ export type Configuration = {
    * Debugging settings.
    */
   debugging?: {
+     /**
+     * Should I emit window post messages with session and filesystem information?
+     *
+     * @default true
+     */
+    emitWindowPostMessages?: boolean
```
icidasset, line 28: 👍

src/index.ts
```diff
@@ -619,6 +620,7 @@ export async function assemble(config: Configuration, components: Components): P
   // Create `Program`
   const program = {
     ...shorthands,
+    ...{ ...Events.listenTo(fsEvents), ...Events.listenTo(sessionEvents)},
```
icidasset, line 623: Doesn't this override the file system events with the session events?
I think it might be better to just create one overarching type for the events and listen to that?
```ts
// events.ts
type All = Events.FileSystem & Events.Session<Session>

// index.ts
const events = Events.createEmitter<Events.All>()
...Events.listenTo(events)
```
Also means you have to pass less instances around which is nice.

That said, I'm not sure if you can use a subset as a param with this 🤔 
For example, if you create a Session and pass in the "all" emitter,
would you be able to use the following as the type of the constructor param?
```ts
Events.Emitter<Events.Session<Session>>
```
I would think that would work, because it is a subset of the `All` type and you only use the session events. Thoughts?

src/index.ts
```diff
@@ -635,9 +637,38 @@ export async function assemble(config: Configuration, components: Components): P
 
     if (inject) {
       const container = globalThis as any
-      container.__webnative = container.__webnative || {}
-      container.__webnative.programs = container.__webnative.programs || {}
-      container.__webnative.programs[ namespace(config) ] = program
+      container.__odd = container.__odd || {}
+      container.__odd.programs = container.__odd.programs || {}
+      container.__odd.programs[ namespace(config) ] = program
```
icidasset, line 642: 👀

src/extension/index.ts
```diff
@@ -0,0 +1,239 @@
...
+
+// CREATE
+
+
+type Config = {
+  namespace: AppInfo | string
+  session: Maybe<Session>
+  capabilities?: Permissions
+  shorthands: ShortHands
+  lookupDataRoot: (username: string) => Promise<CID | null>
```
icidasset, line 21: Can we try to not use the shorthands from `index.ts` here?
You can use:
```ts
import { agent as agentDID } from "../did/index.js"
// Or:
import * as DID from "../did/index.js"
DID.agent(cryptoComponent)
```
In other places in the code we pass in the components entirely.
So I'd replace `shortHands` and `lookupDataRoot` with:
```ts
import { Crypto, Reference } from "../components.js"

// Following the pattern of other files
dependencies: { crypto: Crypto.Implementation; reference: Reference.Implementation }
```

src/extension/index.ts
```diff
@@ -0,0 +1,239 @@
...
+}
+
+function listen(connection: Connection, config: Config): Listeners {
+  async function handleLocalChange(params: { root: CID; path: DistinctivePath<[ Partition, ...string[] ]> }) {
+    const { root, path } = params
+    const state = await getState(config)
+
+    globalThis.postMessage({
+      id: connection.extensionId,
+      type: "filesystem",
```
icidasset, line 105: Maybe `fileSystem` with a capital S to be consistent?

src/extension/index.ts
```diff
@@ -0,0 +1,239 @@
...
+      timestamp: Date.now(),
+      state,
+      detail: {
+        type: "publish",
+        root: root.toString()
+      }
+    })
```
icidasset, line 129: Nitpick. Looks like you can abstract a bit of this into a separate function so you don't need to repeat it every time? So it ends up being for example:
```ts
const { root } = params

postMessageHelper({
  type: "fileSystem",
  detail: { ... }
})
```

src/extension/index.ts
```diff
@@ -0,0 +1,239 @@
...
+
+// STATE
+
+
+type State = {
+  app: {
+    namespace: AppInfo | string
+    capabilities?: Permissions
+  }
+  filesystem: {
```
icidasset, line 195: Same here, `fileSystem`

icidasset (commented):
Some concerns and nitpicks, but overall looking great!
Yay for a great dev experience! 🎉 Next step:

src/extension/index.ts
icidasset: Feel free to ignore this one if you disagree.

src/extension/index.ts
bgins: Yep, good call. Updated that. 👌

src/extension/index.ts
bgins: Yep, for sure! Updated to pass in the components directly instead of relying on shorthands and the `lookupDataRoot` function.

src/components/auth/implementation/base.ts
bgins: Yeah, that's a great idea. Much simpler! 💯 

Updated to move it into the session constructor.

src/index.ts
bgins: > Doesn't this override the file system events with the session events?

Yes, it does. Good catch. 🙏 

> I think it might be better to just create one overarching type for the events and listen to that?
> 
> ```ts
> // events.ts
> type All = Events.FileSystem & Events.Session<Session>
> 
> // index.ts
> const events = Events.createEmitter<Events.All>()
> ...Events.listenTo(events)
> ```

We can't use the `Session` type in `events.ts` because that would create a circular dependency. What we could do is:

> 
> ```ts
> // events.ts
> export type All<S> = FileSystem & Session<S>
> 
> // index.ts
> const events = Events.createEmitter<Events.All<Session>>()
> ...Events.listenTo(events)
> ```

---

> That said, I'm not sure if you can use a subset as a param with this 🤔 For example, if you create a Session and pass in the "all" emitter, would you be able to use the following as the type of the constructor param?
> 
> ```ts
> Events.Emitter<Events.Session<Session>>
> ```
> 
> I would think that would work, because it is a subset of the `All` type and you only use the session events. Thoughts?

Unfortunately, this does not appear to work out. Not entirely sure why, perhaps the rules are different within a generic?

One option is to pass around `all` events everywhere, but we'd be losing some type information on which events different parts of the code emit. I suspect that we will have to pass `all` around anyways because that is what will be listened upon. But perhaps we can do something with `extends` or something to document which parts of `all` are used. Or maybe there is a way to construct `all` from `fileSystem` and `session` emitters somehow. 🤔

icidasset: What if you keep the individual emitters, but then write a function to merge them together to use with `Events.listen()`?
```ts
merge(a: EventEmitter<A>, b: EventEmitter<B>): EventEmitter<A & B> 
```
Just realised I repeated what you said 🙈

bgins: Started experimenting with this a bit, but it seems like it pushes the typing issues into the merge function. Here's what I have so far:

```typescript
export function merge<A, B>(a: EventEmitter<A>, b: EventEmitter<B>): EventEmitter<A & B>  {
  const merged = createEmitter<A & B>()
  const aEmit = a.emit
  const bEmit = b.emit

  a.emit = (...args) => {
    aEmit.call(a, ...args)
    merged.emit(...args)
  }

  b.emit = (...args) => {
    bEmit.call(b, ...args)
    merged.emit(...args)
  }

  return merged
}
```

The idea is to rewire each event emitter's `emit` function so that it also calls `merged.emit`. But we end up with type errors on `merged.emit` like:

```
(parameter) args: [eventName: K, event: A[K]]
Argument of type 'A[K]' is not assignable to parameter of type '(A & B)[K]'.
  Type 'A' is not assignable to type 'A & B'.
    Type 'A' is not assignable to type 'B'.
      'B' could be instantiated with an arbitrary type which could be unrelated to 'A'.
```

icidasset:
```ts
  a.emit = <K extends keyof A>(eventName: K, event: (A & B)[ K ]) => {
    aEmit.call(a, eventName, event)
    merged.emit(eventName, event)
  }

  b.emit = <K extends keyof B>(eventName: K, event: (A & B)[ K ]) => {
    bEmit.call(b, eventName, event)
    merged.emit(eventName, event)
  }
  ```
  This seems to work?

bgins: Yesss! That did it! 🙌 

Pushed the changes up. Thanks for the help thinking through this. 🙏

icidasset (approved):

src/extension/index.ts
bgins: Holding off on this for now, but it does seem like we will want this when we start adding more messages. 👌

PR #506, opened 2023-04-03
Rename webnative to odd

### LINKED ISSUE #505: Ensure the global we use in the CDN package is oddjs
# Summary

From this page:

https://guide.fission.codes/developers/webnative/installation#cdn

Currently the CDN file adds the 'webnative' global:

```javascript
const wn = self.webnative
```

We need to ensure that the published ODD SDK CDN script supports the 'oddjs' name instead:

```javascript
const odd = self.oddjs
```

### LINKED ISSUE #507: Rename Webnative to ODD
# Summary

## Problem

Webnative is named incorrectly. It should get ODD.

### Impact

Folks won't know about the rename if we don't make it so.

## Solution

Rename from Webnative to ODD in all the places.

bgins:
## Summary

This PR renames Webnative to ODD with the following changes:

* [x] Rename package from `webnative` to `@oddjs/odd`
* [x] Update GitHub repo references to https://github.com/oddsdk/ts-odd
* [x] Update `webnative-walletauth` to `odd-walletauth` at https://github.com/oddsdk/odd-walletauth
* [x] Update `webnative-app-template` to `odd-app-template` at https://github.com/oddsdk/odd-app-template
* [x] Update website reference to https://odd.dev
* [x] Update guide links to point to https://docs.odd.dev
* [x] Update API docs URL to https://api.odd.dev
* [x] Rename Webnative to ODD in README, comments, and error messages
* [x] Update README references to `globalThis.__webnative.programs` to `globalThis.__odd.programs`
* [x] Expose ESM as `oddjs` instead of `webnative`
* [x] Update nix flake name from `webnative` to `oddjs`
* [x] Rename integration test CAR files
* [x] Rename typedoc name from `Webnative SDK` to `ODD SDK`
* [x] Rename iframe and classified post message to odd
* [x] Set `sharedRepo` to `false`
* [x] Update docs app_url to `odd.fission.app` 

~~Note that the docs are still published to `webnative.fission.app` and we will point https://api.odd.dev to that for the moment.~~ The docs have been updated to publish to `odd.fission.app`.

### What's missing?

WNFS has not been renamed yet. Can we and should we reframe this to use the term ODD Volumes?

~~There are a couple of places where there are breaking changes in the interaction with the lobby and the names of IndexedDB keys~~. These have been addressed based on PR feedback below:

https://github.com/fission-codes/webnative/blob/27842451b01510a46524b35d997627c3fdc844c9/src/index.ts#L865-L867

Left this as is because it is for backwards compatibility.

https://github.com/fission-codes/webnative/blob/27842451b01510a46524b35d997627c3fdc844c9/src/components/capabilities/implementation/fission-lobby.ts#L97

Set `sharedRepo` to `false`.

https://github.com/fission-codes/webnative/blob/27842451b01510a46524b35d997627c3fdc844c9/src/components/capabilities/implementation/fission-lobby.ts#L157

https://github.com/fission-codes/webnative/blob/27842451b01510a46524b35d997627c3fdc844c9/src/components/capabilities/implementation/fission-lobby.ts#L204-L207

These last two have been renamed from `webnative` to `odd`.

## Test plan (required)

The usual test suites should run as before because this is mostly about renaming.

## Closing issues

Closes #505
Closes #507

## After Merge
* [ ] Does this change invalidate any docs or tutorials? Yesss, so much. We will.
* [ ] Does this change require a release to be made? It will. 👍


icidasset:
> (this could be changed! Maybe https://api.odd.dev/?)

Yes please 👍 

> couple of places where there are breaking changes

1. This only applies to versions of Webnative before `v0.35`
    It's an upgrade step from `< 0.35` to `>= 0.35`, you can leave this in or remove it entirely.
2. This one can be removed, so set it to `sharedRepo = false`
3. This `id` doesn't matter I think, feel free to rename.
4. This one doesn't matter either, only `didExchange` is of importance, feel free to rename.

README.md
```diff
@@ -1,56 +1,56 @@
...
   // If you're developing multiple apps on the same localhost port,
   // make sure these differ.
   namespace: { creator: "Nullsoft", name: "Winamp" }
 
 }).catch(error => {
   switch (error) {
-    case webnative.ProgramError.InsecureContext:
-      // Webnative requires HTTPS
+    case odd.ProgramError.InsecureContext:
+      // ODD requires HTTPS
```
icidasset, line 44: When 'ODD' is used on its own (eg. not ODD Program), should we say ODD.js or ODD SDK instead?

bgins: Yeah, good point. I think the distinction here is ODD SDK when used as a noun, and ODD when used as an adjective. I'll update the noun spots. 👌

avivash: nice, I've been using `ODD SDK` for those cases in the new odd.dev site too 👍🏼

bgins:
> Just flagging that we'll want to change the workflow urls too once those domains are ready https://github.com/fission-codes/webnative/blob/rename-webnative-to-odd/.github/workflows/docs.yaml#L33

Yep, good call. I'm not entirely sure how we have this set up to publish. 

@walkah could we look into updating this workflow URL to `odd.fission.app`? Not a high priority because it will be hidden behind https://api.odd.dev.

icidasset (approved):

avivash (approved):
Looks great! I'm going to start integrating this branch with a few of the `webnative`->`odd` branches i have going in the different template repos 👍🏼

bgins:
> Just flagging that we'll want to change the workflow urls too once those domains are ready https://github.com/fission-codes/webnative/blob/rename-webnative-to-odd/.github/workflows/docs.yaml#L33

Updated this to `odd.fission.app`.

PR #525, opened 2023-06-12
Fix issue with capabilities component

icidasset:
Turns out it didn't work after all when changing the message passed to `postMessage` 🙈 Also got rid of the `via-postmessage` check which added support for the old lobby which has long been deprecated for a while, and improved the retry function which wasn't working well (actually ran the action 20 times even though initial one succeeded).

### DIFF
--- a/src/components/capabilities/implementation/fission-lobby.ts
+++ b/src/components/capabilities/implementation/fission-lobby.ts
@@ -281,20 +281,23 @@ async function retry<T>(
   action: () => Promise<T>,
   options: { tries: number; timeout: number; timeoutMessage: string }
 ): Promise<T> {
-  return await Promise.race([
-    action(),
-    new Promise<T>((resolve, reject) => {
-      if (options.tries > 0) return setTimeout(
-        () => retry(action, { ...options, tries: options.tries - 1 }).then(resolve, reject),
-        options.timeout
-      )
+  return new Promise((resolve, reject) => {
+    if (options.tries > 0) {
+      const unoMas = () => {
+        retry(action, { ...options, tries: options.tries - 1 })
+      }
 
-      return setTimeout(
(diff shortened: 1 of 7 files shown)


src/components/capabilities/implementation/fission-lobby.ts
```diff
@@ -99,26 +85,26 @@ export async function request(
...
+    ["redirectTo", redirectTo],
+    ["sdk", VERSION.toString()],
+    ["sharedRepo", sharedRepo ? "t" : "f"],
+    ["sharing", sharing ? "t" : "f"]
 
   ].concat(
-    app ? [ [ "appFolder", `${app.creator}/${app.name}` ] ] : [],
-    fs?.private ? fs.private.map(p => [ "privatePath", Path.toPosix(p, { absolute: true }) ]) : [],
-    fs?.public ? fs.public.map(p => [ "publicPath", Path.toPosix(p, { absolute: true }) ]) : [],
-    raw ? [ [ "raw", Base64.urlEncode(JSON.stringify(raw)) ] ] : [],
```
icidasset, line 113: Really need to add a commit hook with some default formatting 😅 I switched editors.

src/components/capabilities/implementation/fission-lobby.ts
```diff
@@ -295,20 +281,23 @@ async function retry<T>(
...
-
-      return setTimeout(
-        () => reject(new Error(options.timeoutMessage)),
-        options.timeout
-      )
-    })
-  ])
+  return new Promise((resolve, reject) => {
+    if (options.tries > 0) {
+      const unoMas = () => {
```
bgins, line 286: Claro que si. Uno mas esta bien. 😄

bgins (approved):
Nice! Looks good. 🎉 

Left one comment about naming where it looks like we went back to "webnative".

PR #532, opened 2023-07-05
[NEXT] Adapt to UCAN 0.9

icidasset:
This PR aims to bring the UCAN 0.9 to the SDK. The prototype version of UCAN we currently have is mixed into a lot aspects of the SDK. Because of this I have disabled and removed a lot functionality, such as most importantly, device linking, but more on that later.

After merging this, the `next` branch will be able to make somewhat functioning apps again. You can't make a Fission account yet though. You can load the file system without having an account and play around with it, like so:
```ts

const program = await odd.program({
  namespace: "odd-test"
})

const fs = await program.fileSystem.load()
```

### UCAN

Ideally we'd upgrade directly to UCAN v0.10. But we can't do that just yet (waiting on libs to upgrade), so I opted for 0.9 for now. That said, it should be quite straightforward to upgrade to the next version.

### File System

The `next` branch is already using rs-wnfs primarily, but you can't make an app with it yet. This PR builds further on that and connects various pieces. Most importantly here is the private node mounting system. Which is where you take a [private ref](https://docs.rs/wnfs/0.1.21/wnfs/private/struct.PrivateRef.html) to load a private node using rs-wnfs. These secrets need to be passed from device to device in order to gain access to various parts of the private file system.

### Linking

The SDK also has this delegation system (auth lobby). We've gradually abstracted this lobby system more and more, and this PR takes another step further. Because device linking and app-to-app linking is so similar, we've decided to merge the APIs and the channel over which the UCANs and secrets are transferred.

I've completed the initial API surface which you can see in various places, but it's not finished just yet. The idea is to do this:

```ts
// authority
program.authority.provide()
program.authority.on("query", ...)

// delegate
program.authority.request({
  from: ...,
  query: [ odd.authority.fileSystem("read", path) ]
})
program.authority.on("challenge", ...)
```
This'll be implemented using two components: authority & channel
Other things to check out: `src/authority/query.ts`

### Accounts & Identifiers

The account system and identity layer have also been abstracted. We currently have usernames mixed in everywhere, this PR gets rid of that. Also very important, @hugomrdias had the great idea to always delegate from the identifier to the agent, which this PR implements. I'm using the naming "identifier" here instead of "identity" because that's @bmann his preferred naming which I agree with. Humans have multiple identifiers, not identities.

This makes implementing various identifier systems much easier. There will always be a delegation from the identifier to agent, and the agent delegates to the account system. Account and identifier systems can be implemented using their relevant components (account & identifier). The first account system to be implemented will be the new Fission server.

Registering an account has changed a bit because of the abstraction:
```ts
// Instead of passing the username directly, this now depends on the component implementation.
// But it'll always be an object you pass in.
const formValues = { username: "example", email: "..." }
const { ok } = await program.account.canRegister(formValues)
const { ok } = await program.account.register(formValues)
```
If either `ok` is false, there will be `reason` attached to it so you know why it failed.

You can also add additional methods to the implementation that should be made available on the `program.account` object. This is called the `annex`. So for example, the fission account implementation will most likely have a method to request a code via email which is required for the registration.

### Depot

I've added a `flush` method to the `Depot` component. This is part of separating the data from the account layer. This should make it possible to for example use Web3Storage as the file system block storage (or in addition to Fission's storage).

The way this works is, you have two components that are called when a mutation is made in the file system (after a short debounce). You have the account system which is responsible for the data root and the depot, which is responsible for the IPLD block storage. Whenever the data root is going to be updated, `flush` is called. So you could use this to boot up your IPFS node in the browser, setting up a bitswap connection. Or to gather the changed IPLD blocks, package them up in a CAR file and push them to Web3Storage.

### Cabinet

I've renamed the UCAN repository to the "cabinet". That is where your "ucan tickets" and access keys (file system secrets) are stored. The idea here is that this would be the place where you go look to check if you have access to a particular resource, be it related to your account or your file system.

## New flow

With all these changes, the flow to work with a program has changed to:
```ts
// Root device, where you register your account
const outcome = await program.authority.isGranted()

if (outcome.granted) {
  // Already registered
  const fs = await program.fileSystem.load()

} else {
  // No access, could log this to find out what is missing.
  console.log(outcome.reason)

  // But for now let's assume we haven't registered yet.
  await program.account.register()

}
```
From here you'd link a device or provide partial access to another entity (as mentioned above).

## Smaller Changes

- Replaces user session objects with `program.access.isGranted()`
- Removed the DID code, uses @hugomrdias his libraries
- Removed keystore-idb
- Removed DIDs from the Program object, these don't really make sense anymore. Agent DID should never be used anymore by the developer, always use identifier DID. Sharing/exchange DID needs to be rethought, will be implemented later together with private-data sharing.
- Removed the crypto component, doesn't make sense anymore to have this as a component. Every JS language supports the web crypto API now anyway. Less references to pass around. Agent keys can still be customised using the agent component.
- Reference component has been removed, functionality has been moved to other components or into their own component.
- Username availability check is now done through DNS to avoid fetch errors in browser.
- Cleans up various unused `common` code.
- Account UCAN is no longer stored separately
- Storage locations have changed so it will not cause a conflict with older versions of the SDK
- Simplified the debugging config, `debug` and `debugging` config params have been merged.
- Backwards compatibility with older Webnative clients has been removed.
- Added pre-commit hook that formats all the source files, markdown and json.
- File system is not loaded automatically anymore.

### In this PR

- [x] Finish up `login` behaviour/implementation
- [x] Update comments in `index`
- [x] Fix tests
- [x] Fix recursive imports

### In other PRs

Things that are missing, need to re-implemented or still need to be fixed:

- Update rs-wnfs
- Integrate rs-ucan
- Use iso-signatures to sign data instead of using custom code
- Reduce bundle size (see [Discord discussion](https://discord.com/channels/478735028319158273/646375854628405273/1131534022514839644))
- Linking (access component implementation)
- Extensions
- Private data sharing
- File system versioning
- File system recovery
- Unix FS tree (pretty tree)


bgins:
Looking the linking API:

```typescript
// authority
const program = await odd.program(config, "authority")

program.access.provide()
program.access.on("query", ...)

// delegate
const program = await odd.program(config, "delegate", {
  request: [ odd.access.fileSystem("read", path) ]
})

program.access.request()
program.access.on("challenge", ...)
```

Love how this is coming along, "request... access" reads great to me. ❤️ 

One question here is whether or not this supports a `delegate` requesting access from another `delegate`. In the past, @expede has mentioned that we would like to support multiple levels of subdelegation.

Would that work with this API? Does putting the request into the program initialization make it difficult to declare who we are requesting access from?

bgins:
> This makes implementing various identifier systems much easier. There will always be a delegation from the identifier to agent, and the agent delegates to the account system. Account and identifier systems can be implemented using their relevant components (account & identifier). The first account system to be implemented will be the new Fission server.

This hierarchy sounds great for supporting interoperability. 💯 

> Replaces user session objects with program.isConnected()

Nice! Like the "connected" naming.

> Username availability check is now done through DNS to avoid fetch errors in browser.

Is there a risk for race conditions here? Say a user registers a name, and a second user comes along and wants the same name. If we are waiting for DNS to propagate, could the username check fail? This might be alright, as I imagine registration itself would fail.

> Backwards compatibility with older Webnative clients has been removed

Yes, bon voyage 👋 . When we have a rough timeline for next, we should announce this in Discord to give folks a heads up.

icidasset:
> One question here is whether or not this supports a delegate requesting access from another delegate. In the past, @expede has mentioned that we would like to support multiple levels of subdelegation.

@bgins Yeah good point. I don't know if you saw the recent discussion about that on Discord. But I'd like some well-defined use cases for this, otherwise I don't know really how to shape the API. But maybe we opt to make it generic instead?

I'd like to avoid roping the initialisation into this, but not sure how. With these changes when you run your program in `authority` mode, you get a `register` and `login` function on your program. Because, based on the lobby model, you would only register an account or do device linking in the lobby (the authority), not the "delegate". Then there's also the question of how `isConnected()` should function, it'll need different functionality based on the mode it's operating in (authority or delegate). Right?

There's something here about specific use cases. We assume there's an "authority" which requires full read & write access to the file system and a "delegate" which could operate with a very small subset of that (eg. only read portion of file system). But is that always the case? 🤷‍♂️ 

Maybe we could do something like this instead?
```ts
const program = await odd.program(config)

// authority
await program.account.register()
await program.account.login()

await program.account.isConnected()
// NOTE: Could check if UCAN has been delegated with full write access and if we can update the data root.
//             Also should potentially check if the configured account system has all the capabilities it needs.

// delegate
await program.access.request({
  from: ...,
  query: [ ... ]
})

await program.access.isGranted(query)
```
Thoughts?

> Is there a risk for race conditions here? Say a user registers a name, and a second user comes along and wants the same name. If we are waiting for DNS to propagate, could the username check fail? This might be alright, as I imagine registration itself would fail.

Blaine is setting up some DNS magic which will make it look up the values in the database directly, so propagation will be faster I assume. That said, I think it'll be fine too, the registration itself would indeed fail.

bgins:
> But I'd like some well-defined use cases for this, otherwise I don't know really how to shape the API. But maybe we opt to make it generic instead?

Yeah, it would be nice to have use cases. Not sure what they are, and I can see how this makes it more challenging to design an API.

> I'd like to avoid roping the initialisation into this, but not sure how. With these changes when you run your program in authority mode, you get a register and login function on your program. Because, based on the lobby model, you would only register an account or do device linking in the lobby (the authority), not the "delegate". 

I'm agreed that there is a fundamental difference between an authority and a delegate, and I think it's a good design to only expose what one or the other needs.

> Then there's also the question of how `isConnected()` should function, it'll need different functionality based on the mode it's operating in (authority or delegate). Right?

Are there cases where a delegate could start with a local-only file system that later gets merged into a file system granted by an authority? If so, I think `isConnected()` works for both. I think this works for offline use cases, where a delegate app may want to store some data then later request access to persist and sync it.

The way these would function differently is the method of requesting authority. The authority gets it by registering an account, and the delegate gets it from the authority.

---

```typescript
const program = await odd.program(config)

// authority
await program.account.register()
await program.account.login()

await program.account.isConnected()
// NOTE: Could check if UCAN has been delegated with full write access and if we can update the data root.
//             Also should potentially check if the configured account system has all the capabilities it needs.

// delegate
await program.access.request({
  from: ...,
  query: [ ... ]
})

await program.access.isGranted(query)
```

Yes, I think this has potential. It would make the request for access more flexible. It might be possible to retain the distinction between authority and delegate? An authority will never need to request access, and a delegate won't need to register or login.

bgins:
I was discussing a use cases for multiple levels of subdelegation with @QuinnWilton today, and I think we have a good example.

Suppose we have the following apps:

- An auth lobby (an authority)
- Diffuse, a music player (a delegate)
- Notify, another music player (a delegate)
- MetaFixer, an app that applies metadata to music files (a delegate)

Diffuse and Notify request access from the auth lobby to access `public/music` and `public/notify`, respectively. These are the directories where they store music.

MetaFixer is a secondary app that is accessible from Diffuse and Notify. It doesn't do much on its own, but on requesting access from Diffuse or Notify, it will look up metadata and store it with your music.

Diffuse and Notify have a button that starts linking directly with MetaFixer, subdelegating access to the music directories they have access to.

Alternatively, the user could be redirected to the MetaFixer where they request permission from the auth lobby. The disadvantage to this approach is the user loses context when they request access from the auth lobby. 

Also, MetaFixer could request access to directories that it doesn't need. The user may not be aware that MetaFixer is asking for more, whereas requesting from Diffuse or Notify guarantees MetaFixer can only request what it needs.

Note also that MetaFixer will get access on an as needed basis. A user can start by applying metadata to music in `public/music` for Diffuse without access to `public/notify`. Later, they can request access through Notify for `public/notify`. In each case, they retain context for the access they are granting.

QuinnWilton:
👋 I just read through this thread too, and I have one more thought on a possible use-case for sub-delegation:

> We assume there's an "authority" which requires full read & write access to the file system and a "delegate" which could operate with a very small subset of that (eg. only read portion of file system). But is that always the case? 🤷‍♂️

I actually don't think this is necessarily always the case, even if in the current world it is! Baked into this sentence is the assumption that WNFS is the only service for which an authority may delegate access to, but there's a possible future where apps built using OddSDK may integrate with other UCAN enabled services too, for things like message buses, email, or SMS.

If these services are third-party, then the Fission auth lobby isn't likely to serve as an authority for all of them, and so some applications may require delegations from multiple different authorities. This means that it actually becomes possible for individual applications to hold more capabilities than any individual authority, and so the ability to sub-delegate from that rights-amplified application reduces the friction involved in sharing those capabilities with other applications.

In practice, this might show up as something like a Heroku or AWS style dashboard that accumulates capabilities from disparate, possibly third-party services, and can then further attenuate those capabilities to other applications.

Anyway, this is just me spitballing about another instance where sub-delegation shows up, and I understand that this isn't how things work today, so this isn't me suggesting that it's a use-case to be targeted by you in this refactor.

icidasset:
> I was discussing a use cases for multiple levels of subdelegation with @QuinnWilton today, and I think we have a good example.

Thanks you two! And for picking an example I can relate to ❤️ Sounds like a great use case.

> This means that it actually becomes possible for individual applications to hold more capabilities than any individual authority, and so the ability to sub-delegate from that rights-amplified application reduces the friction involved in sharing those capabilities with other applications.

Makes sense yeah. I'm actually getting very close to this, so I'll start working in that direction for the API 👍

README.md
```diff
@@ -174,34 +172,36 @@ const { Branch } = odd.path
 
 // List the user's private files
 await fs.ls(
-  odd.path.directory(Branch.Private)
+  odd.path.directory(Branch.Private),
```
bgins: Is this trailing comma from auto-linting? My preference would be to omit these trailing commas in function params and objects, but purely a style preference. They mostly feel like syntactic noise to me.

src/components/account/implementation.ts
```diff
@@ -0,0 +1,55 @@
...
+
+////////
+// 🧩 //
+////////
+
+export type AnnexParentType = Record<string, Function>
+
+export type Implementation<Annex extends AnnexParentType> = {
+  /**
+   * Additional methods you want to be part of `program.accounst`
```
bgins:
```suggestion
   * Additional methods you want to be part of `program.accounts`
```

src/events.ts
```diff
@@ -27,27 +25,32 @@ export type ListenTo<EventMap> = Pick<
...
+export type CapabilityConsumer = {
+  "capabilities:challenge": any // TODO
 }
 
+export type CapabilityProvider = {
+  "capabilities:approved": void
+  "capabilities:challenge": any // TODO
+  "capabilities:dismissed": void
+  "capabilities:query": { capabilities: Record<string, any> } // TODO
+}
```
bgins: Will these be renamed to `authority`? Also, will consumer be renamed to requestor? That would align with the terms used in the new AWAKE spec.

Understood this code is WIP, so no need to change anything on this PR. 🙂

bgins (commented):
Left a few comments on my initial scan through. More to come! 👀

src/events.ts
icidasset: Ah yes, forgot to rename these. I'll change it to requestor 👍

README.md
icidasset: It's the formatter yeah, I've just adjusted the formatting so that it doesn't do this.
I've disabled trailing commas for arguments and params.
If it needs to be changed elsewhere let me know, these are all the options: https://dprint.dev/plugins/typescript/config/

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+  "typescript.tsdk": "node_modules/typescript/lib"
+}
```
bgins: Do we need this to set the TypeScript version? Can VSCode auto-detect it?

icidasset: Oh whoops, added this by accident. Will remove 👍

src/index.ts
```diff
@@ -1,214 +1,127 @@
...
- * This will give you a `Program` object which has the following properties:
- * - `session`, a `Session` object if a session was created before.
- * - `auth`, a means to control the various auth strategies you configured. Use this to create sessions. Read more about auth components in the toplevel `auth` object documention.
- * - `capabilities`, a means to control capabilities. Use this to collect & request capabilities, and to create a session based on them. Read more about capabilities in the toplevel `capabilities` object documentation.
- * - `components`, your full set of `Components`.
+ * This will give you a `Program` object which will your main interaction point.
+ *
+ * This gives you three systems to work with:
+ * - `account`, the account system, use this to register an account.
+ * - `authority`, the authority system, request or provide authority to parts of the (or entire) file system and account system.
```
```suggestion
 * - `authority`, the authority system, request or provide authority to parts of (or the entire) file system and account system.
```

Rearranged this a bit. Question though, can authority be requested for parts of the account system? If so, would recommend this instead:

```
* - `authority`, the authority system, request or provide authority to parts of (or the entire) file system and account system.
```

icidasset: Yeah, parts of both the file and account system.

bgins:
> Removed DIDs from the Program object, these don't really make sense anymore. Agent DID should never be used anymore by the developer, always use identifier DID.

I think this information could still be useful to developers for understanding authorization flows. They may want to log the agent DID to track which agent they are working in, and we might want to show this information in the browser extension. We will eventually extend the browser extension to display UCANs, so this information will be good context for that as well.

> File system is not loaded automatically anymore.

Yep, agreed it will be better to require developers to handle this explicitly. 👌

src/index.ts
bgins: Gotcha, I updated the suggestion to reflect that. 👌

src/index.ts
```diff
@@ -218,8 +131,20 @@ export type FileSystemShortHands = {
...
-  if (!settings) throw new Error("Expected a settings object of the type `Partial<Components> & Configuration` as the first parameter")
+export async function program<Annex extends AnnexParentType>(
+  settings: Partial<Components<Annex>> & { account: Account.Implementation<Annex> } & Configuration
+): Promise<Program<Annex>>
+export async function program(
+  settings: Partial<Omit<Components<FissionAccountsAnnex>, "account">> & Configuration
+): Promise<Program<FissionAccountsAnnex>>
+export async function program(
+  settings: Partial<Components<any>> & Configuration
+): Promise<Program<any>> {
```
bgins, line 175: Can we get rid of these `any`s? Is it possible to use the `Annex extends Account.AnnexParentType` used in the `Program` type?

icidasset: I haven't found a way to get this working without `any` sadly. It should be fine though because this is the implementor type, Typescript will be mainly looking at types preceding this type (last type is a special case).

src/components/account/implementation/local-only.ts
```diff
@@ -0,0 +1,100 @@
...
+    (set: Set<string>, identifierUcan): Set<string> => {
+      const iss = rootIssuer(identifierUcan, ucanDictionary)
+      return set.add(iss)
+    },
+    new Set() as Set<string>
+  )
+
+  if (rootIssuers.size > 1) {
+    console.warn(
+      "Encounter more than one root issuer in the identifier UCANs set. This should ideally not happen. Using the first one in the set."
```
bgins:
```suggestion
      "Encountered more than one root issuer in the identifier UCANs set. This should ideally not happen. Using the first one in the set."
```

src/ucan/lookup.ts
```diff
@@ -0,0 +1,78 @@
...
+        // Always prefer the first proof.
+        // TBH, not sure what's best here.
+        if (acc) return acc
+
+        const prfUcan = ucanDictionary[prf]
+        if (!prfUcan) throw new Error("Missing a UCAN in the repository")
+
+        return rootIssuer(prfUcan, ucanDictionary)
+      }
+    )
```
bgins, line 44: What are the cases where we might have multiple options for a root issuer? Is the function only for resolving top-level identifier DIDs or is it more general?

src/components/account/implementation/fission-base.ts
```diff
@@ -0,0 +1,294 @@
...
+
+  email = email.trim()
+
+  const rawtoken = await Ucan.build({
+    audience: await Fission.did(endpoints, dependencies.dns),
+    issuer: await Ucan.keyPair(dependencies.agent),
+    capabilities: [DELEGATE_ALL_PROOFS],
+    facts: [{ placeholder: "none" }], // For rs-ucan compat (obsolete, but req'd for 0.2.x)
+    proofs: [], // rs-ucan compat
+  })
```
bgins: What are the time bounds on this token? It might be good to explicitly declare them here so we can easily see what they are.

src/components/account/implementation/fission-base.ts
```diff
@@ -0,0 +1,294 @@
...
+  return response.ok
+    ? { ok: true }
+    : { ok: false, reason: `Server error: ${response.statusText}` }
+}
+
+export async function canRegister(
+  endpoints: Fission.Endpoints,
+  dependencies: Dependencies,
+  formValues: Record<string, string>
+): Promise<{ ok: true } | { ok: false; reason: string }> {
```
bgins: Are we re-creating the `Result` type here and in other places?

We have a `Result` implementation here: https://github.com/oddsdk/ts-odd/blob/f90bde37416d9986d1c0afed406182a95ce7c1d7/src/common/types.ts#L6-L8

But perhaps it's a bit different because the error cases are `Error` and not strings. Also, we often don't have a value to return.

src/components/account/implementation/fission-base.ts
```diff
@@ -0,0 +1,294 @@
...
+
+export async function register(
+  endpoints: Fission.Endpoints,
+  dependencies: Dependencies,
+  formValues: Record<string, string>,
+  identifierUcan: Ucan.Ucan
+): Promise<
+  { ok: true; ucans: Ucan.Ucan[] } | { ok: false; reason: string }
+> {
+  let username = formValues.username
```
bgins: Should this function trim the username like `canRegitser` does?

src/components/account/implementation/fission-base.ts
```diff
@@ -0,0 +1,294 @@
...
+  identifierUcans: Ucan.Ucan[],
+  ucanDictionary: Ucan.Dictionary
+): Promise<boolean> {
+  const facts = identifierUcans.map(
+    ucan => listFacts(ucan, ucanDictionary)
+  )
+
+  // TODO: Check if we have the capability to update the data root.
+  //       Or in the case of the old Fission server, any account UCAN.
+  return facts.some(f => !!f["username"])
```
bgins: Why are facts used to determine capability to update the data root?

src/components/account/implementation/fission-base.ts
```diff
@@ -0,0 +1,294 @@
...
+  const ucan = await Ucan.build({
+    // Delegate to self
+    audience: await AgentDID.signing(dependencies.agent),
+    issuer: await Ucan.keyPair(dependencies.agent),
+
+    capabilities: [DELEGATE_ALL_PROOFS],
+    proofs: await Promise.all(
+      proofs.map(prf => Ucan.cid(prf).then(c => c.toString()))
+    ),
+  })
```
bgins: Similar comment to above. Can we make the time bounds explicit or add a comment about it?

src/components/agent/implementation/web-crypto-api.ts
```diff
@@ -0,0 +1,74 @@
+import { DIDKey } from "iso-did/key"
+import { spki } from "iso-signatures/spki"
+import * as crypto from "../../../common/crypto.js"
+import { Store } from "../../../common/crypto/store.js"
+import { hasProp, isObject } from "../../../common/index.js"
```
bgins:
```suggestion
import { isObject } from "../../../common/index.js"
```

Looks like `hasProp` is unused.

src/index.ts
```diff
@@ -230,728 +155,512 @@ export async function program(settings: Partial<Components> & Configuration): Pr
...
 /**
- * Predefined crypto configurations.
+ * Predefined channel configurations.
  *
- * The crypto component is responsible for various cryptographic operations.
- * This includes AES and RSA encryption & decryption, creating and storing
- * key pairs, verifying DIDs and defining their magic bytes, etc.
+ * A channel serves as a transport between multiple devices,
+ * used to transfer UCANs and file system secrets, but also,
+ * to notify other devices of file system changes.
```
bgins, line 283: > to notify other devices of file system changes

So excited for when we have this. 😄

src/components/manners/implementation.ts
```diff
@@ -45,10 +23,35 @@ export type Implementation = {
...
+   * Cabinet configuration.
+   *
+   * Responsible for storing UCANs and keys.
+   */
+  cabinet: {
+    hooks: {
+      inventoryChanged: (cabinetItems: CabinetCollection) => Promise<void>
     }
   }
-}
```
bgins: Very cool. Can see this hook being useful in a bunch of places. 💯

bgins (commented):
Have looked through all the components. A bit more to go, but getting there. 😃

src/components/manners/implementation.ts
icidasset: I just noticed I accidentally exposed the `cabinet` naming here. Should we rename this to `authority` and/or move it to the authority component? 🤔

src/components/account/implementation/fission-base.ts
icidasset: Not sure, this module is still in progress, Blaine is working on this. But we'll add them for sure 👍

src/components/account/implementation/fission-base.ts
icidasset: Yeah, still working on this module.

src/components/account/implementation/fission-base.ts
icidasset: See the TODO above this line.
It will check for an actual capability in the future.

src/components/account/implementation/fission-base.ts
icidasset: Yeah good point. I haven't found a perfect pattern for this, we should probably use the same pattern everywhere. The last iteration of this used with the `isGranted` function I like the most so far, which would be:
```ts
// isGranted()
{ granted: true } | { granted: false, reason: string }
```
I think this pattern is ok?
The boolean property is named according to the name of the function.
So for `register` we could do: `{ registered: true } | ...`
Thoughts?

I'm not entirely sure we can make a generalised type for this 🤔  We could have 0, 1 or more values next to the boolean property, and maybe a reason why the boolean is false or an error.

src/ucan/lookup.ts
icidasset: Yeah this function is lacking in several areas. Ideally it should use the implementation in the used UCAN library. It's just to get the root issuer. I don't see a use case in our ecosystem, but since the SDK can be used with other ecosystems as well, I'm just preparing for any possibility.

bgins: Makes sense. If I were to guess, other ecosystems would want to (optionally) pass in a capability as a discriminant to do determine root issuer for what? If they all have the same capability, it probably doesn't matter.

src/components/account/implementation/fission-base.ts
bgins: Sounds good 👍

src/components/account/implementation/fission-base.ts
bgins: Yeah, that's a great point. Naming the booleans attaches some semantic information that we'd lose returning `ok` everywhere. Also not sure how we would implement that in general.

For returning multiple values, we could package them up as an object and let the generic fill in the details. We probably would only return one error at a time, so likely not an issue there.

src/components/manners/implementation.ts
bgins: Yeah! That fits with my understanding of the `authority` component so far.

Say we are linking a device, and we emit events `challenge` and `approved` events. We have been preparing to send UCANs and secrets that will end up in the cabinet. Once they are added, authority has changed so it would make sense to expect an `authority:changed` event instead of `collection:changed`.

That's speaking in terms of events, but I think it works for the hooks as well. Not sure about all the details, but something like:

```typescript
  authority: {
    hooks: {
      changed: (cabinetItems: CabinetCollection) => Promise<void>
    }
  }
```

src/ucan/lookup.ts
icidasset: Yeah maybe. I also made sure that this function is only used in the implementations themselves. So if people were to write another account system implementation, they can use something else.

src/index.ts
bgins: Ah yeah, I see. This is for the overload.

It might be possible to define each of the program types separately and then union them in the function signature to avoid the function overload. Not entirely sure this would work, but maybe.

bgins: I'll add though that this use of `any` is not that big of deal because it is localized and won't be consumed elsewhere.

src/components/manners/implementation.ts
icidasset: Hmm yeah, speaking about events. I might just make this a `authority` event instead of a hook, since we have the events anyway 🤷 I should also mention that the `collection:changed` thing is an internal thing for all repositories (cabinet is a repository, as is the cid log).

src/common/crypto.node.test.ts
```diff
@@ -0,0 +1,21 @@
+import expect from "expect"
+
+import * as Crypto from "./crypto.js"
+
+describe("Crypto", async () => {
```
bgins, line 5: It might be good to add an encrypt/decrypt test to cover the exchange key use case.

src/components/account/implementation/fission-base.ts
icidasset: Removed the result type and replaced the `{ ok }` stuff with better naming.

src/index.ts
icidasset: > Not entirely sure this would work, but maybe.

Yeah tried that, doesn't seem to work.

src/common/crypto.node.test.ts
icidasset: Added 👍

src/components/manners/implementation.ts
icidasset: Removed the hook for now, added a note to add an event for this later when we figure out the details of the authority interface.

icidasset:
@bgins Added back the DID shorthand methods, except for the sharing one, because that still needs to be figured out.

src/common/event-emitter.ts
hugomrdias, line 1: we should look into using https://github.com/sindresorhus/emittery for a more robust event emitter

src/common/fission.ts
hugomrdias, line 1: should i add did dns resolver to iso-did for this ?

src/components/account/implementation.ts
```diff
@@ -0,0 +1,55 @@
...
+  canRegister: (formValues: Record<string, string>) => Promise<
+    { canRegister: true } | { canRegister: false; reason: string }
+  >
+
+  /**
+   * How to register an account with this account system.
+   */
+  register: (formValues: Record<string, string>, identifierUcan: Ucan) => Promise<
+    { registered: true; ucans: Ucan[] } | { registered: false; reason: string }
+  >
```
hugomrdias, line 30: can we normalize the return types to a generic `Result` type { result?: T, error: Error }

src/components/account/implementation/fission-base.ts
```diff
@@ -0,0 +1,294 @@
...
+      ],
+      // TODO: We need some UCANs here. We should get capabilities from the Fission server.
+    }
+  }
+
+  return {
+    registered: false,
+    reason: `Server error: ${response.statusText}`,
+  }
+}
```
hugomrdias: we should extract code like this to a fission server client and also export from that odd sdk components

src/components/account/implementation/fission/data-root.ts
```diff
@@ -70,66 +72,60 @@ export async function lookupOnFisson(
...
     else dependencies.manners.log("🔥 Failed to update DNSLink for:", cid)
-    return { success: response.status < 300 }
-
+    return response.ok ? { updated: true } : { updated: false, reason: response.statusText }
   }).catch(err => {
     dependencies.manners.log("🔥 Failed to update DNSLink for:", cid)
     console.error(err)
-    return { success: false }
-
+    return { updated: false, reason: err }
```
hugomrdias, line 116: these return types should be normalised IMO and should return something other than a boolean, either cid or an Error. Especially `reason` should only be of one single type.

src/components/agent/implementation.ts
```diff
@@ -0,0 +1,44 @@
...
+  /**
+   * This goes hand in hand with the DID `keyTypes` record from the crypto component.
+   */
+  keyAlgorithm: () => Promise<KeyType>
+
+  /**
+   * The JWT algorithm string for agent UCANs.
+   */
+  ucanAlgorithm: () => Promise<SignatureAlgorithm>
+}
```
hugomrdias, line 44: why are these async ? they should be pretty much constants

src/components/agent/implementation.ts
```diff
@@ -0,0 +1,44 @@
...
+
+  /**
+   * Key pair used to sign data.
+   */
+  signingKey: () => Promise<CryptoKeyPair>
+
+  /**
+   * DID associated with this agent.
+   */
+  did: () => Promise<string>
```
hugomrdias, line 18: why is this async ?

src/components/agent/implementation/web-crypto-api.ts
hugomrdias, line 1: why didnt you use the RSA Signer from iso-signature here ?

src/components/identifier/implementation/web-crypto-api.ts
```diff
@@ -0,0 +1,28 @@
...
+    WebCryptoAPIAgent.createSigningKey
+  )
+
+  const exportedKey = await exportPublicKey(signingKey).then(spki.decode)
+
+  return {
+    did: async () => DIDKey.fromPublicKey("RSA", exportedKey).toString(),
+    sign: async data => WebCryptoAPIAgent.sign(data, signingKey),
+    ucanAlgorithm: async () => "RS256",
+  }
```
hugomrdias, line 27: same here why not just use Signer ?

src/ucan/lookup.ts
bgins: Got it, yeah. That sounds more flexible for folks in general. 👌

src/components/manners/implementation.ts
bgins: Sounds good 👌

hugomrdias (approved):
LGTM left a few comments but nothing blocking we should merge asap and work of small PRs

src/fileSystem.ts
```diff
@@ -1,276 +1,190 @@
...
+ */
+export async function accountDID<Annex extends AnnexParentType>({ account, cabinet, identifier }: {
+  account: Account.Implementation<Annex>
+  cabinet: Cabinet
+  identifier: Identifier.Implementation
+}): Promise<string> {
+  const identifierDID = await identifier.did()
+  const identifierUcans = cabinet.audienceUcans(identifierDID)
+  return account.did(identifierUcans, cabinet.ucansIndexedByCID)
 }
```
bgins, line 163: Writing this out to check my understanding of the interaction between the identifier and the account system.

The user will have a signing key which is associated with their identifier DID:

```typescript
const identifierDID = await identifier.did()
```

The account system will have delegated authority to the identifier DID, giving it the ability to sync files to the service (e.g. the Fission server). We gather any UCANs that have delegated to the identifier DID:

```typescript
const identifierUcans = cabinet.audienceUcans(identifierDID)
```

Once we have these UCANs, we can determine the account DID (the issuer in one of the UCANs):

```typescript
return account.did(identifierUcans, cabinet.ucansIndexedByCID)
```  

The account DID is associated with a key pair created on the Fission server at account creation time, which is immediately burned to prevent the server from mutating the user's file system without permission from the user.

Am I understanding that all correctly?

bgins (approved):
Fantastic work on this! So much good stuff in here. ✨ 

I'm enjoying how the new auth and file system APIs are shaping up. It's looking great!

src/common/event-emitter.ts
bgins: It may not be worth adding a dependency if we don't need the additional functionality.

src/components/account/implementation.ts
bgins: @hugomrdias a bit of discussion around this here: https://github.com/oddsdk/ts-odd/pull/532#discussion_r1267286390

src/common/event-emitter.ts
bgins: Adding some additional context. We considered using Emittery when we initially implemented the event emitter. 

Our thought process was: This is simple enough. We can implement it ourselves and save on bundle size while reducing the risk of supply chain attacks by having one less dependency.

Our plan was to extend it as needed, but at a certain point it may not be worth our time and effort.

src/common/fission.ts
icidasset: Up to you. Would it allow you to customise how the DNS resolving is done?

src/components/account/implementation/fission-base.ts
icidasset: Good idea yeah. Not sure if we'd ever make use of that outside of the SDK though, since the CLI will be written in Rust.

src/components/agent/implementation.ts
icidasset: Just keeping all the options open, doesn't hurt since all the functions that depend on these are async anyway.

src/components/agent/implementation.ts
icidasset: Same answer as with the agent methods.

src/components/agent/implementation/web-crypto-api.ts
icidasset: Oh yeah, totally forgot about that. I wrote this before you released that, will change soon.

src/fileSystem.ts
icidasset: > Am I understanding that all correctly?

Pretty much yeah.

> The account DID is associated with a key pair created on the Fission server at account creation time, which is immediately burned to prevent the server from mutating the user's file system without permission from the user.

I just realised made some mistakes here. Blaine very recently made this diagram: https://github.com/fission-codes/fission-server/blob/main/fission-server/docs/account-creation.md

The account DID isn't going to be the identifier DID of the device that you used to sign up. Instead the public key of the key pair generated on the server. So instead of looking for the root issuer, we'll have to look for the UCAN we got back from the server containing the server capabilities.

src/common/event-emitter.ts
hugomrdias: I agree with that approach with a few exceptions one being event emitter because they are very very tricky and are prone to memory leaks and uncaught exceptions. 
`emittery` defends against a bunch of weird stuff ppl do inside handlers and adds a couple of nice to haves, it net positive IMO.

https://pkg-size.dev/emittery its small and has 0 deps.

src/common/fission.ts
hugomrdias: i would do DoH by default with round robin over like 5 options and allow for optional custom resolver

src/components/account/implementation/fission-base.ts
hugomrdias: I really dislike this, IMO we should have a consistent Result type in the public and private API. We should always have something to return with a few exceptions like `isUsernameAvailable` and similar things and failures should always be an Error instance with a `cause: Error` from the upstream error

Something like
```js
export interface ResultError {
  error: OddError
  result?: null
}

export type OddResult<R> =
  | {
      error?: null
      result: R
    }
  | ResultError

```

src/components/account/implementation/fission-base.ts
hugomrdias: it allows for easier and faster iteration on server logic without messing with the SDK and we extract Fission related components

src/components/agent/implementation.ts
hugomrdias: its creating micro tasks in a bunch of places to return a constant string

src/components/account/implementation/fission-base.ts
icidasset: You can already iterate on that without messing with the SDK though, thanks to component system:
```ts
odd.program({
  account: fissionAccountComponentWithAdjustments
})
```

src/common/event-emitter.ts
icidasset: I'd be down for this change as well.

bgins: Alright, let's change it. One request, we should make sure the type information is not degraded making the switch.

src/components/account/implementation/fission-base.ts
bgins: I find this interface confusing. How does one discriminate between success and failure? 

Also, why would success have an `error` and failure have a `result`? It seems this information is not needed in these cases.

But -- we've closed this PR now, so maybe better to continue this discussion elsewhere. 😄

src/fileSystem.ts
bgins: Alright, thanks. The pieces are all starting to make sense now. 😄

PR #533, opened 2023-07-12
Testing Server Integration

blaine:


src/components/account/implementation/fission-base.ts
```diff
@@ -6,224 +6,276 @@ import { DELEGATE_ALL_PROOFS } from "../../../ucan/capabilities.js"
...
+      ).join(", ")}.`,
+    };
 
-  username = username.trim()
+  email = email.trim();
 
-  if (Fission.isUsernameValid(username) === false) return {
-    ok: false,
-    reason: "Username is not valid."
+  const rawtoken = await Ucan.build({
```
blaine, line 42: This is the token that's not working 🙃

PR #535, opened 2023-07-20
Improve bundle size

icidasset:
This gets the core of the SDK down to 97KB + 173KB shared minified chunk as opposed to 983KB which it is now.
Downside is that you will have to import the Fission component set explicitely:

```ts
import * as fission from "@oddjs/odd/compositions/fission"

// REGULAR USAGE
const config = { namespace: "odd-test", debug: true }
const components = await fission.components(config)

odd.program(config, components)

// CUSTOM COMPONENTS
// Note: I moved the implementations up a directory so it's easier to import
import * as localAccount from "@oddjs/odd/components/account/local"

odd.program(
  config,
  { ...components, account: await localAccount.implementation({ storage: components.storage }) }
)
```
It removes some of the magic, but it does show the developer that other environments/setups are possible.

Thoughts?

## Bundled output

```
173k ├── chunk-S3AV6ETP.js
 52k ├── chunk-S3AV6ETP.js.gz
815k ├── chunk-S3AV6ETP.js.map
   - ├── compositions
763k │  ├── fission.esm.min.js
213k │  ├── fission.esm.min.js.gz
3.3M │  └── fission.esm.min.js.map
 97k ├── index.esm.min.js
 27k ├── index.esm.min.js.gz
501k ├── index.esm.min.js.map
   - └── umd
   -    ├── compositions
927k    │  ├── fission.umd.min.js
261k    │  ├── fission.umd.min.js.gz
4.1M    │  └── fission.umd.min.js.map
267k    ├── index.umd.min.js
 76k    ├── index.umd.min.js.gz
1.3M    └── index.umd.min.js.map
```

## Question

I'm not sure if we need anything in our import/export map in package.json to refer to our ESM/UMD bundles?


scripts/build-minified.js
```diff
@@ -9,7 +9,7 @@ const globalName = "oddjs"
 console.log("📦 Bundling & minifying...")
 
 await esbuild.build({
-  entryPoints: ["src/index.ts"],
+  entryPoints: ["src/index.ts", "src/compositions/fission.ts"],
```
icidasset, line 12: I wish I could use a glob here or point to a directory 🤔  We should probably come back to this script later when add more compositions and then use Node to list the files in that directory.

src/components/account/README.md
```diff
@@ -0,0 +1 @@
+An account system is responsible for user accounts and having a location for where the data root is stored. This allows for the file system to be loaded on another device, even if the related devices are offline.
```
icidasset, line 1: I moved the explanations of each component to a README file in their respective directory, since we no longer have those component variables in the main index file.

src/components/account/fission.ts
```diff
@@ -306,9 +307,11 @@ export async function did(
 ////////
 
 export function implementation(
-  endpoints: Fission.Endpoints,
-  dependencies: Dependencies
+  dependencies: Dependencies,
+  optionalEndpoints?: Fission.Endpoints
 ): Implementation<Annex> {
+  const endpoints = optionalEndpoints || Fission.PRODUCTION
```
icidasset, line 313: All Fission components now default to production env, instead of having multiple files that you could import.

src/index.ts
```diff
@@ -577,165 +288,3 @@ export async function assemble<Annex extends AnnexParentType>(
...
-    return await new Promise(resolve => {
-      const db = indexedDB.open("testDatabase")
-      db.onsuccess = () => resolve(true)
-      db.onerror = () => resolve(false)
-    })
-  } catch {
-    console.warn("IndexedDB is not supported")
-    return false
-  }
-}
```
icidasset, line 726: I moved this `isSupported` check to the storage component itself so this behaviour can be customised.

avivash (approved):
Looks great! I love the `compositions` idea too 👏🏼

hugomrdias (approved):
LGTM

src/compositions/fission.ts
```diff
@@ -0,0 +1,52 @@
...
+import * as Identifier from "../components/identifier/web-crypto-api.js"
+import * as Manners from "../components/manners/default.js"
+import * as Storage from "../components/storage/indexed-db.js"
+
+/**
+ * The default Fission stack using web crypto auth and IPFS.
+ */
+export async function components(
+  settings: Configuration & {
+    environment?: string
```
bgins, line 20: `environment` is unused. Should this be passed to the account implementation?

src/index.ts
```diff
@@ -1,13 +1,10 @@
...
-
 import * as Auth from "./auth.js"
 import * as Events from "./events.js"
 import * as Path from "./path/index.js"
 import * as Cabinet from "./repositories/cabinet.js"
 import * as CIDLog from "./repositories/cid-log.js"
 
 import { Query } from "./authority/query.js"
-import { Store } from "./common/crypto/store.js"
 import { Account, Agent, Channel, DNS, Depot, Identifier, Manners, Storage } from "./components.js"
```
```suggestion
import { Account } from "./components.js"
```

Looks like after the changes we are only using `Account`.

src/compositions/fission.ts
icidasset: Ah yup, good catch! Fixed it 👍

bgins (approved):
Looks good! And wow, that simplifies `src/index.ts` quite a bit. 🎉

PR #536, opened 2023-07-24
Update WNFS & UnixFS compat layer

### LINKED ISSUE #546: rs-wnfs - Unix FS tree (pretty tree)
ETA: 2023-09-15

icidasset:
Updates `wnfs` to the latest version: Doesn't use name filters anymore, `AccessKey` instead of `PrivateRef` and fixes a bug we had with older access keys.

This PR also adds the UnixFS compatibility layer. Any time a mutation is made in the public file system, the change is reflect in the UnixFS node. This was added so that public content can be rendered on the IPFS gateway (eg. apps or images)

Other changes:
- Renamed `capsuleRef` to `capsuleKey`
- Now uses a single `Uint8Array` instead of an object with multiple.
- No more need for custom encoding/decoding of the "access key" 🎉 
- Split up event types to avoid circular imports.
- Make `appPath` function applicable to both public and private paths.

Closes #546 and #547


src/events/authority.ts
```diff
@@ -0,0 +1,10 @@
+export type AuthorityRequestor = {
+  "challenge": undefined // TODO
+}
+
+export type AuthorityProvider = {
+  "approved": undefined
+  "challenge": undefined // TODO
+  "dismissed": undefined
+  "query": Record<string, any> // TODO
+}
```
matheus23, line 10: Some remaining TODOs. Just flagging in case it was unintentional

src/fs/rootTree.ts
```diff
@@ -25,20 +30,38 @@ export type RootTree = {
...
+  const rsaMod = await webcrypto.subtle
+    .exportKey("jwk", rsaKey.publicKey)
+    .then(a => {
+      if (a.n) return a.n
+      else throw new Error("Expected public RSA key to have `n` property")
+    })
+    .then(n => Uint8Arrays.fromString(n, "base64url"))
+
+  return new PrivateForest(rng, rsaMod)
+}
```
matheus23, line 52: :+1:

matheus23 (commented):
`unixfs.ts` looks like what I'd expect :) :+1: 

Keep in mind, we'll be doing a release of rs-wnfs in like, just an hour or so. That'll make your tests run faster.

Also, have you tried using rs-wnfs' native `.cp` function in the case where you're copying files or directories within the private partition? It should be a little faster than doing more round-trips from JS to Wasm.
And after the new rs-wnfs release, there will also be a public filesystem `cp` function.

src/events/listen.ts
```diff
@@ -0,0 +1,39 @@
...
+export type Listener<EventMap extends Record<string, unknown>, Name extends keyof EventMap> = (
+  eventData: EventMap[Name]
+) => void | Promise<void>
+export { Listener as EventListener }
+
+/**
+ * Events interface.
+ *
+ * Subscribe to events using `on` and unsubscribe using `off`.
+ * There's also `once`, `onAny`, `offAny`, `anyEvent` and `events`.
```
avivash, line 12: Love this!

src/fs/class.ts
```diff
@@ -318,21 +326,33 @@ export class FileSystem {
...
   ): Promise<AnySupportedDataType<V>> {
     return this.#transactionContext().read<DataType, V>(path, dataType, options)
   }
 
+  /**
+   * Create a permalink to some public file system content.
+   * @group Querying
+   */
+  permalink(dataRoot: CID, path: Path.Distinctive<Path.Partitioned<Path.Partition>>): string {
+    if (this.#dependencies.depot.permalink) {
```
avivash, line 344: note: when i started integrating the `ts-odd` `next` branch into collector this week I had to add this `babel` plugin to handle `#` methods:
```
@babel/plugin-transform-private-methods
```
This may just be a quirk of react-native(I haven't tried the latest updates in a svelte app yet), but perhaps it's something we want to highlight in the readme?

src/fs/unix.ts
```diff
@@ -0,0 +1,203 @@
...
+ */
+export async function store(node: PBNode, depot: Depot.Implementation): Promise<CID> {
+  return depot.putBlock(
+    DagPB.encode(node),
+    DagPB.code
+  )
+}
+
+////////
+// ㊙️ //
```
avivash, line 188: I still find the emoji section headers a bit confusing 🙈 after some searching, it looks like this means secret? ㊙️ Should we just add the word `Secret` to these sections? May be clearer for outside contributors 🤔

avivash (approved):
I'm not fully up-to-speed on the requirements of the UnixFS compatability layer, but I think everything else looks good! Left a few notes

src/fs/class.ts
icidasset: Ah good point, maybe it’s because it’s a relatively new js feature (2021), ecmascript 2022 I think. The readme still needs to be rewritten but yeah we should definitely mention that. Can you change the compile target to es2022 in react native?

src/fs/unix.ts
icidasset: Ah right I promised you that emoji guide, I'll do that in this PR. I like that the un-exported items are grouped together in a section that is color coded by a red emoji 😅 (and conveniently it's the sign for "secret" yeah). I can also append the word secret after the emoji? 🤷

icidasset:
Thanks @matheus23! I was waiting for the new release for the `cp` stuff, I'll add that tomorrow now that the release is out 👍

src/fs/class.ts
avivash: looks like I can! It's strange, the app extends the expo config, which specifies a target of `ESNext`, which seems like it should have that functionality, but specifying `ES2022` explicitly seems to fix it 👌🏼

src/fs/unix.ts
avivash: haha no worries, yeah I think just including the word `Secret` next to the emoji would be helpful, rather than having an emoji guide to refer to. I'm open to anything though. Whatever the most clear solution is for outside developers to tell what the section refers to 👍🏼

src/fs/class.ts
icidasset: Nice 👍 Yeah, esnext is confusing because it depends on when they last updated the pointer (could even be es2020 🤷). I'll mention in the readme that the SDK requires es2022.

src/fs/unix.ts
icidasset: Oh wait I did already add this, it's in `src/README.md`. It doesn't have an index table though, I'll add that.

PR #538, opened 2023-07-27
Account system & file system loading improvements

icidasset:
This PR refactors a bunch of things related to the account system, the file system, UCANs and some other parts.

## UCAN

I created a new class for the "UCAN Dictionary". This is a layer on top of the cabinet, making it read-only and adding a bunch of query methods to it. Making it safe and easy to pass to places that leak out of the SDK internals (non-SDK code should not be able to manipulate the internal cabinet).

```ts
ucanDictionary.lookupByAudience(audienceDID)
```

## Account component

- Rewrote the account component readme, hopefully a bit clearer on its purpose.
- Instead of passing the identifier UCANs (UCANs from store where audience is the identifier) and a record of `cid → ucan` (what we called UCAN dictionary before), we now pass the identifier component and the new UCAN dictionary (as explained above).
- Removed the data root methods from the implementation (some account systems might not be able to deal with this).
- Added the `hasSufficientAuthority` method, which is used to check if all authority is present.
- Added the `provideAuthority` method, which provides UCANs to other devices/clients.
- The `annex` part of the implementation has been improved so it can be used to implement more useful things, such as the following.
- Added a `volume` method which can be used to load the file system, more on this next.

## File System

It's now possible to load multiple file systems from different users. You have to pass in the associated DID of the file system and optionally the data root (if not passed it creates a new file system). The DID has two roles. First, it's needed to find the capabilities related to the file system in order to make a mutation. And secondly, it's used to make a CID log, which is namespaced per file system. Maybe we could drop the CID log stuff and that would allow us to make that DID argument optional, making it easier to work with.

```ts
// API
const params: { dataRoot?, dataRootUpdater?, did }
program.fileSystem.load(params)

// Progressive file system, uses local file system if not authed, otherwise data root from Fission server
program.fileSystem.load(
  await program.account.volume() // returns { dataRoot?, dataRootUpdater, did }
)

// FUTURE GOAL: Load volume of other fission user
program.fileSystem.load(
  await program.account.volume("other-username")  // returns { dataRoot, did }
)
```

Other changes:
- Moved the online/offline behaviour to their related components (components might not always depend on an internet connection).
- Made the `updateDataRoot` file system param optional.
- Moved the file system events to the file system itself instead of the program, since we're dealing with multiple file systems now.
- Removed as much component dependencies as possible.

## Authority

- Added methods to build authority queries.
- Renamed `isGranted` to `has` 
- `has` function now requires you to pass a one or more queries (account and/or fs).

```ts
const authority = await program.authority.has([
  odd.authority.account,
  ...odd.authority.fileSystem.rootAccess
])

if (authority.has) {
  // Got needed account and file system capabilities
}
```
The query specifying probably could be done better, but it's ok I think?
`odd.authority.account` is a single query and `odd.authority.fileSystem.rootAccess` is multiple.

## More changes

- Improved documentation of the `program` function (let me know what you think of this)
- Added local implementations of the channel and depot components.
- Added a local composition for working solely local.
- Moved the DID functions in the program to `program.did`
- Account DID can be `null` (instead of throwing an error)
- Improves the cabinet so that it can store access keys for multiple file systems (indexed by the file system DID).
- Same for the CID log, adjusted for multiple file systems.


src/components/account/fission.ts
```diff
@@ -19,6 +20,21 @@ export type Annex = {
   requestVerificationCode: (
     formValues: Record<string, string>
   ) => Promise<{ requested: true } | { requested: false; reason: string }>
+
+  /**
+   * Create a progressive volume for a Fission account.
+   *
+   * This method can be used to load a file system before an account is registered.
+   * When you register an account, you can just keep using it as before.
```
bgins:
```suggestion
   * This method can be used to load a local-only file system before an account is registered.
   * When you register an account, the file system will sync with the Fission server, making it available through Fission IPFS nodes.
```

Possible rewrite to acknowledge the benefit of registering an account.

src/components/account/fission.ts
```diff
@@ -300,6 +352,32 @@ export async function did(
   const root = Array.from(rootIssuers.values())[0]
   if (!root) throw new Error("Expected a root issuer to be found")
   return root
+
+  // TODO: Get public key from fission server generated key pair
+  //       -> issuer from fission-server issued UCAN
+}
+
+export async function hasSufficientAuthority(
```
bgins, line 360: What kind of authority does the function check for? Is it full account authority? Or it should work for subdelegation use cases? (Seems like maybe the latter because we still want to load volumes in apps without full account access.)

src/components/channel/local.ts
```diff
@@ -0,0 +1,22 @@
...
+import { Implementation } from "./implementation.js"
+
+////////
+// 🛠️ //
+////////
+
+export function establish(
+  options: ChannelOptions
+): Promise<Channel> {
+  throw new Error("No local channel available just yet.") // NOTE: Do WebRTC implementation?
```
bgins, line 11: Must haz local-only WebRTC party.

src/index.ts
```diff
@@ -122,13 +130,15 @@ export type FileSystemShortHands = {
...
- * This object also has a few other functions, for example to load a filesystem.
- * These are called "shorthands" because they're the same functions available
- * through other places in the ODD SDK, but you don't have to pass in the components.
+ * An ODD program revolves around two main things, the file system (WNFS) and authorization (UCANs).
+ * Both of these can be used entirely locally without depending on an external service.
  *
- * See `assemble` for more information. Note that this function checks for browser support,
- * while `assemble` does not. Use the latter in case you want to bypass the indexedDB check,
- * which might not be needed, or available, in certain environments or using certain components.
+ * On top of that we've built components that can be swapped out with different
```
```suggestion
 * We've built components that can be swapped out with different
```

This reads alright without the transition phrase.

src/index.ts
```diff
@@ -202,29 +218,38 @@ export async function program<Annex extends AnnexParentType>(
...
-      return components.account.did(identifierUcans, cabinet.ucansIndexedByCID)
-    },
-
-    async agentDID() {
-      return components.agent.did()
-    },
-
-    async identifierDID() {
-      return components.identifier.did()
+    dids: {
```
bgins: Not sure, but maybe make this is a singular `did`? That way you would call `program.did.account()`.

bgins:
> It's now possible to load multiple file systems from different users.

Yes, this is amazing. 🔥 

Out of curiosity, will users eventually be able to grant authority to other users to mutate their file system?

```typescript
// Progressive file system, uses local file system if not authed, otherwise data root from Fission server
program.fileSystem.load(
  await program.account.volume() // returns { dataRoot?, dataRootUpdater, did }
)
```

> Added a local composition for working solely local.

This seems like a reasonable way to shape things. If a developer wants a progressive file system, they should expect that it will eventually be associated with an account system. And if they want always local-only, that's an option too.

There is a slight bit of cognitive dissonance with the naming, passing in the account before we have one. But I'm not sure how else we could approach it, so let's go with this for now.

> The query specifying probably could be done better, but it's ok I think?

This looks good to me!

bgins (approved):
Looks good! 🎉  Left a few comments, nothing that needs changing.

src/index.ts
icidasset: 💯

src/components/account/fission.ts
icidasset: It depends on the implementation. The idea is that you'd use a different account implementation based on the authority needed. There is an implementation for a Fission dev account which checks for capabilities to update the volume and register + publish apps. But there's also implementations for app accounts and delegated access, although I'm guessing these could be the same seeing that they only capability they have is to update their volume.

> Seems like maybe the latter because we still want to load volumes in apps without full account access

I should also mention that this method is only for the account capabilities, not file system capabilities and access keys, so loading volumes is independent of this.

icidasset:
Whoops, forgot to answer your other question.

> Out of curiosity, will users eventually be able to grant authority to other users to mutate their file system?

Yes 💯 We now have the UCAN store and file system pieces, we just need to get the UCANs and access keys to the other user. After that the idea is to use the "delegated-access account implementation" that we still need to write, using that you'll be able to update the data root.

bgins:
>> Out of curiosity, will users eventually be able to grant authority to other users to mutate their file system?

> Yes 💯 We now have the UCAN store and file system pieces, we just need to get the UCANs and access keys to the other user. After that the idea is to use the "delegated-access account implementation" that we still need to write, using that you'll be able to update the data root.

That's great! Pinging @therealjeffg because this will support your proposal for developers and users writing data into each other's file systems.

PR #541, opened 2023-07-28
Replace expect with assert

### LINKED ISSUE #540: Use assert in tests
# Summary

## Problem

We would like to run our unit tests in a headless browser, but `expect` uses node APIs that make this difficult.

### Impact

No running unit tests in headless browsers.

## Solution

Replace `expect` with `assert`: https://www.npmjs.com/package/assert

bgins:
## Summary

This PR implements the following features:

* [x] Replace expect with assert

The PR replaces `expect` with `assert` to make browser testing easier. `expect` relies on node APIs while `assert` does not.

We will add browser testing in a future PR.

## Test plan (required)

Run `npm test`. All tests should pass like before.

## Closing issues

Closes #540 

## After Merge
* [x] Does this change invalidate any docs or tutorials? No
* [x] Does this change require a release to be made? No


src/fs/class.node.test.ts
```diff
@@ -1016,16 +949,13 @@ describe("File System Class", async () => {
...
-      await fs.read(Path.file("public", "file"), "utf8")
-    ).toEqual(
+    assert.equal(
+      await fs.read(Path.file("public", "file"), "utf8"),
       "💃"
     )
   })
 
   it("doesn't commit a transaction when an error occurs inside of the transaction", async () => {
-    expect.assertions(1)
```
bgins, line 1027: Wasn't able to find an equivalent in `assert`, but the test passes so probably alright?

icidasset: It only fails in certain scenarios when the code doesn't properly work.
This is saying that the test should run exactly one assertion, no less.
I think we should replace this with something like this from the `assert` lib: https://nodejs.org/api/assert.html#trackerverify

bgins: Oh yeah, how did I miss that. 😆 

Thanks for finding it. Updated the test case to use that.

icidasset (approved):

PR #542, opened 2023-08-01
Integration with the new Fission server

icidasset:
New fission server integration and setup for multiple account types and implementations. With this version of the SDK you can register a new verified account on the Fission server, progressively load your file system and update your volume pointer on the server. All this can be ran locally.

The DIDs used throughout the flow aren't correct yet. Blaine is working on delegation chain resolving to remedy that.

```ts
import * as fission from "@oddjs/odd/compositions/fission"

odd.program(
  config,
  await fission.components(config)
)

// Or use a different account implementation (there's multiple ways of doing this)
odd.program(
  config,
  await fission.accountImplementations.delegated(config)
)
```

Moved the fission account stuff around a bit:
```ts
import * as fissionAccount from "@oddjs/odd/components/account/fission"

const defaultComponents = await fission.components(config, { accountImplementation: "standard" })
const finalComponents = { ...defaultComponents, account: fissionAccount.delegated(defaultComponents) }
```

Using your own Fission server instance:
```ts
const myCustomFissionSetup = {
  apiPath: "/api",
  ipfsGateway: "https://ipfs.accounts.bbc",
  server: "https://server.accounts.bbc",
  userDomain: "accounts.bbc",
}

odd.program(
  config,
  await fission.components(config, { environment: myCustomFissionSetup })
)
```

## Other changes

- Improved the UCAN lookups and added some helper functions for it
- Various improvements to the IPFS Bitswap depot implementation, including webtransport support.
- Updated all dependencies
- Fission DNS implementation
- Better query params type for `authority.has` function
- Modified ts-ucan to generate 0.9 UCANs
- Allows passing other usernames to the `volume` function to load file systems of other users
- Improved the API docs
- Switched to [emittery](https://github.com/sindresorhus/emittery/tree/main) for the event emitter as discussed before (gives us neat things like `once`, `onAny` and async iterators for events)
- Online connectivity tracking can be customised through the `manners` component because `navigator.onLIne` and the associated events are not available on all platforms.
- Channel context

## Test plan

It works with the current `main` branch of the new fission server (commit hash: `421528e005317bd87d35157978f4fe75cc0b3170`). Until Blaine merges his work, then we'll probably need to update the UCANs that we send to the server (or at least we'll want to change this).

When you use Nix with the fission server it should automatically set up a database. IPFS will be needed to started manually I think. Also the current server readme says to start the IPFS node in offline mode, **this does NOT work** to connect to the local browser. Webtransport should work out of the box if your browser supports is (ie. not Safari), Websocket transport in Kubo needs to be enabled manually I believe. Websocket transport can be enabled by adding the `/ip4/0.0.0.0/tcp/4002/ws` and `/ip6/::1/tcp/4002/ws` addresses to `Addresses.Swarm` in the config.

You'll also have to enable CORS on your IPFS node. That can be done adding this to the config:
```json
  "API": {
    "HTTPHeaders": {
      "Access-Control-Allow-Origin": [ "*" ]
    }
  },
```
And finally, you'll have to set the correct peer address on the Fission server.
See the addresses in `src/fission-server/config/settings.yaml` under `ipfs.peers`.
You can override them there or put them in an env variable when you start the server like so:
```shell
APP__IPFS__PEERS="/ip4/127.0.0.1/tcp/4002/ws/p2p/12D3KooWLXH5BFzscChpdjtotAvv94hDR6bVxMKLHBHrSnugCRWs,/ip4/127.0.0.1/udp/4001/quic-v1/webtransport/certhash/uEiC4_sAMUfcxEJtqIlVWRGlHrTSSYuyk5Ulqfl6CjRiOHw/certhash/uEiANUHX9dRBqphQzZINo5WzkStJ7qevCr_2ZAUzLEbqoFw/p2p/12D3KooWLXH5BFzscChpdjtotAvv94hDR6bVxMKLHBHrSnugCRWs" cargo run --features ansi-logs
```


src/components/depot/ipfs-bitswap-websockets.ts
```diff
@@ -28,44 +31,78 @@ export type Transport = {
   libp2p: Libp2p
 }
 
-export async function createTransport(
+export async function createTransport<FS>(
   blockstore: Blockstore,
+  manners: Manners.Implementation<FS>,
```
bgins, line 36: Nice! Good to get the manners in here. 👌

src/components/manners/default.ts
```diff
@@ -1,32 +1,81 @@
...
+  globalThis.addEventListener("offline", async () => {
+    online = false
+    log("🌍 Program is offline")
+    programEmitter.emit("offline", void null)
+  })
+
+  globalThis.addEventListener("online", async () => {
+    online = true
+    log("🌍 Program is online")
+    programEmitter.emit("online", void null)
```
bgins, line 45: Yeah, this is great. Will be nice to have these offline/online events exposed on the program.

src/compositions/fission.ts
```diff
@@ -7,51 +24,137 @@ import * as Account from "../components/account/fission.js"
...
+  /**
+   * The default Fission stack using app and/or verified accounts.
+   *
+   * Verified accounts require an email address. Before registration
+   * users will receive an email with a code or link that they'll
+   * need before creating their actual account.
+   *
+   * App accounts don't require an email address, meaning you can
+   * set up some external identifier verification, but it's not required.
+   * More importantly, the app developer pays for the storage of app accounts.
```
bgins:
```suggestion
   * More importantly, the app developer pays for the storage used by app accounts.
```

src/events/index.ts
```diff
@@ -0,0 +1,75 @@
...
+  "approved": undefined
+  "challenge": undefined // TODO
+  "dismissed": undefined
+  "query": Record<string, any> // TODO
+}
+
+export type FileSystem = {
+  "local-change": { dataRoot: CID; path: DistinctivePath<Partitioned<Partition>> }
+  "publish": { dataRoot: CID; proofs: Ucan[] }
+}
```
bgins, line 52: It looks like these events may not be narrowing to `local-change` and `publish`. With the following code:

```typescript
  fs.on('local-change', event => {
    event
  })
```

The type of `event` is:

```typescript
(parameter) event: {
    dataRoot: odd.CID<unknown, number, number, Version>;
    path: odd.path.DistinctivePath<[odd.path.Partition, ...string[]]>;
} | {
    dataRoot: odd.CID<unknown, number, number, Version>;
    proofs: odd.ucan.Ucan[];
}
```

which includes the parameters for both `local-only` and `publish`, but we should only get the parameters for `local-change`.

bgins (approved):
Looks good! I didn't fully test the registration loop with email, but will test it more fully when the DID flows are better in place.

Left a comment about the file system event emitters.

src/events/index.ts
icidasset: Oh good catch! Fixed 👍

PR #551, opened 2023-08-04
Add TS and rimraf to Flake

expede:
## Summary

I'm on a relatively fresh system, and don't have global JS/TS tooling set up. I was unable to run npm tasks in this repo without installing TypeScript and `rimraf`.

This PR adds them to the Flake environment!

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -17,7 +17,11 @@
       {
         devShell = pkgs.mkShell {
           name = "oddjs";
-          buildInputs = with pkgs; [ nodejs-18_x ];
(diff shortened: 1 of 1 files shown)


icidasset:
I've changed the base to `next` because all our work is being done on that base branch. James also made some changes to the Nix flake on that branch.

You shouldn't need to add these dependencies to the Nix file though, they are installed via NPM 🤔 (see package.json dev dependencies). So not sure why it isn't working for you. What commands are you running that you have trouble with?

PR #564, opened 2023-08-21
Allow storing any type of UCAN in the cabinet

icidasset:
The goal is to allow storing any type of UCAN in the internal cabinet, removing the ties to a specific version of UCAN. I'm calling these "tickets" now in the hopes of removing some confusion, but totally down to rename this if people don't like it (again, this is just for internal use).

**This moves all the UCAN creation into components.** This happens mostly in the (partially) new authority component and in the account component. This allows us to separate ts-ucan from the core odd package.

Most of the code now deals with these "tickets" which contains the encoded UCAN token. This does require us to decode the tokens every time we need to do some searching or filtering. I've avoided this until now because of the types, but I was thinking to avoid this we could keep a simple object around (eg. `Record<string, unknown>`). In that case you'd only need to deal with these types and not decoding. Thoughts?

The UCAN `Dictionary` is now referred to as the tickets `Inventory`.


src/auth.ts
```diff
@@ -21,44 +20,16 @@ async (formValues: Record<string, string>): Promise<
   { registered: true } | { registered: false; reason: string }
 > => {
   // Do delegation from identifier to agent
-  const agentDelegation = await identifierToAgentDelegation({ agent, identifier })
-  await cabinet.addUcan(agentDelegation)
+  const agentDelegation = await authority.clerk.tickets.misc.identifierToAgentDelegation(identifier, agent)
```
blaine, line 23: clerk is probably a confusing term here; I haven't dug in, but seeing this immediately makes me think "what's the difference between an authority and a clerk? Is a clerk just an instance of an authority? Why the indirection?"

src/auth.ts
```diff
@@ -21,44 +20,16 @@ async (formValues: Record<string, string>): Promise<
...
+  await cabinet.addTicket("misc", agentDelegation)
```
blaine, line 24: "cabinet" feels a bit awkward here, and might be difficult for non-native english speakers? "store" or "storage" or even "ticketStore" might be more clear?

src/auth.ts
```diff
@@ -21,44 +20,16 @@ async (formValues: Record<string, string>): Promise<
...
 
   // Call account register implementation
   const result = await account.register(formValues, agentDelegation)
 
   if (result.registered) {
-    await cabinet.addUcans(result.ucans)
+    await cabinet.addTickets("account", result.tickets)
```
blaine, line 30: is there a scenario where we might store anything _but_ ucans here? Or is the abstraction just for different versions of ucans?

blaine (commented):
Not too worried since this is all (presumably?) intended for internal use, but it feels like there's a lot of complexity growing here.

Despite the flexibility, I'm not totally clear how we'd e.g. correlate / store a UCAN issued by the server to either the filesystem (i.e., the CID that corresponds to the server-stored volume) or the account username? Maybe that's a layer above – but I worry that the extra abstraction could make it difficult for developers to be able to debug issues when they come up?

Nothing blocking, though, just trying to keep an eye on complexity vs flexibility. 😊

src/auth.ts
icidasset: Yeah, cabinet also stores access keys. The ticket abstraction is for different versions of ucans (or ucan-like).

src/auth.ts
icidasset: Yeah naming's difficult. It does store more than tickets, it's like this internal storage thing one. Confusingly we also have a storage component, which is what the cabinet is based on (ie. storage component determines how, cabinet determines what)

src/auth.ts
icidasset: The thinking here was that 'clerk' is a thing that manages tickets.
Authority is what the 'clerk' is grouped under.
Authority module also concerns other things, such as file system access keys.
I'm open for other suggestions 💯 

I should also note that the authority module is still a work in progress.

icidasset:
> Not too worried since this is all (presumably?) intended for internal use, but it feels like there's a lot of complexity growing here.

@blaine Yeah internal use only. I'd definitely like to reduce complexity where possible!

> I'm not totally clear how we'd e.g. correlate / store a UCAN issued by the server to either the filesystem (i.e., the CID that corresponds to the server-stored volume) or the account username? Maybe that's a layer above – but I worry that the extra abstraction could make it difficult for developers to be able to debug issues when they come up?

Not entirely sure I understand your question. But the file system is tied the identifier DID, which in turn is associated with the account. This abstraction shouldn't make it more difficult, it's just stored encoded instead of not encoded. When we add back the developer tools (extension) we'll make sure it is decoded automatically. Or do you mean for developers who write custom components?

I also added the tagging (ticket category) to make it easier to find UCANs that are associated to the account, file system, or some thing else. Not sure if that helps? We can definitely add more utility functions to the `Inventory` class to make searching/filtering/etc easier.

avivash (approved):
This LGTM 👍🏼 I'm not entirely sure about the `clerk` naming, but I think it works for now and ~after looking through a synonym generator~ I can't think of a better name 😅

blaine:
> I also added the tagging (ticket category) to make it easier to find UCANs that are associated to the account, file system, or some thing else. Not sure if that helps? We can definitely add more utility functions to the `Inventory` class to make searching/filtering/etc easier.

Yeah, that sounds helpful. I was thinking more along the lines of "which UCAN is for my `blaine@fission.codes` dev account?" or when the UCAN is used, having some metadata that shows what it's for. That might be best handled by the server, to be fair. Once I try it out I'm sure I'll get the hang of it. 😊

The debugging stuff I was thinking about was sort of along the lines of:

"Error accessing ${x}" -> Why? (expired, invalid aud/iss, insufficient caps, invalid signature, etc) -> Which token was used / what properties were expected to be present in the token? Partially for folks building tooling on top of ts-odd, but also for situations where something breaks and an end-user developer needs to debug why something's broken.

icidasset:
> I was thinking more along the lines of "which UCAN is for my blaine@fission.codes dev account?"

@blaine Oh ya, that makes sense. There's a `findAccountUCAN` function in `src/components/account/fission/implementations/common.ts` which finds the UCAN that is addressed to the identifier and has the username fact in itself or one of its proofs. That in turn uses some of the ticket/ucan filtering functions located on that `Inventory` class.

I would like to know more about the metadata use cases.

> "Error accessing ${x}" -> Why? (expired, invalid aud/iss, insufficient caps, invalid signature, etc) -> Which token was used / what properties were expected to be present in the token? Partially for folks building tooling on top of ts-odd, but also for situations where something breaks and an end-user developer needs to debug why something's broken.

Oh yeah 💯 More of these things should be added along the way while we get to feature parity.

PR #566, opened 2023-08-25
Authority API and URL & QR code device linking

icidasset:
Allows for authority transfer between multiple clients at the same time.

```ts
const volume = await program.account.volume()

// Provider
await program.authority.provide([
  odd.authority.account,
  odd.authority.fileSystem.rootAccess(volume.did)
])

// Requestor
await program.authority.request(
  odd.authority.account
)

await program.authority.request(
  odd.authority.fileSystem.rootAccess(volume.did)
)
```

## Demo

https://github.com/oddsdk/ts-odd/assets/296665/c1ccff06-87cd-4586-a874-20b2ffce5edf

⚠️ There's an issue with the websocket server causing connection issues. When you close the connection and reopen it, the messages aren't received by the other end.

## How it works

1. One or more clients call ` program.authority.provide()` with authorisation **queries**. Queries are a way to select various UCANs and access keys. Above we have three queries: (1) query for the account capabilities (account component determines what these will be), (2 & 3) query for read & write access to the public & private roots of the file system
2. You get a URL as the result from `.provide()`. This URL can be converted into a QR code.
3. The URL is opened elsewhere.
4. Whenever `program.authority.request()` is called, it checks for the query params associated with the authority component implementation. `authority[challenge]` and `authority[publicKey]` specifically.
5. `.request()` detects the query params, establishes the channel via the channel component, making a websocket connection to the producer.
6. The second part of the handshake is done, confirming various cryptographic values. Producer tells requestor that everything looks in order.
7. Requestor sends over their query.
8. Producer compares the query of the requestor with their own. The received query must be equal of contained by configured provision query (eg. if you share the path `private/subdir/`, the requestor cannot ask for acces to `private/`)
9. Producer looks up the access keys and UCANs, then sends them over the channel encrypted.
10. Requestor decrypts keys and UCANs.
11. Authority transfer completed ✅

## Additional notes

- This does not ping the producer until they're online. Because that concept doesn't make sense here seeing that every session will have unique key-pair and thus unique URL/QR-code. If you've opened a link and closed the producer, the session becomes useless. Instead you open the producer again and use the new URL/QR-code.
- You can `provide` and `request` at the same time.

## Other changes

- Adjusted `Inventory` to also work with access keys.
- file system authority queries now require a DID parameter.
- Fixes some file system issues (rs-wnfs has been updated too)
- Moved all dependencies that are only used by components to peer dependencies (as to separate them from the sdk core)

## Test Plan

1. Register an account
12. Load a file system (may also be done before registering)
13. Make a change in the file system
14. Provide authority (can be done earlier too)
15. Request authority using the code above
16. Success! Hopefully 🙈


package.json
```diff
@@ -135,11 +119,28 @@
...
+    "@libp2p/webtransport": "^3.0.6",
+    "@multiformats/multiaddr": "^12.1.7",
+    "@noble/ciphers": "^0.1.4",
+    "@noble/curves": "^1.2.0",
+    "@noble/hashes": "^1.3.2",
+    "@ucans/core": "^0.12.0",
+    "blockstore-level": "^1.1.3",
+    "ipfs-bitswap": "^19.0.0",
+    "libp2p": "^0.46.6",
+    "localforage": "^1.10.0"
```
icidasset, line 144: All of these are component dependencies and are not part of the SDK core.

icidasset:
Superseded by #572

PR #570, opened 2023-09-05
Unique DIDs for file systems

icidasset:
This PR allows for file systems to have a unique ID instead of using the (root) identifier DID. This work continues on the work done for the authority system.


icidasset:
Superseded by #572

